PR #3, opened 2024-02-07
[DMS-4]  Ping Endpoint

CSR2017:


src/services/EdFi.DataManagementService.Api/Dockerfile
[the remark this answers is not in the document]
stephenfuqua: Will do this in a separate ticket.

src/services/EdFi.DataManagementService.Api/Program.cs
```diff
@@ -6,6 +6,10 @@
 var builder = WebApplication.CreateBuilder(args);
 var app = builder.Build();
 
+app.UsePathBase("/api");
+app.UseRouting();
+
 app.MapGet("/", () => "Data Management Service");
+app.MapGet("/ping", () => Results.Ok(new { StatusCode = 200, CurrentDateTime = DateTime.Now.ToString() }));
```
simpat-adam, line 13: Suggestion, it's not necessary to send the status code in the response, it's taken care of with `Results.Ok`.

src/services/EdFi.DataManagementService.Api/Program.cs
```diff
@@ -6,6 +6,10 @@
...
 
 app.Run();
```
simpat-adam, line 15: You can add a test for this change by using `Microsoft.AspNetCore.Mvc.Testing` . This is a good walkthrough: https://www.twilio.com/en-us/blog/test-aspnetcore-minimal-apis

simpat-adam (changes requested):

src/services/EdFi.DataManagementService.Api/Program.cs
CSR2017: > Suggestion, it's not necessary to send the status code in the response, it's taken care of with `Results.Ok`.

True, just created custom response with status code.

src/services/EdFi.DataManagementService.Api/Program.cs
```diff
@@ -6,6 +6,12 @@
 var builder = WebApplication.CreateBuilder(args);
 var app = builder.Build();
 
+app.UsePathBase("/api");
+app.UseRouting();
+
 app.MapGet("/", () => "Data Management Service");
+app.MapGet("/ping", () => Results.Ok(new { CurrentDateTime = DateTime.Now.ToString() }));
```
simpat-adam, line 13: Sorry I didn't mention this yesterday but you should omit the `ToString()`. Even though this is a contrived endpoint and it doesn't really matter, someone might use it as an example. There's no reason to convert our datetimes to strings and it just introduces concerns about default string formats etc. If you remove the `ToString` your test will still behave exactly the same.

src/.editorconfig
```diff
@@ -121,6 +121,7 @@ dotnet_diagnostic.IDE0005.severity=suggestion # Remove unnecessary using directi
 dotnet_diagnostic.S4136.severity=suggestion # Method overloads should be grouped together
 dotnet_diagnostic.S1135.severity=suggestion # Complete TODO comments
 dotnet_diagnostic.S112.severity=suggestion # 'System.Exception' should not be thrown by user code.
+dotnet_diagnostic.S1118.severity=suggestion # Only considered for utility classes
```
simpat-adam, line 124: Let's have a quick team discussion on this. You had to add this because of `public partial class Program { }` for unit testing. I think you could also just add an empty protected or private constructor to that class explicitly. Have we made this severity modification in other projects? I don't feel strongly but just wanted to check.

stephenfuqua: https://sonarsource.github.io/rspec/#/rspec/S1118/csharp

I would prefer not to modify this code rule on a general basis. Yes, please see if there is a different resolution, or if not practical, try suppressing the rule in the one file rather than in the global settings.

simpat-adam (approved):

PR #4, opened 2024-02-12
[DMS-9] Pre-wire for appSettings

CSR2017:
Added LogAppSettingsService for enabling the IOptions pattern to do dependency injection on AppSettings. "LogAppSettingsService" may be refactored to work with logging.

### DIFF
--- a/src/services/EdFi.DataManagementService.Api/Infrastructure/LogAppSettingsService.cs
+++ b/src/services/EdFi.DataManagementService.Api/Infrastructure/LogAppSettingsService.cs
@@ -1,0 +1,31 @@
+// SPDX-License-Identifier: Apache-2.0
+// Licensed to the Ed-Fi Alliance under one or more agreements.
+// The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+// See the LICENSE and NOTICES files in the project root for more information.
+
+using EdFi.DataManagementService.Api.Configuration;
+using Microsoft.Extensions.Options;
+
+namespace EdFi.DataManagementService.Api.Infrastructure
+{
+    // Example class for IOptions 
+    public class LogAppSettingsService()
+    {
+        private readonly ILogger<LogAppSettingsService>? _logger;
+        private readonly IOptions<AppSettings>? _options;
+
+        public LogAppSettingsService(ILogger<LogAppSettingsService>? logger,
+            IOptions<AppSettings>? options) : this()
+        {
+            _logger = logger;
+            _options = options;
+        }
+
+        public void Log()
+        {
+            var appSettings = _options?.Value;
+            _logger?.LogInformation("BeginAllowedSchoolYear = {year}.", appSettings?.BeginAllowedSchoolYear);
+            _logger?.LogInformation("EndAllowedSchoolYear = {year}.", appSettings?.EndAllowedSchoolYear);
+        }
+    }
+}
--- a/src/services/EdFi.DataManagementService.Api/Infrastructure/LogAppSettingsService.cs
+++ b/src/services/EdFi.DataManagementService.Api/Infrastructure/LogAppSettingsService.cs
@@ -22,6 +22,6 @@ namespace EdFi.DataManagementService.Api.Infrastructure
             _options = options;
         }
 
-        public void LogToConsole() => _logger.LogInformation(message: JsonSerializer.Serialize(_options?.Value));
+        public void LogToConsole() => _logger.LogInformation(message: JsonSerializer.Serialize(_options.Value));
     }
 }
--- a/src/services/EdFi.DataManagementService.Api/Infrastructure/LogAppSettingsService.cs
+++ b/src/services/EdFi.DataManagementService.Api/Infrastructure/LogAppSettingsService.cs
@@ -5,27 +5,23 @@
 
 using EdFi.DataManagementService.Api.Configuration;
 using Microsoft.Extensions.Options;
+using System.Text.Json;
 
 namespace EdFi.DataManagementService.Api.Infrastructure
 {
     // Example class for IOptions 
     public class LogAppSettingsService()
     {
-        private readonly ILogger<LogAppSettingsService>? _logger;
-        private readonly IOptions<AppSettings>? _options;
+        private readonly ILogger<LogAppSettingsService> _logger = null!;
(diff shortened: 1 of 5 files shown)


src/services/EdFi.DataManagementService.Api/Infrastructure/LogAppSettingsService.cs
```diff
@@ -0,0 +1,31 @@
...
+using EdFi.DataManagementService.Api.Configuration;
+using Microsoft.Extensions.Options;
+
+namespace EdFi.DataManagementService.Api.Infrastructure
+{
+    // Example class for IOptions 
+    public class LogAppSettingsService()
+    {
+        private readonly ILogger<LogAppSettingsService>? _logger;
+        private readonly IOptions<AppSettings>? _options;
```
stephenfuqua, line 15: Why allow these to be nullable? I don't see that bringing any value.

src/services/EdFi.DataManagementService.Api/Infrastructure/LogAppSettingsService.cs
```diff
@@ -0,0 +1,31 @@
...
+            IOptions<AppSettings>? options) : this()
+        {
+            _logger = logger;
+            _options = options;
+        }
+
+        public void Log()
+        {
+            var appSettings = _options?.Value;
+            _logger?.LogInformation("BeginAllowedSchoolYear = {year}.", appSettings?.BeginAllowedSchoolYear);
```
stephenfuqua, line 27: How does the application respond if you enter a non-numeric value for the school year?

src/services/EdFi.DataManagementService.Api/Infrastructure/WebApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,19 @@
...
+
+using EdFi.DataManagementService.Api.Configuration;
+
+namespace EdFi.DataManagementService.Api.Infrastructure
+{
+    public static class WebApplicationBuilderExtensions
+    {
+        public static void AddServices(this WebApplicationBuilder webAppBuilder)
+        {
+            webAppBuilder.Services.Configure<AppSettings>(webAppBuilder.Configuration.GetSection("AppSettings"));
```
stephenfuqua, line 14: I like how this is parsing directly into our class without having to write a bunch of cast statements from strings.

src/services/EdFi.DataManagementService.Api/Infrastructure/LogAppSettingsService.cs
```diff
@@ -0,0 +1,31 @@
...
+        private readonly IOptions<AppSettings>? _options;
+
+        public LogAppSettingsService(ILogger<LogAppSettingsService>? logger,
+            IOptions<AppSettings>? options) : this()
+        {
+            _logger = logger;
+            _options = options;
+        }
+
+        public void Log()
```
stephenfuqua, line 24: Just reading the class and function name, I do not know what to expect. Is the purpose of this class just to write out a log of the appsettings?

The word "log" can be used a verb or a noun. When it is used by itself, it sounds like a noun, which does not make for a good function name.

src/services/EdFi.DataManagementService.Api/Infrastructure/LogAppSettingsService.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            var appSettings = _options?.Value;
+            _logger?.LogInformation("BeginAllowedSchoolYear = {year}.", appSettings?.BeginAllowedSchoolYear);
+            _logger?.LogInformation("EndAllowedSchoolYear = {year}.", appSettings?.EndAllowedSchoolYear);
```
stephenfuqua, line 28: If this is a generic logger, then perhaps it would be easier to JSON serialize the `AppSettings` and write that out, instead of hard-coding each new setting.

src/services/EdFi.DataManagementService.Api/Infrastructure/LogAppSettingsService.cs
CSR2017: Now appsettings values getting logged as Json. This is a sample class for enabling IOption pattern. So, did not include validation and error handling.

src/services/EdFi.DataManagementService.Api/Infrastructure/LogAppSettingsService.cs
```diff
@@ -0,0 +1,27 @@
...
+using Microsoft.Extensions.Options;
+using System.Text.Json;
+
+namespace EdFi.DataManagementService.Api.Infrastructure
+{
+    // Example class for IOptions 
+    public class LogAppSettingsService()
+    {
+        private readonly ILogger<LogAppSettingsService> _logger = null!;
+        private readonly IOptions<AppSettings> _options = null!;
```
stephenfuqua, line 16: I haven't seen `null!` before. [Looks intriguing](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-forgiving)

src/services/EdFi.DataManagementService.Api/Infrastructure/LogAppSettingsService.cs
```diff
@@ -0,0 +1,27 @@
...
+
+        public LogAppSettingsService(ILogger<LogAppSettingsService> logger,
+            IOptions<AppSettings> options) : this()
+        {
+            _logger = logger;
+            _options = options;
+        }
+
+        public void LogToConsole() => _logger.LogInformation(message: JsonSerializer.Serialize(_options?.Value));
```
stephenfuqua, line 25: Can the `?` be removed here, since `_options` is not nullable?

src/services/EdFi.DataManagementService.Api/Infrastructure/LogAppSettingsService.cs
CSR2017: If it is not declared as nullable then will get ([CS8618 ](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/nullable-warnings?f1url=%3FappId%3Droslyn%26k%3Dk(CS8618)#nonnullable-reference-not-initialized)). I can not assign null value to non-nullable, so declaring the default value as not-null.

simpat-adam (approved):

PR #5, opened 2024-02-13
[DMS-8] Add fixed window rate limit capability

simpat-adam:


README.md
```diff
@@ -21,6 +21,19 @@ Developer Documents:
...
 
+## Configuration
+
+### RateLimit
+Basic rate limiting can be applied by supplying a `RateLimit` object in the `appsettings.json` file. If no `RateLimit` object is supplied, rate limiting is not configured for the application. Rate limiting (when applied) will be set globally and apply to all application endpoints. 
+
+The `RateLimit` object should have the following parameters.
+
+Parameter       | Description
+----------------|----------------
```
stephenfuqua, line 32: I don't recall ever seeing a markdown table that doesn't have  leading and trailing pipes. But it works nicely here ✅.

The line wrapping comment above of course does not apply to the table.

The Google Style Guide for markdown suggests that this should be a list instead of a table, because of the line width. In this particular case, the table seems easily readable as is, and no need to change.

README.md
```diff
@@ -21,6 +21,19 @@ Developer Documents:
 * [Copyright and License Notices](./NOTICES.md)
 * [License](./LICENSE)
 
+## Configuration
+
+### RateLimit
+Basic rate limiting can be applied by supplying a `RateLimit` object in the `appsettings.json` file. If no `RateLimit` object is supplied, rate limiting is not configured for the application. Rate limiting (when applied) will be set globally and apply to all application endpoints.
```
stephenfuqua, line 27: Two recommendations:

1. Add a blank line above this, so that there's always a line between the heading and the text.
2. Add hard line wrapping.

Just a few weeks ago I published a [Markdown Best Practices](https://techdocs.ed-fi.org/display/ETKB/Markdown+Best+Practices) document, but forgot to share it with folks. Includes some useful VS Code extensions.

src/services/EdFi.DataManagementService.Api.Tests/RateLimitTests.cs
```diff
@@ -0,0 +1,37 @@
...
+            });
+        using var client = factory.CreateClient();
+
+        // Act
+        var response1 = await client.GetAsync("/api/ping");
+        var response2 = await client.GetAsync("/api/ping");
+
+        // Assert
+        response1.StatusCode.Should().Be(HttpStatusCode.OK);
+        response2.StatusCode.Should().Be(HttpStatusCode.TooManyRequests);
```
stephenfuqua, line 35: That's a nice test 😆 . Truly, I love it.

src/services/EdFi.DataManagementService.Api/Configuration/RateLimitOptions.cs
```diff
@@ -0,0 +1,19 @@
...
+// See the LICENSE and NOTICES files in the project root for more information.
+
+namespace EdFi.DataManagementService.Api.Configuration
+{
+    public class RateLimitOptions
+    {
+        public const string RateLimit = "RateLimit";
+
+        //Number of requests per window
+        public int PermitLimit { get; set; }
```
stephenfuqua, line 13: When I try setting a negative value, I get an unhandled exception and the application doesn't startup. Failing to start makes sense, and the unhandled exception should be covered in the upcoming DMS-8 work. No concern in this ticket. @CSR2017 in the DMS-8 work, you can use this test case when validating the logging and exception handling.

stephenfuqua (approved):

PR #8, opened 2024-02-19
[DMS-12] API route validation

bradbanister:
Manual testing can be done with the VS Code Rest Client extension using EdFi.DataManagementService.Api.Tests/RestClient/test.http


.gitattributes
```diff
@@ -0,0 +1,93 @@
+## GITATTRIBUTES FOR WEB PROJECTS
```
bradbanister, line 1: Taken from Meadowlark repo

src/services/EdFi.DataManagementService.Api.Tests/RestClient/test.http
```diff
@@ -0,0 +1,165 @@
+### Test POST of invalid resource
```
bradbanister, line 1: Taken from Meadowlark.net spike

src/services/EdFi.DataManagementService.Api/ApiSchema/Model/MetaEdProjectName.cs
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: Apache-2.0
+// Licensed to the Ed-Fi Alliance under one or more agreements.
+// The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+// See the LICENSE and NOTICES files in the project root for more information.
+
+namespace EdFi.DataManagementService.Api.ApiSchema.Model;
+
+/**
```
bradbanister, line 8: Ugh, need to fix comment style

src/services/EdFi.DataManagementService.Api/ApiSchema/ProjectSchema.cs
```diff
@@ -0,0 +1,44 @@
...
+{
+    private readonly Lazy<MetaEdProjectName> _projectName = new(() =>
+    {
+        return new MetaEdProjectName(_projectSchemaNode.SelectRequiredNodeFromPathAs<string>("$.projectName"));
+    });
+
+    /// <summary>
+    /// The ProjectName for this ProjectSchema, taken from the projectName
+    /// </summary>
+    public MetaEdProjectName ProjectName => _projectName.Value;
```
bradbanister, line 25: This seems like the way to memoize in C#

src/services/EdFi.DataManagementService.Api/AspNetCoreFrontend.cs
```diff
@@ -0,0 +1,80 @@
...
+using EdFi.DataManagementService.Api.Core;
+using EdFi.DataManagementService.Api.Core.Middleware;
+using EdFi.DataManagementService.Api.Core.Model;
+
+namespace EdFi.DataManagementService.Api
+{
+    /// <summary>
+    /// A thin static class that converts from ASP.NET Core to the DMS facade.
+    /// </summary>
+    public static class Frontend
```
bradbanister, line 16: Moves this stuff out of Program.cs

src/services/EdFi.DataManagementService.Api/Core/CoreFacade.cs
```diff
@@ -0,0 +1,104 @@
...
+/// The intent of this design is to provide a web framework-independent interface
+/// for 1) ease of testing and 2) ease of supporting future frontends e.g.
+/// AWS Lambda, Azure functions, etc.
+/// </summary>
+public static class CoreFacade
+{
+    /// <summary>
+    /// The pipeline steps to satisfy an upsert request
+    /// </summary>
+    private static readonly PipelineProvider _upsertSteps = new PipelineProvider()
```
bradbanister, line 25: Where the pipeline steps are built. Static so it's only done on first request.

src/services/EdFi.DataManagementService.Api/Core/Handler/DeleteByIdHandler.cs
```diff
@@ -0,0 +1,21 @@
...
+// See the LICENSE and NOTICES files in the project root for more information.
+
+using EdFi.DataManagementService.Core.Pipeline;
+
+namespace EdFi.DataManagementService.Api.Core.Handler;
+
+/// <summary>
+/// Handles a delete request that has made it through the middleware pipeline steps.
+/// </summary>
+public class DeleteByIdHandler : IPipelineStep
```
bradbanister, line 13: Handlers are the final step in the pipeline. They are where core will interface with the backend DB plugins through a to-be-defined interface.

src/services/EdFi.DataManagementService.Api/Core/Middleware/ApiSchemaLoadingMiddleware.cs
```diff
@@ -0,0 +1,25 @@
...
+
+namespace EdFi.DataManagementService.Api.Core.Middleware;
+
+
+public class ApiSchemaLoadingMiddleware : IPipelineStep
+{
+    public async Task Execute(PipelineContext context, Func<Task> next)
+    {
+        // Hardcoded and synchronous way to read the API Schema file - every time!
+        JsonNode apiSchemaRootNode =
```
bradbanister, line 17: Seems like this should be singleton injected

src/services/EdFi.DataManagementService.Api/Core/Middleware/BuildResourceInfoMiddleware.cs
```diff
@@ -0,0 +1,27 @@
...
+namespace EdFi.DataManagementService.Api.Core.Middleware;
+
+/// <summary>
+/// Builds the ResourceInfo to pass to the backends
+/// </summary>
+public class BuildResourceInfoMiddleware : IPipelineStep
+{
+    public async Task Execute(PipelineContext context, Func<Task> next)
+    {
+        context.ResourceInfo = new(
```
bradbanister, line 17: Will use in handler logging at first

src/services/EdFi.DataManagementService.Api/Core/Middleware/ParsePathMiddleware.cs
```diff
@@ -0,0 +1,77 @@
...
+        Match? match = pathExpression.Match(path);
+
+        if (match == null) return null;
+
+        string? documentUuidValue = match.Groups["documentUuid"].Value;
+        if (documentUuidValue == "") documentUuidValue = null;
+
+        return new(
+          ProjectNamespace: new(match.Groups["projectNamespace"].Value.ToLower()),
+          EndpointName: new(Decapitalize(match.Groups["endpointName"].Value)),
```
bradbanister, line 41: Need to loosen this up for case insensitive routes

src/services/EdFi.DataManagementService.Api/Core/Model/FrontendRequest.cs
```diff
@@ -0,0 +1,37 @@
...
+
+using System.Text.Json.Nodes;
+using EdFi.DataManagementService.Api.Core.Middleware;
+
+namespace EdFi.DataManagementService.Api.Core.Model;
+
+/// <summary>
+/// An API request sent from the frontend to be processed
+/// </summary>
+public record FrontendRequest(
```
bradbanister, line 14: A standardized web framework request, without framework-specific objects

src/services/EdFi.DataManagementService.Api/Core/Model/FrontendResponse.cs
```diff
@@ -0,0 +1,11 @@
...
+// Licensed to the Ed-Fi Alliance under one or more agreements.
+// The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+// See the LICENSE and NOTICES files in the project root for more information.
+
+namespace EdFi.DataManagementService.Api.Core.Model;
+
+/// <summary>
+/// The API response returned to the frontend
+/// </summary>
+public record FrontendResponse(int StatusCode, string Body);
```
bradbanister, line 11: A standardized response, without web framework-specific objects

src/services/EdFi.DataManagementService.Api/Core/Model/No.cs
```diff
@@ -0,0 +1,50 @@
...
+// The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+// See the LICENSE and NOTICES files in the project root for more information.
+
+using System.Text.Json.Nodes;
+using EdFi.DataManagementService.Api.ApiSchema;
+
+namespace EdFi.DataManagementService.Api.Core.Model;
+
+/// <summary>
+/// Null objects to avoid nullable types when null is irrelevant
```
bradbanister, line 12: A set of null objects, mostly to simplify pipeline step code by avoiding nullable types in PipelineContext

src/services/EdFi.DataManagementService.Api/Core/Model/ResourceInfo.cs
```diff
@@ -0,0 +1,49 @@
...
+// See the LICENSE and NOTICES files in the project root for more information.
+
+using EdFi.DataManagementService.Api.ApiSchema.Model;
+
+namespace EdFi.DataManagementService.Api.Core.Model;
+
+/// <summary>
+/// Base API resource information for passing along to backends.
+/// </summary>
+public record BaseResourceInfo(
```
bradbanister, line 13: Split into base and extended ResourceInfo, because it will be necessary for reference extraction shortly.

src/services/EdFi.DataManagementService.Api/Core/Model/TraceId.cs
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: Apache-2.0
+// Licensed to the Ed-Fi Alliance under one or more agreements.
+// The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+// See the LICENSE and NOTICES files in the project root for more information.
+
+namespace EdFi.DataManagementService.Api.Core.Model;
+
+/// <summary>
+/// A string type branded as a traceId, which is a UUID representing a unique request
```
bradbanister, line 9: Will change comment, it's frontend dependent, doesn't have to be a UUID

src/services/EdFi.DataManagementService.Api/EdFi.DataManagementService.Api.csproj
```diff
@@ -10,6 +10,7 @@
 
   <ItemGroup>
     <Content Include="EdFi.DataManagementService.Api.nuspec" CopyToPublishDirectory="Always" CopyToOutputDirectory="Never" />
+    <None Include="ApiSchema/DataStandard-5.0.0-ApiSchema.json" CopyToPublishDirectory="Always" CopyToOutputDirectory="Always"/>
```
bradbanister, line 13: Is this the "correct" way to have the ApiSchema.json copied/bundled? Used <None> instead of <Content> just because it did what I wanted.

src/services/EdFi.DataManagementService.Api/Core/Pipeline/PipelineContext.cs
```diff
@@ -0,0 +1,50 @@
...
+
+using EdFi.DataManagementService.Api.ApiSchema;
+using EdFi.DataManagementService.Api.Core.Model;
+
+namespace EdFi.DataManagementService.Core.Pipeline;
+
+/// <summary>
+/// The full context of the request, enriched by pipeline steps and the final handler
+/// </summary>
+public class PipelineContext(FrontendRequest _frontendRequest)
```
bradbanister, line 14: Except for FrontendRequest, PipelineContext is intended to be mutable as passed down the pipeline.

src/services/EdFi.DataManagementService.Api/Infrastructure/LoggingMiddleware.cs
```diff
@@ -26,11 +26,19 @@ public async Task Invoke(HttpContext context, ILogger<LoggingMiddleware> logger)
...
         catch (Exception ex)
         {
-            var response = context.Response;
-            response.ContentType = "application/json";
             logger.LogError(JsonSerializer.Serialize(new { message = "An uncaught error has occurred", error = new { ex.Message, ex.StackTrace }, traceId = context.TraceIdentifier }));
-            response.StatusCode = (int) HttpStatusCode.InternalServerError;
-            await response.WriteAsync(JsonSerializer.Serialize(new { message = "The server encountered an unexpected condition that prevented it from fulfilling the request.", traceId = context.TraceIdentifier }));
+
+            var response = context.Response;
+            if (!response.HasStarted)
```
bradbanister, line 32: InvalidOperationException gets thrown if a response has started elsewhere and a middleware tries to modify it. This was happening when I was returning 204 with an empty string body, causing an exception because ASP.NET core only accepts a null body with 204, but after the response had started. The InvalidOperationException was obscuring the actual error.

src/.editorconfig
```diff
@@ -122,6 +122,8 @@ dotnet_diagnostic.S4136.severity=suggestion # Method overloads should be grouped
 dotnet_diagnostic.S1135.severity=suggestion # Complete TODO comments
 dotnet_diagnostic.S112.severity=suggestion # 'System.Exception' should not be thrown by user code.
 
+# Allow empty records for discriminated union types
+dotnet_diagnostic.S2094.severity=none # S2094: Classes should not be empty
```
bradbanister, line 126: Pattern for discriminated unions requires an empty record superclass

src/services/EdFi.DataManagementService.Api/Core/Middleware/ParsePathMiddleware.cs
```diff
@@ -0,0 +1,79 @@
...
+        );
+    }
+
+    /// <summary>
+    /// Check that this is a well formed UUID v4
+    /// </summary>
+    private static bool IsDocumentUuidWellFormed(DocumentUuid documentUuid)
+    {
+        // Regex for a UUID v4 string
+        Regex uuid4Regex = new(@"^[0-9a-f]{8}-[0-9a-f]{4}-[4][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$");
```
stephenfuqua, line 52: Suggest static instance with Compiled flag.

.gitattributes
stephenfuqua: Its missing a lot of .net specific stuff, like ignoring `obj/` and `bin/`.

src/services/EdFi.DataManagementService.Api.Tests/EndpointsTests.cs
```diff
@@ -3,10 +3,10 @@
...
 // See the LICENSE and NOTICES files in the project root for more information.
 
+using System.Net;
 using FluentAssertions;
 using Microsoft.AspNetCore.Mvc.Testing;
 using NUnit.Framework;
-using System.Net;
 
 namespace EdFi.DataManagementService.Api.Tests
 {
```
stephenfuqua, line 12: I'd be interested in seeing what csharpier does if you change this to the newer syntax: add a semicolon at the end of line 11, delete line 12 and the closing brace somewhere below, and then you're able to outdent the entire block to save some left side whitespace.

src/services/EdFi.DataManagementService.Api.Tests/RestClient/test.http
stephenfuqua: What do you think about moving RestClient tests up much higher, maybe into `/tests`?

src/services/EdFi.DataManagementService.Api/ApiSchema/ApiSchemaLoader.cs
```diff
@@ -0,0 +1,36 @@
...
+    /// </summary>
+    public JsonNode ApiSchemaRootNode => _apiSchemaRootNode;
+
+    public ApiSchemaFileLoader(ILogger<ApiSchemaFileLoader> _logger)
+    {
+        _logger.LogDebug("Entering ApiSchemaFileLoader");
+
+        // Hardcoded and synchronous way to read the API Schema file
+        string fileLocation = $"{AppContext.BaseDirectory}/ApiSchema/DataStandard-5.0.0-ApiSchema.json";
+        JsonNode? rootNodeFromFile = JsonNode.Parse(File.ReadAllText(fileLocation));
```
stephenfuqua, line 27: This might be difficult to work with from a unit testing perspective.

Maybe helpful to have a lightweight utility intereface & class that replaces File.ReadAllText.

Or alternately, could change the `fileLocation` to a protected property. In the unit tests, override the value of `fileLocation` to something specific for the unit test project.

src/services/EdFi.DataManagementService.Api/ApiSchema/ApiSchemaLoader.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        if (rootNodeFromFile == null)
+        {
+            _logger.LogCritical("Unable to read and parse Api Schema file at {fileLocation}", fileLocation);
+            throw new InvalidOperationException("Unable to read and parse Api Schema file.");
```
stephenfuqua, line 32: The exception will be caught by the logging middleware and will also generate an error message.

I agree with this being "critical". One way out of this might be to create a custom "critical exception" that could be thrown here, with no manual logging. In the logging middleware, handle the custom exception using `LogCritical`.

src/services/EdFi.DataManagementService.Api/ApiSchema/DataStandard-5.0.0-ApiSchema.json
stephenfuqua, line 1: I'm going to download this today and try to create an initial NuGet package, so that you can remove this before merging the PR.

src/services/EdFi.DataManagementService.Api/ApiSchema/Extensions/JsonHelperExtensions.cs
```diff
@@ -0,0 +1,116 @@
...
+    /// </summary>
+    public static JsonNode? SelectNodeFromPath(this JsonNode jsonNode, string jsonPathString, ILogger logger)
+    {
+        try
+        {
+            JsonPath? jsonPath = JsonPath.Parse(jsonPathString);
+            if (jsonPath == null)
+            {
+                logger.LogError("Malformed JSONPath string '{jsonPathString}'", jsonPathString);
+                throw new InvalidOperationException($"Malformed JSONPath string '{jsonPathString}'");
```
stephenfuqua, line 25: No need to log the error here. Middleware will take care of the logging.

Should this be critical / deadly?

src/services/EdFi.DataManagementService.Api/ApiSchema/Extensions/JsonHelperExtensions.cs
```diff
@@ -0,0 +1,116 @@
...
+                    $"JSONPath string '{jsonPathString}' selected multiple values"
+                );
+            }
+
+            return result.Matches[0].Value;
+        }
+        catch (PathParseException)
+        {
+            logger.LogError("Unexpected Json.Path error for '{jsonPathString}", jsonPathString);
+            throw new InvalidOperationException($"Unexpected Json.Path error for '{jsonPathString}'");
```
stephenfuqua, line 55: Recommend adding this as an inner exception. Or just let the exception go uncaught here, if we don't need to wrap in an outer exception.

```
        catch (PathParseException ppe)
        {
            throw new InvalidOperationException($"Unexpected Json.Path error for '{jsonPathString}'", ppe);
```

src/services/EdFi.DataManagementService.Api/ApiSchema/ProjectSchema.cs
stephenfuqua: Yes, exposing a private instance property as a public property...

src/services/EdFi.DataManagementService.Api/AspNetCoreFrontend.cs
```diff
@@ -0,0 +1,105 @@
...
+{
+    /// <summary>
+    /// A thin static class that converts from ASP.NET Core to the DMS facade.
+    /// </summary>
+    public static class AspNetCoreFrontend
+    {
+        /// <summary>
+        /// Takes an HttpRequest and returns a deserialized request body
+        /// </summary>
+        private static async Task<JsonNode?> ExtractJsonBodyFrom(HttpRequest request)
```
stephenfuqua, line 21: https://fakeiteasy.github.io/docs/7.4.0/Recipes/faking-http-client/

Related, but not exactly the same and might not be helpful.

src/services/EdFi.DataManagementService.Api/Core/ICoreFacade.cs
```diff
@@ -0,0 +1,37 @@
...
+/// The intent of this design is to provide a web framework-independent interface
+/// for 1) ease of testing and 2) ease of supporting future frontends e.g.
+/// AWS Lambda, Azure functions, etc.
+/// </summary>
+public interface ICoreFacade
+{
+    /// <summary>
+    /// DMS entry point for API upsert requests
+    /// </summary>
+    public Task<FrontendResponse> Upsert(FrontendRequest frontendRequest);
```
stephenfuqua, line 21: Consider renaming these with "Async" to follow common C# idiom, e.g. `UpsertAsync`.

Since there is no alternative, I'm of an open mind on this point.

src/services/EdFi.DataManagementService.Api/Core/Middleware/ParsePathMiddleware.cs
```diff
@@ -0,0 +1,96 @@
...
+/// Parses and validates the path from the frontend is well formed. Adds PathComponents
+/// to the context if it is.
+/// </summary>
+public partial class ParsePathMiddleware(ILogger _logger) : IPipelineStep
+{
+    // Matches all of the following sample expressions:
+    // /ed-fi/Sections
+    // /ed-fi/Sections/
+    // /ed-fi/Sections/idValue
+    [GeneratedRegex(@"\/(?<projectNamespace>[^/]+)\/(?<endpointName>[^/]+)(\/|$)((?<documentUuid>[^/]*$))?")]
```
stephenfuqua, line 21: https://learn.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.generatedregexattribute?view=net-8.0

If I'm reading that correctly, this successfully gives us a pre-compiled regex ✅ .

src/services/EdFi.DataManagementService.Api/Core/Pipeline/IPipelineStep.cs
```diff
@@ -0,0 +1,24 @@
...
+///
+/// You can think of everything done in a pipeline step before the next() call as being
+/// pre-processing and everything after as post-processing.
+///
+/// Pipeline steps MUST NOT maintain object state, as object state would persist across
+/// requests.
+/// </summary>
+public interface IPipelineStep
+{
+    Task Execute(PipelineContext context, Func<Task> next);
```
stephenfuqua, line 23: Same "Async" moniker question as above.

src/services/EdFi.DataManagementService.Api/Core/Pipeline/PipelineProvider.cs
```diff
@@ -0,0 +1,42 @@
...
+/// <summary>
+/// Implements a simple Data Management Service pipeline provider, which
+/// is modeled on the Middleware pattern.
+///
+/// Add pipeline steps in execution order with StartWith() and AndThen().
+/// Run the pipeline steps with Run().
+/// </summary>
+public class PipelineProvider
+{
+    private readonly List<IPipelineStep> _steps = [];
```
stephenfuqua, line 17: Wondering if we have any potential for thread safety problems with this approach. There's a performance hit to using a locking collection, so we don't want to overdo it. Haven't fully thought this through yet... probably fine since the pipeline is setup and then used, but not altered.

https://learn.microsoft.com/en-us/dotnet/standard/collections/thread-safe/

src/services/EdFi.DataManagementService.Api/EdFi.DataManagementService.Api.csproj
stephenfuqua: Right, but we really want it bundled into a NuGet package, so we want something entirely different.

src/services/EdFi.DataManagementService.Api/Program.cs
```diff
@@ -22,15 +23,23 @@
...
+        {
+            logger.LogInformation($"Log information while debugging");
+        }
+        logger.LogInformation(message: JsonSerializer.Serialize(options.Value));
+        return Results.Ok("Successfully logged Application Settings");
     }
-    logger.LogInformation(message: JsonSerializer.Serialize(options.Value));
-    return Results.Ok("Successfully logged Application Settings");
-});
+);
```
stephenfuqua, line 37: @CSR2017 is it safe to remove the `/LogAppSettings` endpoint now, since it has served its temporary purpose?

.gitattributes
bradbanister: I think you mean .gitignore 🙂

src/services/EdFi.DataManagementService.Api.Tests/RestClient/test.http
bradbanister: Done

.gitattributes
stephenfuqua: Yes, good point, I thought I was looking at something different.

src/services/EdFi.DataManagementService.Api/ApiSchema/ApiSchemaLoader.cs
bradbanister: I have this injected as an IApiSchemaProvider, which only has a JsonNode get property. So unit tests can just be given a JsonNode directly.

src/services/EdFi.DataManagementService.Api/Core/Middleware/ParsePathMiddleware.cs
bradbanister: Yes, this seemed the "modern" way to do it.

src/services/EdFi.DataManagementService.Api.Tests/EndpointsTests.cs
bradbanister: It did the right thing.

src/services/EdFi.DataManagementService.Api/Core/Pipeline/PipelineProvider.cs
bradbanister: The `AndThen()` stuff was cool, but because it was stateful issue I dropped in in favor of just a constructor with the whole list. Wrapped construction in the facade with a `Lazy<>`  and we are good to go.

src/services/EdFi.DataManagementService.Api/Program.cs
bradbanister: Removed

src/services/EdFi.DataManagementService.Api/ApiSchema/ApiSchemaLoader.cs
bradbanister: Right, that was my thought

src/services/EdFi.DataManagementService.Api.Tests/ApiSchemaBuilder.cs
```diff
@@ -0,0 +1,155 @@
...
+using EdFi.DataManagementService.Api.ApiSchema;
+using Microsoft.Extensions.Logging.Abstractions;
+
+namespace EdFi.DataManagementService.Api.Tests;
+
+/// <summary>
+/// This class provides a fluent interface for building an ApiSchema suitable for unit testing,
+/// allow tests to focus on scenarios without getting bogged down in JSON authoring
+/// </summary>
+public class ApiSchemaBuilder
```
bradbanister, line 16: Like MetaEd's MetaEdTextBuilder but for the DMS. A fluent interface for building ApiSchema.json for test scenarios.

bradbanister: I should note that this is just enough to support these first tests. As DMS expands its usage of the ApiSchema.json, we'll add test support here for those newly used elements.

src/services/EdFi.DataManagementService.Api/AspNetCoreFrontend.cs
```diff
@@ -0,0 +1,105 @@
...
+using EdFi.DataManagementService.Api.Core;
+using EdFi.DataManagementService.Api.Core.Middleware;
+using EdFi.DataManagementService.Api.Core.Model;
+
+namespace EdFi.DataManagementService.Api
+{
+    /// <summary>
+    /// A thin static class that converts from ASP.NET Core to the DMS facade.
+    /// </summary>
+    public static class AspNetCoreFrontend
```
bradbanister, line 16: @CSR2017 The purpose of this class is a simple translation from what ASP.NET Core handlers get to a DMS core request. The idea is for DMS Core to have no ASP.NET in it. With a well defined facade to DMS Core, it will be trivial for contributors to add frontends for e.g. AWS Lambda, Azure Functions, etc.

It's a static class because it's no different from having the functions inlined in Program.cs. The dependency ICoreFacade is injected by ASP.NET core either way.

src/services/EdFi.DataManagementService.Api/Backend/DeleteRequest.cs
```diff
@@ -0,0 +1,21 @@
...
+namespace EdFi.DataManagementService.Api.Backend;
+
+/// <summary>
+/// A delete request to a document repository
+/// </summary>
+/// <param name="DocumentUuid">The document UUID to delete</param>
+/// <param name="ResourceInfo">The ResourceInfo for the resource being deleted</param>
+/// <param name="validateNoReferencesToDocument">If true, validates there are no references to the document</param>
+/// <param name="TraceId">The request TraceId</param>
+public record DeleteRequest(
```
bradbanister, line 16: @CSR2017 These request/result records are backend-independent messages for the communication between DMS core and a particular backend as implemented by IDocumentStoreRepository, which will eventually be pluggable. Ed-Fi implementations will include SQL Server and PostgreSQL, but other contributors will have a well-defined interface for other backends e.g. MongoDB.

src/services/EdFi.DataManagementService.Api/AspNetCoreFrontend.cs
CSR2017: @bradbanister 
Thanks for explaining in details.

CSR2017 (approved):

PR #12, opened 2024-02-28
[DMS-5] Adding initial e2e test framework with Playwright

andonyns:
Note: Requires the server running


build.ps1
```diff
@@ -159,7 +159,7 @@ function RunTests {
 }
 
 function UnitTests {
-    Invoke-Execute { RunTests -Filter "*.Tests" }
+    Invoke-Execute { RunTests -Filter "*.Api.Tests" }
```
stephenfuqua, line 162: We will have other unit test projects in the future. Perhaps now is a good time to set a better naming convention.

What if we use `xyz.Tests.<type>`? For example, `EdFi.DataManagement.Api.Tests.Unit` and `EdFi.DataManagement.Api.Tests.E2E`?

src/services/EdFi.DataManagementService.Api.E2E.Tests/EdFi.DataManagementService.Api.E2E.Tests.csproj
```diff
@@ -0,0 +1,44 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
```
stephenfuqua, line 6: I find Microsoft's choice of property name very confusing. This setting does the opposite of what it looks like: it makes it so that reference types are _not nullable_. Which is what we want. But I have to look it up from time to time to confirm that I'm not misremembering.

bradbanister: Totally agree

andonyns: That is confusing

build.ps1
andonyns: I will make the change

.github/workflows/after-pullrequest.yml
```diff
@@ -16,12 +16,20 @@ jobs:
     name: Report Test Results
```
andonyns, line 16: We had a conversation in Meadowlark regarding this action, we should discuss this to see if we want to keep it and handle it this way.

This is failing at the moment since it's not downloading the artifact

build.ps1
```diff
@@ -77,6 +79,7 @@ param(
 )
 
 $solutionRoot = "$PSScriptRoot/src"
+$defaultSolution = "$solutionRoot/EdFi.DataManagementService.sln"
```
andonyns, line 82: Fixes broken build when doing `dotnet build` without parameters since there are two solution files

build.ps1
```diff
@@ -159,7 +162,17 @@ function RunTests {
...
-    Invoke-Execute { RunTests -Filter "*.Tests" }
+    Invoke-Execute { RunTests -Filter "*.Tests.Unit" }
+}
+
+function RunE2E {
+    Invoke-Execute { RunTests -Filter "*.Tests.E2E" }
+}
+
+function E2ETests {
+    Invoke-Step { DockerBuild }
```
andonyns, line 173: The e2e tests require a local docker image, so the script will do both the build and then run the tests

.github/workflows/after-pullrequest.yml
stephenfuqua: I remember that we had a conversation, but I don't remember the substance of that conversation. Uploading test results for future inspection seems like a useful thing.

src/services/EdFi.DataManagementService.Api.Tests.E2E/PlaywrightContext.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+    private static Task<IAPIRequestContext>? _requestContext;
+
+    private string _apiURL = "http://localhost:5198/";  // Backing store
+
+    public string API_URL
+    {
+        get => _apiURL;
+        set => _apiURL = value;
+    }
```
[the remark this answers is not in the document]
andonyns: I will change this

src/services/EdFi.DataManagementService.Api.Tests.E2E/PlaywrightContext.cs
```diff
@@ -0,0 +1,34 @@
...
+    public void Dispose()
+    {
+        _requestContext?.Dispose();
+    }
+
+    public async Task CreateApiContext()
+    {
+        var playwright = await Playwright.CreateAsync();
+
+        _requestContext = playwright.APIRequest.NewContextAsync(new APIRequestNewContextOptions
```
andonyns: @stephenfuqua not sure if I understand the previous question, but this is the initialization of the playwright API

src/services/EdFi.DataManagementService.Api.Tests.E2E/Management/PlaywrightContext.cs
```diff
@@ -0,0 +1,34 @@
+using Microsoft.Playwright;
+
+namespace EdFi.DataManagementService.Api.Tests.E2E.Management;
+
+public class PlaywrightContext
+{
+    private static Task<IAPIRequestContext>? _requestContext;
+
+    private string _apiURL = "http://localhost:5198/";
```
stephenfuqua, line 9: Restating a prior question: what is this URL for? Our app is starting up at 8080 by default (I think), so I'm curious what this URL is.

andonyns: Interesting, my solution always start in 5198. I added this as the default URL, unless changed by TestContainers

andonyns: The HTTP tests also start with 5198 by default https://github.com/Ed-Fi-Alliance-OSS/Data-Management-Service/blob/main/tests/RestClient/test.http

stephenfuqua: You're right, `dotnet run` does use 5198 on my system. Our Docker container uses 8080, which is what I was thinking about. Can change the port when running the container by setting env var `ASPNETCORE_HTTP_PORT`.

CSR2017 (approved):
@andonyns 
Changes look great. Thanks

CSR2017 (approved):

PR #16, opened 2024-03-05
[DMS-13] example simplified failure response body

bradbanister:


bradbanister:
@CSR2017 Here's an example of what I mean about simplifying the failure response body. I understand that this doesn't do 100% of your implementation, but it's simple without being naïve and does everything necessary.

PR #17, opened 2024-03-06
[DMS-6] Add Kubernetes configuration for DMS

andonyns:


src/deployments/kubernetes/app-secret.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: Secret
+metadata:
+    name: app-secret
+type: Opaque
+data:
+    postgres-password: dG9wc2VjdXJl
```
stephenfuqua, line 7: Is this essentially the equivalent of a `.env` file? If so, would it make sense to rename this as `.example` and force people to create their own?

Would help allay some concern about having a hard-coded password in the repository.

andonyns: We can say that this is a .env file for secrets only, see https://kubernetes.io/docs/concepts/configuration/secret/ for more info.
That value is encrypted and will be set unencrypted in the cluster, this is, in a way a .example file.

In Meadowlark, we talked about keeping it for demo purposes, but not sure if we should remove the file or add a warning in the README about the file. https://github.com/Ed-Fi-Exchange-OSS/Meadowlark/pull/337#discussion_r1446380880

src/deployments/kubernetes/postgres-deployment.yaml
```diff
@@ -0,0 +1,54 @@
...
+      serviceName: postgres
+  strategy:
+    type: Recreate
+  template:
+    metadata:
+      labels:
+        serviceName: postgres
+    spec:
+      containers:
+      - name: postgres-ml
```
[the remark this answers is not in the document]
andonyns: fixed

src/deployments/kubernetes/app-secret.yaml
stephenfuqua: I think its ok if someone has to create this file, and its ok if we provide an example. I just want to lower the risk that someone accidentally uses this in a more serious environment without thought. Maybe we add `app-secret.yaml` to the `.gitignore` file.

src/deployments/kubernetes/README.md
```diff
@@ -0,0 +1,66 @@
...
+`docker build -t local/edfi-data-management-service:latest .`
+
+> [!WARNING]
+> The [deployment file](./data-management-service-deployment.yaml#L21) has a policy
+> of Never pull, which means that will use the locally built image instead of trying
+> to pull the image from Docker Hub which is the default behavior, remove
+> once the image is published to the registry.
+
+* Set the terminal in the */deployments/kubernetes* folder.
+* Run `kubectl apply -f .` to apply all files.
```
stephenfuqua, line 28: image

This is what I'm getting right now. Might be a problem specific to my computer or to the MSDF office network.

CSR2017: I am able to follow and set up the application. 
Screenshot 2024-03-07 100552

src/deployments/kubernetes/app-secret.yaml.example
stephenfuqua, line 1: @andonyns please add `app-secret.yaml` to the `.gitignore` file.

stephenfuqua (approved):

PR #25, opened 2024-03-25
[DMS-119] OSSF Scorecard and Dependabot Configuration

stephenfuqua:
Experimenting with running the [OSSF Scorecard](https://scorecard.dev) for improving security. Have tested [in my fork](https://github.com/stephenfuqua/Data-Management-Service/security), though I don't know if others can see the results in that Security tab.

Need to merge this to main to ultimately see it in action. There are some false alarms in my fork because the fork is new and because it does not have a track record of having pull requests.

The "score" part of this is turned off for now. After some experimentation, and cleaning up some obvious problems, we'll consider turning on the public reporting, which is the only place (so far as I can tell) where you can see the public score.


.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,70 @@
...
+      security-events: write
+      # Needed to publish results and get a badge (see publish_results below).
+      # SFUQUA: commented for now, probably nice to have later.
+      # id-token: write
+      # Uncomment the permissions below if installing in a private repository.
+      # contents: read
+      # actions: read
+
+    steps:
+      - name: "Checkout code"
```
andonyns, line 31: Although not important, name doesn't need to be quoted

stephenfuqua: Agree. I copied and pasted from an example and meant to change that. I'm actually annoyed by unnecessary quotations in yml, so I'm going to change that.

andonyns (approved):

PR #28, opened 2024-04-01
[DMS-48] Fake OAuth Endpoint

CSR2017:


msilesgap:
@CSR2017 I'm wondering if the latest commit addressed the issue causing the E2E failures.

src/services/EdFi.DataManagementService.Api/Infrastructure/WebApplicationBuilderExtensions.cs
```diff
@@ -29,6 +32,27 @@ public static void AddServices(this WebApplicationBuilder webAppBuilder)
...
+                options.TokenValidationParameters = new TokenValidationParameters
+                {
+                    ValidateLifetime = true,
+                    ValidateAudience = false,
+                    ValidIssuer = issuer,
+                    IssuerSigningKey = new SymmetricSecurityKey(Encoding.UTF8.GetBytes(signingKeyValue!))
+                };
+                options.RequireHttpsMetadata = false;
+            });
+        webAppBuilder.Services.AddAuthorization();
```
stephenfuqua, line 54: This is taking the OAuth work much farther than the scope of the ticket. The scope was just to accept a POST request and return a string as a token. 

This might be fine. I need to think about it a little bit.

src/services/EdFi.DataManagementService.Api/Modules/TokenModule.cs
```diff
@@ -0,0 +1,85 @@
...
+                string.IsNullOrEmpty(authenticationRequest.Key)
+                || string.IsNullOrEmpty(authenticationRequest.Secret)
+            )
+        )
+        {
+            httpContext.Response.StatusCode = StatusCodes.Status400BadRequest;
+            await httpContext.Response.WriteAsync("Please provide valid key and secret.");
+            return;
+        }
+
```
stephenfuqua, line 41: Recommendation: instead of generating a JWT here, just return status 200 with this payload:

```json
{
  "access_token": "temporary-fake-token",
  "expires_in": 1800,
  "token_type": "bearer"
}
```

src/services/EdFi.DataManagementService.Api/Infrastructure/WebApplicationBuilderExtensions.cs
stephenfuqua: After further thought, please remove this code.

There is a danger that people interacting with the DMS - including ourselves! - might get confused and think that we have implemented more security mechanisms than we really have yet.

For example, once we add authorization, we'll need to be much more specific about authorization rules than this.

Please also keep in mind the long-term vision is that third party software will be generating the JWT. The DMS will need to validate it, but not generate it.

JBrenesSimpat (approved):

JBrenesSimpat (approved):

PR #29, opened 2024-04-01
[DMS-119] Update score board action version

CSR2017:


CSR2017:
Fixed in: https://github.com/Ed-Fi-Alliance-OSS/Data-Management-Service/pull/30

PR #34, opened 2024-04-11
[DMS-126] Update PowerShell Scripts to Avoid Write-Host

msilesgap:
Replace Write-Host method with a new fuction that allows text color as parameter

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -251,9 +251,9 @@ function PushPackage {
     }
 
     if($DryRun){
-        Write-ColorOutput YELLOW ("Dry run enabled, not pushing package.")
+        Write-Information "Dry run enabled, not pushing package."
     } else {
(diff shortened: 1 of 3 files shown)


build.ps1
```diff
@@ -251,9 +251,9 @@ function PushPackage {
     }
 
     if($DryRun){
-        Write-Host "Dry run enabled, not pushing package."
+        Write-Info "Dry run enabled, not pushing package."
     } else {
-        Write-Host "Pushing $PackageFile to $EdFiNuGetFeed"
+        Write-Info ("Pushing $PackageFile to $EdFiNuGetFeed")
```
stephenfuqua, line 256: The use of parenthesis on PowerShell function calls is not required, and in some cases can actually be a problem. I'll test to see if this works.

stephenfuqua: It works fine. We can leave it alone.

stephenfuqua (approved):

PR #45, opened 2024-04-26
[DMS-100] Line endings in Req'n'roll code-behind files

JBrenesSimpat:


.gitattributes
stephenfuqua, line 1: I checked out this branch and ran `dotnet build`. All four files are still showing that they have changes.

stephenfuqua: But it does work correctly if I change `lf` to `crlf` below, and then run `dotnet build`. I'm using PowerShell Core in Windows.

But the code below might work perfectly if building on Linux or MacOSX.

simpat-adam (approved):

PR #46, opened 2024-04-29
[DMS-42] Validate configuration options and return 500 when invalid

simpat-adam:
The approach here is to validate options on startup, but not fail to start up if configuration is invalid. Instead, we will inject a middleware component to log the failures and return 500 on every request.


src/services/EdFi.DataManagementService.Api/Program.cs
```diff
@@ -3,20 +3,43 @@
...
+
+void InjectInvalidConfigurationMiddleware(WebApplication app)
+{
+    try
+    {
+        // Accessing IOptions<T> forces validation
+#pragma warning disable S1481 // Unused local variables should be removed
+        var appSettings = app.Services.GetRequiredService<IOptions<AppSettings>>().Value;
+        var connectionStrings = app.Services.GetRequiredService<IOptions<ConnectionStrings>>().Value;
+#pragma warning restore S1481 // Unused local variables should be removed
```
stephenfuqua, line 34: Alternately, instead of the `pragma`, you can discard the values by using `_ =` instead of `var appSettings = `.

src/EdFi.DataManagementService.Api.Tests.Integration/ConfigurationTests.cs
```diff
@@ -0,0 +1,110 @@
...
+
+        [TearDown]
+        public void Teardown()
+        {
+            _factory.Dispose();
+        }
+
+        [TestFixture]
+        public class Should_respond_with_internal_server_error_when_missing_authentication_service
+            : Given_a_configuration_with_invalid_app_settings
```
stephenfuqua, line 47: Minor nitpicking: "when" in Gherkin language should be an action statement "given this situation, _when doing this thing_, then (should) have this response".

Perhaps something like `When_requesting_any_endpoint_Should_return_InternalServerError`.

src/services/EdFi.DataManagementService.Api/Configuration/AppSettings.cs
```diff
@@ -3,12 +3,23 @@
...
+public class AppSettingsValidator : IValidateOptions<AppSettings>
 {
-    public class AppSettings
+    public ValidateOptionsResult Validate(string? name, AppSettings options)
     {
-        public int BeginAllowedSchoolYear { get; set; }
-        public int EndAllowedSchoolYear { get; set; }
-        public required string AuthenticationService { get; set; }
+        return options.AuthenticationService == null
+            ? ValidateOptionsResult.Fail("Missing required AppSettings value: AuthenticationService")
```
stephenfuqua, line 22: Interesting, in the log this comes through as an unhandled exception. Obviously we _did_ handle it - resulting in a log message and a 500. It is unsettling to see a full stack trace and "An uncaught error has occurred", making it look like we forgot to put in some important error handling.

src/services/EdFi.DataManagementService.Api/Configuration/ConnectionStrings.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    public required string DatabaseConnection { get; set; }
+}
+
+public class ConnectionStringsValidator : IValidateOptions<ConnectionStrings>
+{
+    public ValidateOptionsResult Validate(string? name, ConnectionStrings options)
+    {
+        return options.DatabaseConnection == null
+            ? ValidateOptionsResult.Fail("Missing required ConnectionStrings value: DatabaseConnection")
```
stephenfuqua, line 20: This case ends up processing differently for some reason: does not have the stack trace. It _does_ produce a 500, but it _also_ provides the proper GET response on the root endpoint.

image

src/services/EdFi.DataManagementService.Api/Configuration/AppSettings.cs
```diff
@@ -3,12 +3,23 @@
...
+
+public class AppSettingsValidator : IValidateOptions<AppSettings>
 {
-    public class AppSettings
+    public ValidateOptionsResult Validate(string? name, AppSettings options)
     {
-        public int BeginAllowedSchoolYear { get; set; }
-        public int EndAllowedSchoolYear { get; set; }
-        public required string AuthenticationService { get; set; }
+        return options.AuthenticationService == null
```
stephenfuqua, line 21: Suggest changing to detect empty string or spaces, in addition to null.

src/services/EdFi.DataManagementService.Api/Program.cs
```diff
@@ -3,20 +3,43 @@
...
+    }
+    catch (OptionsValidationException ex)
+    {
+        app.UseMiddleware<InvalidConfigurationMiddleware>(ex.Failures);
+    }
```
stephenfuqua, line 39: I was stepping through the debugger, and I realized that catching the exception right here - before `app.Run()` executes - means that the web server is still running and all of the routes will still be active.

src/services/EdFi.DataManagementService.Api/Program.cs
simpat-adam: Addressed.

src/EdFi.DataManagementService.Api.Tests.Integration/ConfigurationTests.cs
simpat-adam: Renamed.

src/services/EdFi.DataManagementService.Api/Configuration/ConnectionStrings.cs
simpat-adam: Oops. I didn't have short circuit in the middleware that logs the error and returns 500. Addressed.

src/services/EdFi.DataManagementService.Api/Configuration/AppSettings.cs
simpat-adam: Done.

src/services/EdFi.DataManagementService.Api/Program.cs
simpat-adam: All of the routes are still active so we don't get 404's. This middleware is supposed to set the response to 500 and short circuit the rest of the pipeline.

stephenfuqua:
Temporarily closing and then will re-open to force re-run of the actions.

src/services/EdFi.DataManagementService.Api.Tests.E2E/Management/ContainerSetup.cs
```diff
@@ -17,7 +17,6 @@ public async Task<string> SetupDataManagement()
         var dockerImage = new ContainerBuilder()
             .WithImage(imageName)
             .WithPortBinding(8080)
-            .WithWaitStrategy(Wait.ForUnixContainer().UntilHttpRequestIsSucceeded(r => r.ForPort(8080)))
```
simpat-adam, line 20: When we have invalid appsettings, this WaitStrategy will cause the E2E test suite to run indefinitely, because it is waiting for a successful Http response, which it is never going to get. Removing this line will simply fail the tests in that scenario.

stephenfuqua (approved):

PR #47, opened 2024-04-30
[DMS-137] Do not run PR workflow on main

simpat-adam:


.github/workflows/on-pullrequest.yml
```diff
@@ -6,12 +6,6 @@
 name: On Pull Request
 
 on:
-  push:
-    # Running on push to main to support CodeQL on C#
-    branches:
-      - main
-    paths:
-      - "**/*.cs"
```
stephenfuqua, line 14: According to the comment above, this was done on purpose. At one point, CodeQL required that the default branch be listed. Please review CodeQL configuration notes before we accept this.

It is the dependency review step that is failing, and definitely cannot run on push to `main`. So that step probably needs to be pulled into an alternate workflow, or have a condition added so that it only runs on a pull request.

stephenfuqua: @simpat-adam suggestion: restore this, and [look at this example](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-contexts) to disable the Dependency Review step when not a pull request.

We _do_ want the Unit Tests and CodeQL to run on push to `main`. If PR 1 has green tests but sits around for a few days, and PR 2 comes in and is merged first, there is a small possibility that (a) a rebase is not required for merge and (b) that a conflict occurs. Therefore, best to let these tests run again after the merge.

stephenfuqua (approved):

PR #52, opened 2024-05-07
[DMS-35] Reference extraction

bradbanister:
Not part of the ticket, but minor reorganization of backend interface definitions into core.

Also some frontend query parameter definition and some backend query handler definition, given "Get All" is coming soon.  Didn't implement the connecting middleware however.

### DIFF
--- a/src/services/EdFi.DataManagementService.Api.Tests.Unit/Core/ApiSchema/ResourceSchema/ExtractDocumentReferencesTests.cs
+++ b/src/services/EdFi.DataManagementService.Api.Tests.Unit/Core/ApiSchema/ResourceSchema/ExtractDocumentReferencesTests.cs
@@ -1,0 +1,402 @@
+// SPDX-License-Identifier: Apache-2.0
+// Licensed to the Ed-Fi Alliance under one or more agreements.
+// The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+// See the LICENSE and NOTICES files in the project root for more information.
+
+using System.Text.Json.Nodes;
+using EdFi.DataManagementService.Api.Core.ApiSchema;
+using EdFi.DataManagementService.Api.Core.Model;
+using FluentAssertions;
+using NUnit.Framework;
+using static EdFi.DataManagementService.Api.Tests.Unit.TestHelper;
+
+namespace EdFi.DataManagementService.Api.Tests.Unit.Core.ApiSchema;
+
+[TestFixture]
+public class ExtractDocumentReferencesTests
+{
+    public static ApiSchemaDocument BuildApiSchemaDocument()
+    {
+        return new ApiSchemaBuilder()
+            .WithStartProject()
+            .WithStartResource("Section")
+            .WithIdentityJsonPaths(
+                [
+                    "$.courseOfferingReference.localCourseCode",
+                    "$.courseOfferingReference.schoolId",
+                    "$.courseOfferingReference.schoolYear",
+                    "$.courseOfferingReference.sessionName",
+                    "$.sectionIdentifier"
+                ]
+            )
+            .WithStartDocumentPathsMapping()
+            .WithDocumentPathScalar("SectionIdentifier", "$.sectionIdentifier")
+            .WithDocumentPathReference(
+                "CourseOffering",
+                [
+                    new("$.localCourseCode", "$.courseOfferingReference.localCourseCode"),
+                    new("$.schoolReference.schoolId", "$.courseOfferingReference.schoolId"),
+                    new("$.sessionReference.schoolYear", "$.courseOfferingReference.schoolYear"),
+                    new("$.sessionReference.sessionName", "$.courseOfferingReference.sessionName")
+                ]
+            )
+            .WithDocumentPathReference(
+                "ClassPeriod",
+                [
+                    new("$.classPeriodName", "$.classPeriods[*].classPeriodReference.classPeriodName"),
+                    new("$.schoolReference.schoolId", "$.classPeriods[*].classPeriodReference.schoolId"),
+                ]
+            )
+            .WithEndDocumentPathsMapping()
+            .WithEndResource()
+            .WithEndProject()
+            .ToApiSchemaDocument();
+    }
+
+    [TestFixture]
+    public class Given_extracting_document_references_with_one_as_scalar_and_another_as_collection
+        : ExtractDocumentReferencesTests
+    {
+        public DocumentReference[] documentReferences = [];
+
+        [SetUp]
+        public void Setup()
+        {
+            ApiSchemaDocument apiSchemaDocument = BuildApiSchemaDocument();
+            ResourceSchema resourceSchema = BuildResourceSchema(apiSchemaDocument, "sections");
+
+            documentReferences = resourceSchema.ExtractDocumentReferences(
+                JsonNode.Parse(
+                    """
+                    {
+                        "sectionIdentifier": "Bob",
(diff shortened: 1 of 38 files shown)


src/services/EdFi.DataManagementService.Api.Tests.Unit/Core/ApiSchema/ResourceSchema/ExtractDocumentReferencesTests.cs
```diff
@@ -0,0 +1,402 @@
...
+using EdFi.DataManagementService.Api.Core.ApiSchema;
+using EdFi.DataManagementService.Api.Core.Model;
+using FluentAssertions;
+using NUnit.Framework;
+using static EdFi.DataManagementService.Api.Tests.Unit.TestHelper;
+
+namespace EdFi.DataManagementService.Api.Tests.Unit.Core.ApiSchema;
+
+[TestFixture]
+public class ExtractDocumentReferencesTests
```
bradbanister, line 16: Tests for DMS-35

src/services/EdFi.DataManagementService.Api/Core/Backend/DeleteRequest.cs
```diff
@@ -4,7 +4,7 @@
 // See the LICENSE and NOTICES files in the project root for more information.
 using EdFi.DataManagementService.Api.Core.Model;
 
-namespace EdFi.DataManagementService.Api.Backend;
+namespace EdFi.DataManagementService.Api.Core.Backend;
```
bradbanister, line 7: Moved the backend interface definitions into core.

src/services/EdFi.DataManagementService.Api/Core/Backend/IQueryHandler.cs
```diff
@@ -0,0 +1,16 @@
...
+// Licensed to the Ed-Fi Alliance under one or more agreements.
+// The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+// See the LICENSE and NOTICES files in the project root for more information.
+namespace EdFi.DataManagementService.Api.Core.Backend;
+
+/// <summary>
+/// The handler DMS Core uses to perform document queries.
+/// </summary>
+public interface IQueryHandler
+{
```
bradbanister, line 11: Added backend interface for query handler

src/services/EdFi.DataManagementService.Api/Core/Backend/QueryRequest.cs
```diff
@@ -0,0 +1,34 @@
...
+
+using EdFi.DataManagementService.Api.Core.Model;
+using Microsoft.Extensions.Primitives;
+
+namespace EdFi.DataManagementService.Api.Core.Backend;
+
+/// <summary>
+/// A query request to a query handler
+/// </summary>
+public record QueryRequest(
```
bradbanister, line 14: Backend query interface messages

src/services/EdFi.DataManagementService.Api/Core/Model/FrontendRequest.cs
```diff
@@ -26,6 +27,10 @@ public record FrontendRequest(
     /// </summary>
     JsonNode? Body,
     /// <summary>
+    /// Query parameters provided by the frontend service as a dictionary.
+    /// </summary>
+    Dictionary<string, StringValues> QueryParameters,
```
bradbanister, line 32: Added query parameters to frontend request

src/services/EdFi.DataManagementService.Api/Backend/SuccessDocumentStoreRepository.cs
```diff
@@ -54,4 +55,18 @@ public async Task<DeleteResult> DeleteDocumentById(DeleteRequest deleteRequest)
         );
         return await Task.FromResult<DeleteResult>(new DeleteResult.DeleteSuccess());
     }
+
+    public async Task<QueryResult> QueryDocuments(QueryRequest queryRequest)
```
bradbanister, line 59: Query documents implementation for Success document store

src/services/EdFi.DataManagementService.Api/Core/ApiSchema/Extensions/JsonHelperExtensions.cs
```diff
@@ -56,6 +56,65 @@ public static class JsonHelperExtensions
         }
     }
 
+    /// <summary>
+    /// Helper to go from an array JSONPath selection directly to the selected JsonNodes.
+    /// Returns an empty array if none are selected.
+    /// </summary>
+    public static IEnumerable<JsonNode> SelectNodesFromArrayPath(
```
bradbanister, line 63: JsonPath array selectors to go with the scalar selectors

src/services/EdFi.DataManagementService.Api/Backend/SuccessDocumentStoreRepository.cs
```diff
@@ -11,7 +12,7 @@ namespace EdFi.DataManagementService.Api.Backend;
 /// A document store repository that does nothing but return success
 /// </summary>
 public class SuccessDocumentStoreRepository(ILogger<SuccessDocumentStoreRepository> _logger)
```
stephenfuqua, line 14: We should probably pull this repository into another library. Under a different ticket, not this one.

src/services/EdFi.DataManagementService.Api/Core/ApiSchema/DocumentPath.cs
```diff
@@ -0,0 +1,158 @@
...
+///
+/// </summary>
+public class DocumentPath(JsonNode _documentPathsNode)
+{
+    private readonly Lazy<bool> _isDescriptor =
+        new(() =>
+        {
+            return _documentPathsNode["isDescriptor"]?.GetValue<bool>()
+                ?? throw new InvalidOperationException(
+                    "Expected isDescriptor to be on DocumentPath, not a reference, or invalid ApiSchema"
```
stephenfuqua, line 43: Would it make sense to create an extension method or a private function for getting a value from a `JsonNode`? I see a lot of repeated code here.

But, it is relatively trivial repeated code. If this is the only file that has it, then it would be easy enough to change all of the lines if we wanted a different exception, for example.

My conclusion: could do it, but I don't feel as strongly about it as I would have a few years ago, when "don't repeat yourself" was still a sacred commandment to me. After the `left-pad` npm debacle I've definitely softened up a bit on DRY for utility functions 😆. But if the same functionality were spread across several files, then it still might make sense to have a _local_ shared function, though not a shared _library_.

stephenfuqua:
```csharp
public static T Get<T>(this JsonNode node, string key)
{
     return node[key]?.GetValue<T>()
                ?? throw new InvalidOperationException(
                    $"ApiSchema is invalid: Expected {key} to be on DocumentPath."
}
```

Looking carefully as I read down the list, I see that the messages have some nuance to them. Maybe you find a root level, but then you don't find something below it. So the message says "couldn't find this XYZ inside the ABC", not just "i couldn't find XYZ". Easy to introduce that level of detail in the extension method above, but then it doesn't really take you very far compared to reading / maintaining the original.

bradbanister: Right. The message differences are valuable in diagnosing the actual problem. Also, many of the return types are record types of ours, so `GetValue<T>()` won't work.

CSR2017 (approved):

PR #53, opened 2024-05-07
[DMS-159] Resource Queries and [DMS-160] Paging

elopezgap:
Addition of file ResourceQueriesQueryString.feature.disabled


src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
```diff
@@ -0,0 +1,202 @@
...
+            | 6            | UT Austin College of Education Under Graduate        | [ "Postsecondary" ]       | [ "Educator Preparation Provider" ]  |
+    When sending a GET request to "/ed-fi/schools?nameOfInstitution=UT+AUSTIN+COLLEGE+OF+EDUCATION+UNDER+GRADUATE" 
+            """
+            """
+    Then it should respond with 200
+    And "nameOfInstitution: UT Austin College of Education Under Graduate" 
+
+
+  @ignore
+  Scenario: Ensure clients can GET information filtering by totalCount
```
stephenfuqua, line 27: You already have this in another file, so I think you remove this scenario.

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
```diff
@@ -0,0 +1,202 @@
+Feature: Query String handling for GET requests for Resource Queries
```
stephenfuqua, line 1: The file name says this is about paging. Can you change the feature name above to reflect this? Maybe "Paging support for GET requests for Ed-Fi Resources" or something like that.

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
```diff
@@ -0,0 +1,202 @@
...
+
+@ignore
```
stephenfuqua, line 3: These two case sensitivity scenarios are good, but should move to the other file since they are not related to paging.

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
```diff
@@ -0,0 +1,202 @@
...
+  
+  @ignore
+  Scenario: Ensure clients can't GET information when filtering using invalid values
+    When sending a GET request to "/ed-fi/schools?limit=-1" using values as
+            | Values                       |
+            | -1                           |
+            | 'zero'                       |
+            | '5; select * from users'     |
+            | '0)'                         |
+            | '1%27'                       |
```
stephenfuqua, line 45: The values above are all different sample values that could be used in `?limit=<value>`.  I think you'll want to rewrite this using a [Scenario Outline](https://docs.reqnroll.net/latest/gherkin/gherkin-reference.html#scenario-outline). The table will end up moving below with the label `Examples`, and you'll change line 39 to have `?limit=<value>`. Be sure to change "Values" to "Value" in the table (since each row is just a single value, the column header makes more sense as "Value" than "Values").

FYI, the third, fourth, and fifth examples are all SQL injection attacks, helping us ensure that the API is well secured.

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
```diff
@@ -0,0 +1,202 @@
...
+             """
+              {
+                "detail": "Data validation failed. See 'validationErrors' for details.",
+                "type": "urn:ed-fi:api:bad-request:data",
+                "title": "Data Validation Failed",
+                "status": 400,
+                "correlationId": "9e4f71dd-013c-41ee-9eaf-9c3dc368d206",
+                "validationErrors": {
+                  "$.Limit": [
+                    "The value 'zero' is not valid for Limit."
```
stephenfuqua, line 57: When using a scenario outline, you'll want to use `The value '<value>' is not valid for Limit.` here.

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
```diff
@@ -0,0 +1,202 @@
...
+            | 6            | UT Austin College of Education Under Graduate   |
+            | 7            | UT Austin Extended Campus                       |
+            | 255901001    | Grand Bend High School                          |
+            | 255901044    | Grand Bend Middle School                        |
+    When sending a GET request to "/ed-fi/schools?offset=3&limit=5"
+            """
+            """
+    Then it should respond with status 200
+    And schools returned:
+            | 255901001    | Grand Bend High School                          |
```
stephenfuqua, line 77: Does this table also need the column headers?

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
```diff
@@ -0,0 +1,202 @@
...
+                    "The value 'zero' is not valid for Limit."
+                  ]
+                }
+              }
+             """ 
+  
+  @ignore
+  Scenario: Ensure clients can get information when filtering by limit and and a valid offset 
+    Given 5 schools exist:
+            | schoolId     | nameOfInstitution                               |
```
stephenfuqua, line 66: I think this table also need `gradeLevels` and the categories, so that it matches what we had in another feature file. Otherwise, it will be confusing in the C# step definition, because we'll have to deal with two different table layouts.

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
```diff
@@ -0,0 +1,202 @@
...
+             """  
+
+  ##I need a few more details on this scenario
+  @ignore
+  Scenario: Ensure clients can't GET information when querying with filter and offset using limit without offset
+    Given these schools exist:
+            | schoolId     | nameOfInstitution                               | 
+            | 5            | School with max edorgId value                   | 
+            | 6            | UT Austin College of Education Under Graduate   |
+    When sending a GET request to "/ed-fi/schools?limit=-6&offset=-1"
```
stephenfuqua, line 127: "limit without offset" implies having `?offset=1` (should be a positive number, so that it is an acceptable offset), without providing any `limit` in the query string.

This seems familiar from Meadowlark, but I don't remember why we had this. It should be fine to send a query with an offset but no limit. Please change the scenario to be successful when using an `offset` without also using a `limit` in the query string.

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
```diff
@@ -0,0 +1,202 @@
...
+  Scenario: Ensure clients can GET information when filtering with limits and properties
+    Given these schools exist:
+            | schoolId     | nameOfInstitution                               | 
+            | 5            | School with max edorgId value                   | 
+            | 6            | UT Austin College of Education Under Graduate   |
+    When sending a GET request to "/ed-fi/schools?nameOfInstitution=UT+Austin+College+of+Education+Under+Graduate&limit=2" 
+            """
+            """
+    Then it should respond with 200
+    And response body is
```
stephenfuqua, line 155: There will be some cases where we need to look closely at the entire response body. However, the response body below has some problems:

1. We won't know in advance what the `id` will be.
2. Although we only expect one response, it should be in an array: the GET queries always returns arrays, even if there are _no_ items or just _one_ item in the array.
3. The value from the ODS/API includes an address, but your setup in the table above does not have any addresses.

There are ways we could make this work. But for the scope of this feature file, the details are probably not important. We want to make sure that the response contains exactly one item (no more, no less), and we want to make sure that it is the _right_ item. Since there are two schools in the database, we can simply ensure that the correct `schoolId` is in the response. Suggestion: instead of `And response body is`:

```
And response array contains 1 items
And the response array contains schoolId=6
```

(Although "1 items" is not grammatically correct, this allows us to write a single C# step definition with a variable for the actual number, and the word "items" will make sense in other contexts where we expect multiple).

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
```diff
@@ -0,0 +1,202 @@
...
+            | '1%27'                       | 
+    Then it should respond with 400
+    And response body is
+             """
+             {
+                "detail": "The limit parameter was incorrect.",
+                "type": "urn:ed-fi:api:bad-request:parameter",
+                "title": "Parameter Validation Failed",
+                "status": 400,
+                "correlationId": "b68e4051-8ed4-4055-9e59-4f9c2ef5f4ce",
```
stephenfuqua, line 113: Please use `"correlationId": null` here. We don't know what the actual correlationId will be in advance. Whoever writes the C# will have to remember to ignore the `null`.

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
elopezgap: Removed

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
elopezgap: Removed

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
elopezgap: Complete

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
elopezgap: Complete

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
```diff
@@ -0,0 +1,202 @@
...
+            | 5            | School with max edorgId value                   | 
+            | 6            | UT Austin College of Education Under Graduate   |
+            | 7            | UT Austin Extended Campus                       |
+            | 255901001    | Grand Bend High School                          |
+            | 255901044    | Grand Bend Middle School                        |
+    When sending a GET request to "/ed-fi/schools?offset=6&limit=5"
+            """
+            """
+    Then it should respond with status 200
+    And "[]"
```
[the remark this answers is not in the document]
elopezgap: Complete

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
elopezgap: Complete

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
elopezgap: Done

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
elopezgap: Done

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
elopezgap: Done

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
elopezgap: Just need to make sure the behavior of this scenario because the test referenced in [this link](https://github.com/Ed-Fi-Exchange-OSS/Meadowlark/blob/main/Meadowlark-js/tests/e2e/scenarios/QueryStringValidation.test.ts), has a response body with 400 code and an error message indicating that "Limit must be provided when using offset".
And checking in ODS/API v6.2, v7.1 and v7.2, making a GET request using this query string (/ed-fi/schools?offset=1), throws all the schools present in the database

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
elopezgap: Needs to confirm the behavior of this request because when making a GET request using this query (/ed-fi/schools?nameOfInstitution=UT+Austin+College+of+Education+Under+Graduate&limit=2), I am getting the the results of the limit, and it's not filtering by nameOfInstitution

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
```diff
@@ -0,0 +1,142 @@
+Feature: Query String handling for GET requests for Resource Queries
+
+  @ignore
+  Scenario: Ensure clients can GET information when querying by valid date
+    Given there is a valid Begin date
+            | beginDate    |
+            | 2024-04-09   |
```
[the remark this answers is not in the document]
elopezgap: Done

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
```diff
@@ -0,0 +1,142 @@
...
+    When sending a GET request to "/ed-fi/academicWeeks?beginDate=2024-04-09"  
+            """
+            """
+    Then it should respond with 200
+    And response body
```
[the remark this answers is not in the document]
elopezgap: Complete

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
```diff
@@ -0,0 +1,142 @@
...
+            """
+            {
+                "id": "a930a504a63f46f898a1a3f88f8414a4",
+                "schoolReference": {
+                "schoolId": 255901001,
+                "link": {
+                    "rel": "School",
+                    "href": "/ed-fi/schools/20ec19e5070245128a30fdcc6925bb09"
```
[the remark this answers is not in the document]
elopezgap: Complete

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
```diff
@@ -0,0 +1,106 @@
...
+    When sending a GET request to "/ed-fi/schools?offset=6&limit=5"
+            """
+            """
+    Then it should respond with status 200
+    And response array contains 0 items
+
+  @ignore
+  Scenario: Ensure clients can't GET information when filtering by limit and offset using invalid values
+    When sending a GET request to "/ed-fi/schools?limit=-6&offset=<Value>" 
+
```
msilesgap, line 59: I think this line can be removed

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
```diff
@@ -0,0 +1,105 @@
+Feature: Query String handling for GET requests for Resource Queries
+
+  Background:
```
msilesgap, line 3: Could you confirm if the Background section is required?

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
```diff
@@ -0,0 +1,105 @@
...
+                "correlationId": null,
+                "validationErrors": {
+                    "$.beginDate": [
+                    "The value '<beginDate>' is not valid for BeginDate."
+                    ]
+                }
+            }
+            """
+
+  @ignoreScenario
```
msilesgap, line 48: Replace this with the right tag "@ignore"

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
```diff
@@ -0,0 +1,105 @@
...
+            | 255901001 | Grand Bend High School                         | [ "Tenth grade" ]         | [ "School" ]                             |
+    And there is an unspecified Begin date
+            | weekIdentifier | beginDate  | endDate    | totalInstructionalDays |
+            | string         | 2027-04-09 | 2027-04-09 | 0                      |
+    When sending a GET request to "/ed-fi/academicWeeks?beginDate=<beginDate>"  
+            """
+            """
+    Then it should respond with 200
+    And response body is an empty array    
+
```
msilesgap, line 61: Extra line could be removed

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
```diff
@@ -0,0 +1,106 @@
+Feature: Paging Support for GET requests for Ed-Fi Resources
+
+  
+  @ignore
+  Scenario: Ensure clients can't GET information when filtering using invalid values
+    When sending a GET request to "/ed-fi/schools?limit=<Value>"
+             """
+             """
```
stephenfuqua, line 8: Please remove lines 7 and 8. A GET request will never have a body.

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
```diff
@@ -0,0 +1,106 @@
...
+    Then it should respond with 400
+    And response message includes
+             | The value '<Value>' is not valid for Limit. |
```
stephenfuqua, line 11: Let's use the actual response from the ODS/API here:
```json
{
  "detail": "The limit parameter was incorrect.",
  "type": "urn:ed-fi:api:bad-request:parameter",
  "title": "Parameter Validation Failed",
  "status": 400,
  "correlationId": null,
  "errors": [
    "Limit must be omitted or set to a value between 0 and 500."
  ]
}
```

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
```diff
@@ -0,0 +1,106 @@
...
+            | 255901044 | Grand Bend Middle School                       | [ "Seventh grade" ]       | [ "School" ]                             | 
+          
+    When sending a GET request to "/ed-fi/schools?offset=3&limit=5"
+            """
+            """
+    Then it should respond with status 200
+    And the response array contains 2 items
+            #| schoolId   | nameOfInstitution             | gradeLevels               | educationOrganizationCategories          |
+            #| 255901001  | Grand Bend High School        | [ "Tenth grade" ]         | [ "School" ]                             |
+            #| 255901044  | Grand Bend Middle School      | [ "Seventh grade" ]       | [ "School" ]                             |
```
stephenfuqua, line 38: Why are these three lines commented out?

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
stephenfuqua: Then first scenario covered invalid `limit`. This scenario should cover invalid `offset`. Therefore the `limit` needs to be changed to a number between 0 and 500. The response body should be similar to the Limit example above, except that it will say "Offset" instead of "Limit".

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
```diff
@@ -0,0 +1,106 @@
...
+            | Value                        |
+            | -1                           |
+            | 'zero'                       |
+            | '5; select * from users'     |
+            | '0)'                         |
+            | '1%27'                       |           
+
+  ##I need a few more details on this scenario
+  @ignore
+  Scenario: Ensure clients can't GET information when querying using an offset without providing any limit in the query string
```
stephenfuqua, line 73: Discussed this scenario on Slack and in the ticket: the ODS/API, unlike Meadowlark, _does_ allow this.

Therefore the Data Management Service should allow it. Please change the expected response code to 200 and the response body should contain two records.

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
stephenfuqua: It might be useful to remove all of the "given a school exists" statements below, and have a single Background section with "Given a school exists".

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
```diff
@@ -0,0 +1,105 @@
...
+  @ignore
+  Scenario: Ensure clients can GET information when querying by valid date
+    Given a school exists
+            | schoolId   | nameOfInstitution                             | gradeLevels               | educationOrganizationCategories          |
+            | 255901001 | Grand Bend High School                         | [ "Tenth grade" ]         | [ "School" ]                             |    
+    And there is a valid AcademicWeek using a valid BeginDate
+            | weekIdentifier | beginDate  | endDate    | totalInstructionalDays |
+            | string         | 2024-04-09 | 2024-04-09 | 0                      |
+    When sending a GET request to "/ed-fi/academicWeeks?beginDate=<beginDate>"
```
stephenfuqua, line 14: This should actually fail, since `<beginDate>` is not a real date, and there is no example to be pasted in below.

So let's change this line by replacing `<beginDate>` with `2024-04-09`.

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
```diff
@@ -0,0 +1,105 @@
...
+  
+
+  @ignore
+  Scenario: Ensure clients can GET information when querying by valid date
+    Given a school exists
+            | schoolId   | nameOfInstitution                             | gradeLevels               | educationOrganizationCategories          |
+            | 255901001 | Grand Bend High School                         | [ "Tenth grade" ]         | [ "School" ]                             |    
+    And there is a valid AcademicWeek using a valid BeginDate
+            | weekIdentifier | beginDate  | endDate    | totalInstructionalDays |
+            | string         | 2024-04-09 | 2024-04-09 | 0                      |
```
stephenfuqua, line 13: "string" is a valid week identifier, but to another reviewer it might look very strange. Let's put something more realistic, like "Week One".

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
```diff
@@ -0,0 +1,105 @@
...
+    And response contains "weekIdentifier: "string" "
+
+  @ignoreScenario 
+  Scenario: Ensure clients can't GET information when querying by invalid date
+    Given a school exists
+            | schoolId   | nameOfInstitution                             | gradeLevels               | educationOrganizationCategories          |
+            | 255901001 | Grand Bend High School                         | [ "Tenth grade" ]         | [ "School" ]                             |
+    And there is a valid AcademicWeek using an invalid BeginDate   
+            | weekIdentifier | beginDate | endDate    | totalInstructionalDays |
+            | string         | -04-09    | 2024-04-09 | 0                      |
```
stephenfuqua, line 27: The "Given" section must insert valid data into the database. Therefore the `beginDate` above needs to be a real one.

It is on the "When" line that we have an invalid date. In this case you have a perfectly good line, since `<beginDate>` is, in fact, invalid 😁 .

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
```diff
@@ -0,0 +1,105 @@
...
+
+  @ignoreScenario 
+  Scenario: Ensure clients can GET information when querying by date out of specified dates
+    Given a school exists
+            | schoolId   | nameOfInstitution                             | gradeLevels               | educationOrganizationCategories          |
+            | 255901001 | Grand Bend High School                         | [ "Tenth grade" ]         | [ "School" ]                             |
+    And there is an unspecified Begin date
+            | weekIdentifier | beginDate  | endDate    | totalInstructionalDays |
+            | string         | 2027-04-09 | 2027-04-09 | 0                      |
+    When sending a GET request to "/ed-fi/academicWeeks?beginDate=<beginDate>"
```
stephenfuqua, line 56: The query date above needs to be a valid value. It can be any date _other than_ `2027-04-09`.

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
elopezgap: Complete

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
elopezgap: Complete

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
elopezgap: Complete

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
elopezgap: Complete

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
elopezgap: Complete

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
elopezgap: Removed

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
elopezgap: Complete

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
elopezgap: Complete

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
elopezgap: Complete

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
elopezgap: Understood.
Applied

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
```diff
@@ -0,0 +1,75 @@
...
+             Then it should respond with 200
+              And response body contains ""weekIdentifier": "string""
+
+
+        @ignore
+        Scenario: Ensure clients can GET information when querying by integer identifier
+            Given there is an invalid Begin date
+                  | weekIdentifier | beginDate  | endDate    | totalInstructionalDays |
+                  | Week test 1    | 2024-05-13 | 2024-05-13 | 1                      |
+             When sending a GET request to "/ed-fi/academicWeeks?weekIdentifier=<totalInstructionalDays>"
```
stephenfuqua, line 73: Use the actual value here: `?totalInstructionalDays=1`.

The Scenario says "querying by integer identifier", but that is not accurate. This is not querying by an identifier.

Line 75 should have a count 1. Suggest using single apostrophe around the string instead of quotation marks, because it is hard to work with `""`. That is, change to `'"totalInstructionalDays": 1'` (also remove the comma, because it could be the last thing in the JSON response).

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
```diff
@@ -0,0 +1,75 @@
...
+                  | Week One       | 2024-05-15 | 2024-05-15 | 2                      |
+             When sending a GET request to "/ed-fi/academicWeeks?beginDate=2024-04-09"
+             Then it should respond with 200
+              And response contains "weekIdentifier: "string" "
+
+        @ignore
+        Scenario: Ensure clients can't GET information when querying by invalid date
+            Given there is a valid AcademicWeek using an invalid BeginDate
+                  | weekIdentifier | beginDate | endDate    | totalInstructionalDays |
+                  | 2024-05-15     | -04-09    | 2024-04-09 | 0                      |
```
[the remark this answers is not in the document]
elopezgap: Complete

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
```diff
@@ -0,0 +1,75 @@
...
+             When sending a GET request to "/ed-fi/academicWeeks?beginDate=<beginDate>&endDate=<endDate>"
+             Then it should respond with 200
+              And response body is an empty array
+
+        @ignore
+        Scenario: Ensure clients can GET information when querying by string identifier
+            Given there is a valid Begin date
+                  | weekIdentifier | beginDate  | endDate    | totalInstructionalDays |
+                  | string         | 2024-04-09 | 2024-04-09 | 0                      |
+             When sending a GET request to "/ed-fi/academicWeeks?weekIdentifier=<weekIdentifier>"
```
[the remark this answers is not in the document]
elopezgap: Complete

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
elopezgap: Complete

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
```diff
@@ -0,0 +1,72 @@
...
+        Examples:
+                  | Value                    |
+                  | -1                       |
+                  | 'zero'                   |
+                  | '5; select * from users' |
+                  | '0)'                     |
+                  | '1%27'                   |
+
+        @ignore
+        Scenario: Ensure clients can't GET information when querying using an offset without providing any limit in the query string
```
stephenfuqua, line 55: This scenario actually proves the opposite: that we _can_ get information. Not "can't" 😄

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
```diff
@@ -0,0 +1,72 @@
...
+             When sending a GET request to "/ed-fi/schools?offset=6&limit=5"
+             Then it should respond with status 200
+              And response array contains 0 items
+
+        @ignore
+        Scenario: Ensure clients can't GET information when filtering by limit and offset using invalid offset
+             When sending a GET request to "/ed-fi/schools?limit=6&offset=<Value>"
+             Then it should respond with 400
+              And response body contains
+                  | Offset must be omitted or set to a value between 0 and 500. |
```
stephenfuqua, line 45: It is better to have the complete expected JSON here, rather than just a phrase.

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
```diff
@@ -0,0 +1,58 @@
...
+            Given a school exists
+                  | schoolId  | nameOfInstitution      | gradeLevels       | educationOrganizationCategories |
+                  | 255901001 | Grand Bend High School | [ "Tenth grade" ] | [ "School" ]                    |
+              And there is a valid AcademicWeek
+                  | weekIdentifier | beginDate  | endDate    | totalInstructionalDays |
+                  | Week One       | 2024-05-15 | 2024-05-15 | 2                      |
+
+        @ignore
+        Scenario: Ensure clients can GET information when querying by valid date
+             When sending a GET request to "/ed-fi/academicWeeks?beginDate={beginDate}"
```
msilesgap, line 13: I think is better if we replace the {beginDate} with the current value used in the Background section since this already exists

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
elopezgap: Done

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
elopezgap: Done

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
```diff
@@ -0,0 +1,58 @@
...
+
+        @ignore
+        Scenario: Ensure clients can GET information when querying by date out of specified dates
+             When sending a GET request to "/ed-fi/academicWeeks?beginDate=1970-04-09"
+             Then it should respond with 200
+              And response body is an empty array
+
+        @ignore
+        Scenario: Ensure clients can GET information when querying by begin date and end date
+            Given there is an invalid Begin date
```
msilesgap, line 34: I think the Given can be removed here; and what exactly do you mean by invalid date? I'm thinking invalid as non-valid date format.

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
```diff
@@ -0,0 +1,58 @@
+Feature: Query String handling for GET requests for Resource Queries
+
+        Background:
+            Given a school exists
+                  | schoolId  | nameOfInstitution      | gradeLevels       | educationOrganizationCategories |
+                  | 255901001 | Grand Bend High School | [ "Tenth grade" ] | [ "School" ]                    |
+              And there is a valid AcademicWeek
+                  | weekIdentifier | beginDate  | endDate    | totalInstructionalDays |
+                  | Week One       | 2024-05-15 | 2024-05-15 | 2                      |
```
msilesgap, line 9: Can we edit begin/end dates here, I think is better if they are different to avoid possible issues during search

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
```diff
@@ -0,0 +1,58 @@
...
+        Scenario: Ensure clients can GET information when querying by valid date
+             When sending a GET request to "/ed-fi/academicWeeks?beginDate={beginDate}"
+             Then it should respond with 200
+              And response contains "weekIdentifier: "string" "
+
+        @ignore
+        Scenario: Ensure clients can't GET information when querying by invalid date
+             When sending a GET request to "/ed-fi/academicWeeks?beginDate=024-04-09"
+             Then it should respond with 400
+              And response body
```
msilesgap, line 21: This should be applied to the other scenarios as well, we need to be consistent in the used methods; instead of having 4 different validations in the response body, we could reuse the existing one
`              And the response body is`
Also use the entire json in the response body as well

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
stephenfuqua: @elopezgap this has _not_ been done. Still just has a phrase instead of the expected JSON result.

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
elopezgap: The correct way to execute this scenario is to use dates that are not part of the data

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
elopezgap: Done

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
elopezgap: Complete

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
elopezgap: Complete

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
```diff
@@ -0,0 +1,58 @@
...
+                  """
+                  {"detail":"Data validation failed. See 'validationErrors' for details.","type":"urn:ed-fi:api:bad-request:data","title":"Data Validation Failed","status":400,"correlationId":null,"validationErrors":{"$.beginDate":["The value '<beginDate>' is not valid for BeginDate."]}}
```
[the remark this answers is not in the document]
elopezgap: Complete

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
```diff
@@ -0,0 +1,75 @@
...
+             When sending a GET request to "/ed-fi/academicWeeks?beginDate=1970-04-09"
+             Then it should respond with 200
+              And response body is an empty array
+
+        @ignore
+        Scenario: Ensure clients can GET information when querying by begin date and end date
+            Given there is an invalid Begin date
+                  | weekIdentifier | beginDate  | endDate    | totalInstructionalDays |
+                  | string         | 2024-04-09 | 2024-04-09 | 0                      |
+             When sending a GET request to "/ed-fi/academicWeeks?beginDate=<beginDate>&endDate=<endDate>"
```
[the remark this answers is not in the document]
elopezgap: Complete

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
```diff
@@ -0,0 +1,57 @@
...
+              And response body is
+                  """
+                  {"detail":"Data validation failed. See 'validationErrors' for details.","type":"urn:ed-fi:api:bad-request:data","title":"Data Validation Failed","status":400,"correlationId":null,"validationErrors":{"$.beginDate":["The value '024-04-09' is not valid for BeginDate."]}}
+                  """
+
+        @ignore
+        Scenario: Ensure clients can GET information when querying by date out of specified dates
+             When sending a GET request to "/ed-fi/academicWeeks?beginDate=1970-04-09"
+             Then it should respond with 200
+              And response body is an empty array
```
msilesgap, line 33: I think you can use the same existing method
```
And response body is
"""
{[]}
"""
```

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
```diff
@@ -0,0 +1,74 @@
...
+        Scenario: Ensure clients can get information when filtering by limit and and a valid offset
+             When sending a GET request to "/ed-fi/schools?offset=3&limit=5"
+             Then it should respond with status 200
+              And the response array contains 2 items
+
+        @ignore
+        Scenario: Ensure clients can get information when filtering by limit and offset greater than the total
+             When sending a GET request to "/ed-fi/schools?offset=6&limit=5"
+             Then it should respond with status 200
+              And response array contains 0 items
```
msilesgap, line 38: this should match the previous existing method, on the scenario above you have
`And the response array contains 2 items`

`And response array contains 0 items`

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
```diff
@@ -0,0 +1,74 @@
...
+
+        @ignore
+        Scenario: Ensure clients can GET information when querying using an offset without providing any limit in the query string
+            Given these schools exist:
+                  | schoolId | nameOfInstitution                             | gradeLevels         | educationOrganizationCategories     |
+                  | 5        | School with max edorgId value                 | [ "Postsecondary" ] | [ "School" ]                        |
+                  | 6        | UT Austin College of Education Under Graduate | [ "Postsecondary" ] | [ "Educator Preparation Provider" ] |
+             When sending a GET request to "/ed-fi/schools?offset=1"
+             Then it should respond with 200
+              And response body contains 2 records
```
msilesgap, line 64: Same as previous comment, try to reuse the existing methods, we are validating the # of records in the response array

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
```diff
@@ -0,0 +1,74 @@
...
+
+        @ignore
+        Scenario: Ensure clients can GET information when filtering with limits and properties
+            Given these schools exist:
+                  | schoolId | nameOfInstitution                             | gradeLevels         | educationOrganizationCategories     |
+                  | 5        | School with max edorgId value                 | [ "Postsecondary" ] | [ "School" ]                        |
+                  | 6        | UT Austin College of Education Under Graduate | [ "Postsecondary" ] | [ "Educator Preparation Provider" ] |
+             When sending a GET request to "/ed-fi/schools?nameOfInstitution=UT+Austin+College+of+Education+Under+Graduate&limit=2"
+             Then it should respond with 200
+              And response contains 1 item
```
msilesgap, line 74: Same as previous comment, try to reuse the existing methods, we are validating the # of records in the response array

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
msilesgap: If that's the case I think we can add the values in the url, and you can add a comment mentioning the expected dates

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
elopezgap: Complete

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
elopezgap: Done

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
elopezgap: Done

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
```diff
@@ -0,0 +1,87 @@
...
+                  """
+
+        @ignore
+        Scenario: Ensure clients can get information when filtering by limit and offset greater than the total
+             When sending a GET request to "/ed-fi/schools?offset=6&limit=5"
+             Then it should respond with status 200
+              And response body is
```
msilesgap, line 42: This should be `And the response body is`

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
```diff
@@ -0,0 +1,87 @@
...
+                  """
+                  []
+                  """
+
+        @ignore
+        Scenario: Ensure clients can't GET information when filtering by limit and offset using invalid offset
+             When sending a GET request to "/ed-fi/schools?limit=6&offset=<Value>"
+             Then it should respond with 400
+              And response body is
```
msilesgap, line 51: This should be `And the response body is`

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
```diff
@@ -0,0 +1,87 @@
...
+
+        @ignore
+        Scenario: Ensure clients can GET information when querying using an offset without providing any limit in the query string
+            Given these schools exist:
+                  | schoolId | nameOfInstitution                             | gradeLevels         | educationOrganizationCategories     |
+                  | 5        | School with max edorgId value                 | [ "Postsecondary" ] | [ "School" ]                        |
+                  | 6        | UT Austin College of Education Under Graduate | [ "Postsecondary" ] | [ "Educator Preparation Provider" ] |
+             When sending a GET request to "/ed-fi/schools?offset=1"
+             Then it should respond with 200
+              And response body is
```
msilesgap, line 71: This should be `And the response body is`

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
```diff
@@ -0,0 +1,63 @@
...
+                  | 255901001 | Grand Bend High School | [ "Tenth grade" ] | [ "School" ]                    |
+              And there is a valid AcademicWeek
+                  | weekIdentifier | beginDate  | endDate    | totalInstructionalDays |
+                  | Week One       | 2024-05-15 | 2024-05-22 | 2                      |
+
+        @ignore
+        Scenario: Ensure clients can GET information when querying by valid date
+             When sending a GET request to "/ed-fi/academicWeeks?beginDate=2024-05-15"
+             Then it should respond with 200
+              And response body is
```
msilesgap, line 15: This should be `And the response body is`

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
```diff
@@ -0,0 +1,63 @@
...
+              """
+              {"id":"f0aad8e209b64134b4c056fa9f49eeff","schoolReference":{"schoolId":255901001,"link":{"rel":"School","href":"/ed-fi/schools/56f9effc3c7f44f496fd8d6783dd9360"}},"weekIdentifier":"Week One","beginDate":"2024-05-15","endDate":"2024-05-15","totalInstructionalDays":2,"_etag":"5250200021872954522","_lastModifiedDate":"2024-05-15T20:05:44.5566618Z"}
+              """
+
+        @ignore
+        Scenario: Ensure clients can't GET information when querying by invalid date
+             When sending a GET request to "/ed-fi/academicWeeks?beginDate=024-04-09"
+             Then it should respond with 400
+              And response body is
```
msilesgap, line 24: This should be `And the response body is`

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
```diff
@@ -0,0 +1,63 @@
...
+                  """
+                  {"detail":"Data validation failed. See 'validationErrors' for details.","type":"urn:ed-fi:api:bad-request:data","title":"Data Validation Failed","status":400,"correlationId":null,"validationErrors":{"$.beginDate":["The value '024-04-09' is not valid for BeginDate."]}}
+                  """
+
+        @ignore
+        Scenario: Ensure clients can GET information when querying by date out of specified dates
+             When sending a GET request to "/ed-fi/academicWeeks?beginDate=1970-04-09"
+             Then it should respond with 200
+              And response body is
```
msilesgap, line 33: This should be `And the response body is`

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
```diff
@@ -0,0 +1,63 @@
...
+              """
+              {[]}
+              """
+
+        @ignore
+        Scenario: Ensure clients can GET information when querying by string identifier
+             When sending a GET request to "/ed-fi/academicWeeks?weekIdentifier=Week+One"
+             Then it should respond with 200
+              And response body is
```
msilesgap, line 51: This should be `And the response body is`

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
```diff
@@ -0,0 +1,63 @@
...
+              """
+              {[]}
+              """
+
+        @ignore
+        Scenario: Ensure clients can GET information when querying by begin date and end date that are not part of the data
+             When sending a GET request to "/ed-fi/academicWeeks?beginDate=2024-05-20&endDate=2024-05-20"
+             Then it should respond with 200
+              And response body is
```
msilesgap, line 42: This should be `And the response body is`

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
```diff
@@ -0,0 +1,63 @@
...
+                  """
+                  {"id":"f0aad8e209b64134b4c056fa9f49eeff","schoolReference":{"schoolId":255901001,"link":{"rel":"School","href":"/ed-fi/schools/56f9effc3c7f44f496fd8d6783dd9360"}},"weekIdentifier":"Week One","beginDate":"2024-05-15","endDate":"2024-05-15","totalInstructionalDays":2,"_etag":"5250200021872954522","_lastModifiedDate":"2024-05-15T20:05:44.5566618Z"}
+                  """
+
+        @ignore
+        Scenario: Ensure clients can GET information when querying by integer identifier
+             When sending a GET request to "/ed-fi/academicWeeks?totalInstructionalDays=2"
+             Then it should respond with 200
+              And response body is
```
msilesgap, line 60: This should be `And the response body is`

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
```diff
@@ -0,0 +1,87 @@
...
+                  """
+                  {"id":"f24dc237b73741b4a778ce403906d8f9","schoolId":6,"nameOfInstitution":"UT Austin College of Education Under Graduate","addresses":[{"addressTypeDescriptor":"uri://ed-fi.org/AddressTypeDescriptor#Physical","city":"Austin","postalCode":"78712","stateAbbreviationDescriptor":"uri://ed-fi.org/StateAbbreviationDescriptor#TX","streetNumberName":"1912 Speedway Stop D5000","nameOfCounty":"Travis","periods":[]}],"educationOrganizationCategories":[{"educationOrganizationCategoryDescriptor":"uri://tpdm.ed-fi.org/EducationOrganizationCategoryDescriptor#Educator Preparation Provider"}],"identificationCodes":[],"indicators":[],"institutionTelephones":[],"internationalAddresses":[],"_ext":{"tpdm":{"postSecondaryInstitutionReference":{"postSecondaryInstitutionId":6000203,"link":{"rel":"PostSecondaryInstitution","href":"/ed-fi/postSecondaryInstitutions/02f5fec829834b7691db381ff5f61ff1"}}}},"schoolCategories":[],"gradeLevels":[{"gradeLevelDescriptor":"uri://ed-fi.org/GradeLevelDescriptor#Postsecondary"}],"_etag":"5250169253470311260","_lastModifiedDate":"2024-04-10T05:25:04.2923356Z"}
+                  """
```
stephenfuqua, line 74: I am going to update the response above:

1. Reformat for easier reading.
2. The "Given" sections don't setup any addresses, therefore the response shouldn't have any addresses. I will remove those and anything else that is not in the given/background.
3. Removing `link`, `_tag`, `_lastModifiedDate`, and `_ext`. These are features that we won't support in the Data Management Service right away (in the case of `link`, we'll never support it).

The last detail in particular I wouldn't have expected either of you to know, because this is the first time it has come up.

On the first point, I see that I have setup a tricky situation: I suggested tables instead of JSON for the Givens because it makes the data entry more compact. The table contains only the required fields. But, here's a problem: you can't just paste that exact body into the ODS/API and see what comes back. That is unfortunate, and we should think about whether we want to change that long-term. But not right now - I just mean thinking about it for a future ticket. Perhaps we can discuss in retrospective.

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
elopezgap: Done

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
elopezgap: Done

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
elopezgap: Done

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
```diff
@@ -0,0 +1,87 @@
...
+
+        @ignore
+        Scenario: Ensure clients can GET information when filtering with limits and properties
+            Given these schools exist:
+                  | schoolId | nameOfInstitution                             | gradeLevels         | educationOrganizationCategories     |
+                  | 5        | School with max edorgId value                 | [ "Postsecondary" ] | [ "School" ]                        |
+                  | 6        | UT Austin College of Education Under Graduate | [ "Postsecondary" ] | [ "Educator Preparation Provider" ] |
+             When sending a GET request to "/ed-fi/schools?nameOfInstitution=UT+Austin+College+of+Education+Under+Graduate&limit=2"
+             Then it should respond with 200
+              And response body is
```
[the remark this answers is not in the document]
elopezgap: Done

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
elopezgap: Done

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
elopezgap: Done

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
elopezgap: Done

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
elopezgap: Done

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ResourceQueriesQueryString.feature.disabled
elopezgap: Done

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/PagingQueryString.feature.disabled
```diff
@@ -0,0 +1,151 @@
...
+                  | -1                       |
+                  | 'zero'                   |
+                  | '5; select * from users' |
+                  | '0)'                     |
+                  | '1%27'                   |
+
+        @ignore
+        Scenario: Ensure clients can get information when filtering by limit and and a valid offset
+             When sending a GET request to "/ed-fi/schools?offset=3&limit=5"
+             Then it should respond with status 200
```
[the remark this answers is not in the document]
elopezgap: Done

PR #56, opened 2024-05-08
[DMS-129] Use JSON path in validation message

JBrenesSimpat:


src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ValidationErrors.feature
```diff
@@ -58,3 +58,84 @@ Scenario: Post a valid Descriptor
...
+    }
+    "beginDate": "2024-04-04",
+    "endDate": "2024-04-04",
+    "totalInstructionalDays": 300
+  }
+  """
+  Then the response code is 400
+    And the response body is
+"""
+{"detail":"Data validation failed. See 'validationErrors' for details.","type":"urn:ed-fi:api:bad-request:data","title":"Data Validation Failed","status":400,"correlationId":null,"validationErrors":{"$.":["'\"' is invalid after a value. Expected either ',', '}', or ']'. LineNumber: 5 | BytePositionInLine: 2."]},"errors":[]}
```
JBrenesSimpat, line 115: This error detail differs a bit from the ODS example, here I am simply sending the exception I get in logginMiddleware

stephenfuqua: There is a real problem with this error message though: it says that `\` is invalid after a value, but `\` is not in the payload. This makes me think that our code is not transmitting the payload correctly as shown.

Here's what the ODS/API says:

```json
{
  "detail": "Data validation failed. See 'validationErrors' for details.",
  "type": "urn:ed-fi:api:bad-request:data",
  "title": "Data Validation Failed",
  "status": 400,
  "correlationId": "ec5f955d-f8f8-431b-aaa3-a0e2a684136c",
  "validationErrors": {
    "$.schoolReference": [
      "After parsing a value an unexpected character was encountered: \". Path 'schoolReference', line 6, position 4."
    ]
  }
}
```

Its okay if our validation message is a little different in this case. It would be hard to have exactly the same, since we're using a different JSON parser.

src/services/EdFi.DataManagementService.Api/Core/Validation/DocumentValidator.cs
```diff
@@ -103,16 +103,53 @@ PruneResult PruneOverpostedData(JsonNode? documentBody, EvaluationResults evalua
...
             return validationErrors;
         }
     }
+
+    private static Dictionary<string, string[]> SplitErrorDetail(string error, string propertyName)
+    {
+        var validations = new Dictionary<string, string[]>();
+        if (error.Contains("[") && error.Contains("]"))
+        {
+            MatchCollection hits = Regex.Matches(error, "\"([^\"]*)\"");
```
stephenfuqua, line 135: Suggestion: move this regex to a static property and set the compiled flag. That will make this have higher performance. I think there is another example of that somewhere in the application - just look for other examples of `Regex` in the solution.

src/core/EdFi.DataManagementService.Core/Response/FailureResponse.cs
```diff
@@ -8,7 +8,7 @@ namespace EdFi.DataManagementService.Core.Response;
 /// <summary>
 /// Provides human-readable summary of the failure responses
 /// </summary>
-internal record FailureResponse(
+public record FailureResponse(
```
stephenfuqua, line 11: Does this need to be `public` instead of `internal`? I think Brad just changed this to `internal` in the past day or so.

src/frontend/EdFi.DataManagementService.Frontend.AspNetCore/Infrastructure/LoggingMiddleware.cs
```diff
@@ -41,20 +43,33 @@ public async Task Invoke(HttpContext context, ILogger<LoggingMiddleware> logger)
...
+            failureResponse = FailureResponse.ForDataValidation(
+                "Data validation failed. See 'validationErrors' for details.",
+                validationErrors,
+                new List<string>().ToArray()
+            );
+
+            var options = new JsonSerializerOptions
+            {
+                Encoder = JavaScriptEncoder.UnsafeRelaxedJsonEscaping
+            };
```
stephenfuqua, line 65: If we have another frontend, like a Lambda function, it will also require this same logic.  Perhaps it would make sense to move this logic to the `FailureResponse` class?

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ValidationErrors.feature
stephenfuqua, line 1: @msilesgap please review this feature file to see if you have any recommendations for Johnny's edits.

Note: the JSON responses here are not formatted with nice indentation, unlike with some other feature files. This should not be a problem, because in the C# step definition we will parse the JSON, and the parser won't care about spacing. It is safe to leave this long string online 30 instead of reformatting.

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ValidationErrors.feature
```diff
@@ -44,7 +44,7 @@ Scenario: Post an invalid body for academicWeeks missing schoolid for schoolRefe
 	Then the response code is 400
```
msilesgap, line 44: Nothing to change here, but we are planning to go through the different feature files and make sure we are using the same Step definitions. For instance, in other files we have ` Then it should respond with 400`, basically we should have only one method.

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ValidationErrors.feature
```diff
@@ -58,3 +58,84 @@ Scenario: Post a valid Descriptor
   """
   Then the response code is 201
 
+Scenario: Post an invalid body for academicWeeks missing more than one required field
+  When sending a POST request to "data/ed-fi/academicWeeks" with body
```
msilesgap, line 62: @stephenfuqua Could you verify a portion of the URL? Not sure if this could affect somehow, but I recall one of your earlier remarks suggesting to maintain it as `/ed-fi/academicWeeks`

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ValidationErrors.feature
stephenfuqua: My personal preference is to keep the nicely formatted JSON, with indentations, because it is easier to read. But it is not so strong a preference that I would take time to reformat the JSON in this file.

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/ValidationErrors.feature
stephenfuqua: Good point - we need to be consistent with this between files. My preference: do not include "data" since it is not a required path segment in an Ed-Fi API.

JBrenesSimpat: I tried in Postman without the word "data" in the URL and got a 404 Not Found response.

stephenfuqua: Yes, the URL must have "data" for both DMS and ODS/API. But this is not a requirement of _all_ Ed-Fi API applications. Better to stick with the core requirements here. In the C# step definition, we must combine this path segment with the URL base to construct the real URL. We can inject "data/" into the URL at that point.

src/frontend/EdFi.DataManagementService.Frontend.AspNetCore/Infrastructure/LoggingMiddleware.cs
bradbanister: Also, any idea why we are using `UnsafeRelaxedJsonEscaping`? It is in fact "unsafe". If we really do need it, please move this to a static field - no need to create every time.

src/core/EdFi.DataManagementService.Core/Validation/DocumentValidator.cs
```diff
@@ -107,27 +118,64 @@ PruneResult PruneOverpostedData(JsonNode? documentBody, EvaluationResults evalua
...
                     }
                 }
             }
             return validationErrors;
         }
     }
+
+    private static readonly Regex _propertyRegex = new Regex("\"([^\"]*)\"", RegexOptions.Compiled);
+
+    private static Dictionary<string, string[]> SplitErrorDetail(string error, string propertyName)
```
stephenfuqua, line 156: Is this business logic fully covered by unit tests?

Would it be helpful to make this public and have thorough unit testing directly on the function?

stephenfuqua:
If you haven't already done so, can you look through `*.feature.disabled` to see if there are any other failure responses that need to be adjusted?

src/core/EdFi.DataManagementService.Core/Middleware/ValidateEqualityConstraintMiddleware.cs
```diff
@@ -22,8 +23,13 @@ internal class ValidateEqualityConstraintMiddleware(ILogger _logger, IEqualityCo
     public async Task Execute(PipelineContext context, Func<Task> next)
     {
         _logger.LogDebug("Entering ValidateEqualityConstraintMiddleware- {TraceId}", context.FrontendRequest.TraceId);
+        JsonNode? body = new JsonObject();
+        if (context.FrontendRequest.Body != null)
+        {
+            body = JsonNode.Parse(context.FrontendRequest.Body);
```
bradbanister, line 29: Instead of doing this everywhere, create a new core middleware called ParseBodyMiddleware that does `JsonNode.Parse(context.FrontendRequest.Body)` on POST or PUT and stores it in a new `JsonNode?` field on `PipelineContext` called `ParsedBody`.  Then parse error handling will be in one place, and everywhere that uses `context.FrontendRequest.Body` can instead use `context.ParsedBody`.

See https://github.com/Ed-Fi-Exchange-OSS/Meadowlark/blob/main/Meadowlark-js/packages/meadowlark-core/src/middleware/ParseBodyMiddleware.ts for an example from Meadowlark.js

src/core/EdFi.DataManagementService.Core/Validation/DocumentValidator.cs
```diff
@@ -107,27 +118,64 @@ PruneResult PruneOverpostedData(JsonNode? documentBody, EvaluationResults evalua
...
+                            }
+                        }
                     }
                 }
             }
             return validationErrors;
         }
     }
+
+    private static readonly Regex _propertyRegex = new Regex("\"([^\"]*)\"", RegexOptions.Compiled);
```
bradbanister, line 154: Please add a comment that describes the purpose of this regex. Bonus points for a better variable name that describes the purpose.

bradbanister (changes requested):

src/core/EdFi.DataManagementService.Core/Middleware/ParseBodyMiddleware.cs
```diff
@@ -0,0 +1,41 @@
...
+                    if (body != null)
+                    {
+                        context.ParsedBody = body;
+                    }
+                }
+                catch (Exception ex)
+                {
+                    _logger.LogError(ex, "Unknown Error - {TraceId}", context.FrontendRequest.TraceId);
+
+                    FailureResponse.GenerateFrontendErrorResponse(context, ex.Message, 400);
```
stephenfuqua, line 34: The response from this function call is being thrown away. If that is correct, please change to set  `_ =  FailureResponse.GenerateFrontendErrorResponse(context, ex.Message, 400);`

Alternately, the code might be a little clean if it does not have a side effect of modifying `context`. In that case we can remove `context` as an argument and just use  `context.FrontendResponse = FailureResponse.GenerateFrontendErrorResponse(ex.Message, 400);`

One last thought: is there any possibility that this exception message will be something that we don't want the client application to receive?

src/core/EdFi.DataManagementService.Core/Middleware/ParseBodyMiddleware.cs
```diff
@@ -0,0 +1,41 @@
...
+                    var body = JsonNode.Parse(context.FrontendRequest.Body);
+
+                    if (body != null)
+                    {
+                        context.ParsedBody = body;
+                    }
+                }
+                catch (Exception ex)
+                {
+                    _logger.LogError(ex, "Unknown Error - {TraceId}", context.FrontendRequest.TraceId);
```
stephenfuqua, line 32: Its true that this is an unknown error, but we can do a bit better. Imagine reading the log and seeing "Unknown Error" followed by the details. You have to get the stack trace to figure out where this occurred. Perhaps write `"Unable to parse the request body as JSON - {TraceId}"`.

Actually, since this is a 400 response and a client error, it should not be logged _as an error_ on the server. This should just be a DEBUG message.

src/core/EdFi.DataManagementService.Core/Middleware/CoreLoggingMiddleware.cs
```diff
@@ -27,18 +27,7 @@ public async Task Execute(PipelineContext context, Func<Task> next)
...
-                Body: JsonSerializer.Serialize(
-                    new
-                    {
-                        message = "The server encountered an unexpected condition that prevented it from fulfilling the request.",
-                        traceId = context.FrontendRequest.TraceId
-                    }
-                ),
-                Headers: []
-            );
+            FailureResponse.GenerateFrontendErrorResponse(context, ex.Message, 400);
```
bradbanister, line 30: Undo this. The purpose of CoreLoggingMiddleware's exception handling is to throw 500's, like it says in the comments.

src/core/EdFi.DataManagementService.Core/Middleware/ExtractDocumentInfoMiddleware.cs
```diff
@@ -24,14 +25,22 @@ public async Task Execute(PipelineContext context, Func<Task> next)
             context.FrontendRequest.TraceId
         );
 
-        Debug.Assert(context.FrontendRequest.Body != null, "Body was null, pipeline config invalid");
-        JsonNode documentBody = context.FrontendRequest.Body;
+        if (context.ParsedBody == No.EmptyJsonNode && context.FrontendRequest.Body != null)
+        {
+            var body = JsonNode.Parse(context.FrontendRequest.Body);
```
bradbanister, line 30: You don't need to do any of this because your middleware already does it.

src/core/EdFi.DataManagementService.Core/Middleware/ValidateEqualityConstraintMiddleware.cs
```diff
@@ -23,7 +25,16 @@ public async Task Execute(PipelineContext context, Func<Task> next)
     {
         _logger.LogDebug("Entering ValidateEqualityConstraintMiddleware- {TraceId}", context.FrontendRequest.TraceId);
 
-        string[] errors = _equalityConstraintValidator.Validate(context.FrontendRequest.Body, context.ResourceSchema.EqualityConstraints);
+        if (context.ParsedBody == No.EmptyJsonNode && context.FrontendRequest.Body != null)
+        {
+            var body = JsonNode.Parse(context.FrontendRequest.Body);
```
bradbanister, line 30: You don't need to do any of this because your middleware already does it.

src/core/EdFi.DataManagementService.Core/Model/No.cs
```diff
@@ -40,6 +40,8 @@ internal static class No
     /// </summary>
     public static readonly DocumentUuid DocumentUuid = new("00000000-0000-0000-0000-000000000000");
 
+    public static readonly JsonNode EmptyJsonNode = new JsonObject();
```
bradbanister, line 43: Just call it `JsonNode`. The `Empty` is taken care of when you refer to it in code as `No.JsonNode`

bradbanister (changes requested):

src/core/EdFi.DataManagementService.Core/Middleware/ExtractDocumentInfoMiddleware.cs
JBrenesSimpat: I noticed that without this some test are failing

src/core/EdFi.DataManagementService.Core/ApiService.cs
```diff
@@ -40,6 +40,7 @@ ILogger<ApiService> _logger
                         new ApiSchemaValidationMiddleware(_apiSchemaProvider, _apiSchemaValidator, _logger),
                         new ProvideApiSchemaMiddleware(_apiSchemaProvider, _logger),
                         new ParsePathMiddleware(_logger),
+                        new ParseBodyMiddleware(_logger),
```
stephenfuqua, line 43: This comes before `ValidateDocumentMiddleware`. I just issued this request: `curl -X POST http://localhost:5198/data/ed-fi/students -d "{notjson]"`. I expected an error coming from `ParseBodyMiddleWare`, but note the stack trace below:
 
```none
2024-05-24 16:17:16.453 ERR Unknown Error - TraceId { Value = 0HN3S6RHI5T5L:00000001 }System.Text.Json.JsonReaderException: 'n' is an invalid start of a property name. Expected a '"'. LineNumber: 0 | BytePositionInLine: 1.
   at System.Text.Json.ThrowHelper.ThrowJsonReaderException(Utf8JsonReader& json, ExceptionResource resource, Byte nextByte, ReadOnlySpan`1 bytes)
   at System.Text.Json.Utf8JsonReader.ReadSingleSegment()
   at System.Text.Json.Utf8JsonReader.Read()
   at System.Text.Json.JsonDocument.Parse(ReadOnlySpan`1 utf8JsonSpan, JsonReaderOptions readerOptions, MetadataDb& database, StackRowStack& stack)
   at System.Text.Json.JsonDocument.ParseUnrented(ReadOnlyMemory`1 utf8Json, JsonReaderOptions readerOptions, JsonTokenType tokenType)
   at System.Text.Json.JsonDocument.ParseValue(ReadOnlyMemory`1 json, JsonDocumentOptions options)
   at System.Text.Json.Nodes.JsonNode.Parse(String json, Nullable`1 nodeOptions, JsonDocumentOptions documentOptions)
   at EdFi.DataManagementService.Core.Middleware.ParseBodyMiddleware.Execute(PipelineContext context, Func`1 next) in D:\tanager\Data-Management-Service\src\core\EdFi.DataManagementService.Core\Middleware\ParseBodyMiddleware.cs:line 23

2024-05-24 16:17:16.473 DBG Entering ValidateEndpointMiddleware - TraceId { Value = 0HN3S6RHI5T5L:00000001 }
2024-05-24 16:17:16.480 DBG Entering ValidateDocumentMiddleware- TraceId { Value = 0HN3S6RHI5T5L:00000001 }
```

So the error was detected in the new middle ware, but we keep processing and the client response ends up coming from the validation middleware instead of from the `ParseBodyMiddleware`.

src/core/EdFi.DataManagementService.Core/Middleware/ExtractDocumentInfoMiddleware.cs
bradbanister: Yes, it's a test setup problem. See my comment on the ExtractDocumentInfoMiddlewareTests

src/core/EdFi.DataManagementService.Core.Tests.Unit/Middleware/ExtractDocumentInfoMiddlewareTests.cs
```diff
@@ -62,13 +62,7 @@ public async Task Setup()
...
             context = new(
                 new(
-                    Body: JsonNode.Parse(
-                        """
-                        {
-                            "schoolId": "123"
-                        }
-"""
-                    )!,
+                    Body: """{"schoolId": "123"}"""!,
```
bradbanister, line 65: You need to put the parsed version of the string body in `context.ParsedBody` just like would happen after your new middleware runs.

src/core/EdFi.DataManagementService.Core/Middleware/ParseBodyMiddleware.cs
JBrenesSimpat: > Its true that this is an unknown error, but we can do a bit better. Imagine reading the log and seeing "Unknown Error" followed by the details. You have to get the stack trace to figure out where this occurred. Perhaps write `"Unable to parse the request body as JSON - {TraceId}"`.
> 
> Actually, since this is a 400 response and a client error, it should not be logged _as an error_ on the server. This should just be a DEBUG message.

Thanks @bradbanister , it was my first solution to the problem, however I thought that this change should be made in many places and that's why I added it in `ValidateEqualityConstraintMiddleware`

src/core/EdFi.DataManagementService.Core.Tests.Unit/Middleware/ExtractDocumentInfoMiddlewareTests.cs
JBrenesSimpat: > You need to put the parsed version of the string body in `context.ParsedBody` just like would happen after your new middleware runs.

Thanks @bradbanister , it was my first solution to the problem, however I thought that this change should be made in many places and that's why I added it in ValidateEqualityConstraintMiddleware and ´ExtractDocumentInfoMiddleware´

src/core/EdFi.DataManagementService.Core/Validation/DocumentValidator.cs
```diff
@@ -51,25 +52,38 @@ RequestMethod method
         EvaluationOptions validatorEvaluationOptions =
             new() { OutputFormat = OutputFormat.List, RequireFormatValidation = true };
 
+        // Parse the body just one time
+        JsonNode? parsedBody = JsonNode.Parse(frontendRequest.Body);
```
bradbanister, line 56: Shouldn't you be using `PipelineContext.ParsedBody` and modifying it instead? `frontendRequest.Body` is no longer relevant once `ParseBodyMiddleware` runs. Make sure the tests are updated look at `context.ParsedBody` as well.

bradbanister (changes requested):

PR #60, opened 2024-05-09
[DMS-145] Adding endpoint operations

msilesgap:
Adding endpoint operations for 

- POST
- GET
- UPDATE / PUT
- DELETE

### DIFF
--- a/src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/Resources/CreateResourcesValidation.feature
+++ b/src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/Resources/CreateResourcesValidation.feature
@@ -1,6 +1,6 @@
 # This is a rough draft feature for future use.
 @ignore
-Feature: Resources Create Operation validations
+Feature: Resources "Create" Operation validations
 
         Background:
             Given the Data Management Service must receive a token issued by "http://localhost"
@@ -16,6 +16,9 @@ Feature: Resources Create Operation validations
                   | shortDescription | Sick Leave                                     |
              Then it should respond with 201
               And the record can be retrieved with a GET request
+              # TODO: when writing the step definition,
+              # be sure to validate that the response matches the original payload,
+              # except for the addition of the 'id' field
 
         @ignore
         Scenario: Verify error handling with POST using invalid data
--- a/src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/Resources/CreateResourcesValidation.feature
+++ b/src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/Resources/CreateResourcesValidation.feature
@@ -1,0 +1,44 @@
+# This is a rough draft feature for future use.
+@ignore
+Feature: Validate Creation for Absence Event Category Descriptors
+
+        Background:
+            Given the Data Management Service must receive a token issued by "http://localhost"
+              And user is already authorized
+
+        @ignore
+        Scenario: Verify new resource can be created successfully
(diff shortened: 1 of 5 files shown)


src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/Resources/CreateResourcesValidation.feature
```diff
@@ -0,0 +1,44 @@
+# This is a rough draft feature for future use.
+@ignore
+Feature: Validate Creation for Absence Event Category Descriptors
```
stephenfuqua, line 3: "Absence Event Category" is the specific example, but the feature is more general than this. We are validating the behavior when a client application tries to issue a POST request.

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/Resources/CreateResourcesValidation.feature
```diff
@@ -0,0 +1,44 @@
...
+
+        Background:
+            Given the Data Management Service must receive a token issued by "http://localhost"
+              And user is already authorized
+
+        @ignore
+        Scenario: Verify new resource can be created successfully
+             When sending a POST request with valid body to
+                  | Name          | Endpoint                               |
+                  | academicWeeks | /ed-fi/absenceEventCategoryDescriptors |
```
stephenfuqua, line 13: The C# developer who picks up this story will have to decide about the payload. In this case perhaps you can put the exact payload in here.

"Name = academicWeeks" is actually irrelevant to this descriptor endpoint.

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/Resources/CreateResourcesValidation.feature
```diff
@@ -0,0 +1,44 @@
...
+             Then it should respond with 201
+              And the record can be retrieved with a GET request
```
stephenfuqua, line 15: In Esteban's current work, I discouraged him from putting the exact body here. But for your "CRUD" tests, it would be valuable to confirm that we get the expected response. Your line 15 seems good as it is. Assuming that the programmer can save the body from the "When" step, they can refer to it again in the "Then" step. So I think this is good. Suggestion: add a comment on the next line: `# TODO: when writing the step definition, be sure to validate that the response matches the original payload, except for the addition of the 'id' field`

src/services/EdFi.DataManagementService.Api.Tests.E2E/Features/Resources/CreateResourcesValidation.feature
```diff
@@ -0,0 +1,44 @@
...
+
+        @ignore
+        Scenario: Verify error handling with POST using invalid data
+             When sending a POST request with invalid body to
+                  | Name          | Endpoint                               |
+                  | academicWeeks | /ed-fi/absenceEventCategoryDescriptors |
+             Then it should respond with 400
```
stephenfuqua, line 22: Again, good to show the actual POST request body here, so that the reader can see exactly what is wrong with it. No need to state "invalid body" in the "When" clause, since the Scenario already says the body is invalid.

If you use the exact same structure that we have in other files, then the C# step definition can be reused instead of having to write a new step definition.

msilesgap: got it, reuse the same methods instead of creating valid / invalid.

PR #77, opened 2024-05-20
[DMS-172] Docker container for DMS database

CSR2017:


src/backend/.editorconfig
stephenfuqua, line 1: We the `.editorconfig` file so that the SonarLint analyzer would know how to handle these rules during the  `dotnet build` process. And, because the Dockerfile can only copy files in the current or child directories - not from parent directories - we needed a local copy.

.github/workflows/on-pullrequest-dockerfile.yml
```diff
@@ -31,8 +31,8 @@ jobs:
       matrix:
         dockerfile:
           [
-            { name: "dms_database", path: "src/backend/Dockerfile" },
-            { name: "dms", path: "src/Dockerfile" }
+            { name: "dms_database", path: "src/backend/Dockerfile", postgres_user: postgres, postgres_password: P@ssw0rd },
+            { name: "dms", path: "src/Dockerfile" , postgres_user: postgres, postgres_password: P@ssw0rd }
```
stephenfuqua, line 35: Why is this needed, @CSR2017 ? Is there any possibility that this is creating a default password in the postgresql image? We wouldn't want that. I have not used these arguments locally when I run `docker build`.

CSR2017: @stephenfuqua 
We had those values hardcoded on the docker file. So, thought of passing them as arguments to docker file. But I see your point. Will remove those.

simpat-adam (approved):

stephenfuqua (approved):

PR #120, opened 2024-06-04
[DMS-171] GET request for all resources returns a page of resources

JBrenesSimpat:


src/core/EdFi.DataManagementService.Core/ApiService.cs
```diff
@@ -71,6 +71,25 @@ ILogger<ApiService> _logger
                 )
         );
 
+    /// <summary>
+    /// The pipeline steps to satisfy a get by id request
```
stephenfuqua, line 75: Not getting "by id".

src/core/EdFi.DataManagementService.Core/ApiService.cs
```diff
@@ -126,10 +145,18 @@ public async Task<IFrontendResponse> Upsert(FrontendRequest frontendRequest)
...
-    public async Task<IFrontendResponse> GetById(FrontendRequest frontendRequest)
+    public async Task<IFrontendResponse> Get(FrontendRequest frontendRequest)
     {
         PipelineContext pipelineContext = new(frontendRequest, RequestMethod.GET);
-        await _getByIdSteps.Value.Run(pipelineContext);
+        string[] segments = frontendRequest.Path.Split(new char[] { '/' }, StringSplitOptions.RemoveEmptyEntries);
+        if (segments.Length == 3 && segments[2].Length == 36 && Guid.TryParse(segments[2], out _))
+        {
+            await _getByIdSteps.Value.Run(pipelineContext);
+        }
```
stephenfuqua, line 155: What will happen if someone has a request like `GET .../ed-fi/students/not-a-guid`? That should result in a 404, but it looks like it will end up triggering the Get All request below.

JBrenesSimpat: Is correct it goes by Get All path, however, in one of the pipeline steps (ParsePathMiddleware) it returns a 404

src/backend/EdFi.DataManagementService.Backend.Postgresql/Operation/GetDocumentByResourceName.cs
```diff
@@ -0,0 +1,70 @@
...
+public class GetDocumentByResourceName(NpgsqlDataSource _dataSource, ILogger<GetDocumentByResourceName> _logger) : IGetDocumentByResourceName
+{
+    public async Task<GetResult> GetByResourceName(IGetRequest getRequest, int offset, int limit)
+    {
+        _logger.LogDebug("Entering GetDocumentByResourceName.GetByResourceName - {TraceId}", getRequest.TraceId);
+        try
+        {
+            await using var connection = await _dataSource.OpenConnectionAsync();
+
+            await using NpgsqlCommand command =
```
bradbanister, line 28: Please move the usage of NpgsqlCommand into `SqlAction`. Update and upsert are both doing this, but we haven't moved get-by-id yet. The comment on `SqlAction` explains the division of responsibility with the action classes.

src/backend/EdFi.DataManagementService.Backend.Postgresql/Operation/GetDocumentByResourceName.cs
```diff
@@ -0,0 +1,70 @@
...
+namespace EdFi.DataManagementService.Backend.Postgresql.Operation;
+
+public interface IGetDocumentByResourceName
+{
+    public Task<GetResult> GetByResourceName(IGetRequest getRequest, int offset, int limit);
+}
+
+public class GetDocumentByResourceName(NpgsqlDataSource _dataSource, ILogger<GetDocumentByResourceName> _logger) : IGetDocumentByResourceName
+{
+    public async Task<GetResult> GetByResourceName(IGetRequest getRequest, int offset, int limit)
```
bradbanister, line 21: Your operation should take an `IQueryRequest` and return a `QueryResult`. Those objects handle pagination parameters and multiple document results respectively.

src/core/EdFi.DataManagementService.Core.External/Interface/IApiService.cs
```diff
@@ -24,7 +24,7 @@ public interface IApiService
     /// <summary>
     /// DMS entry point for all API GET by id requests
     /// </summary>
-    public Task<IFrontendResponse> GetById(FrontendRequest frontendRequest);
+    public Task<IFrontendResponse> Get(FrontendRequest frontendRequest);
```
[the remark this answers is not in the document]
bradbanister: Yes. Please call this new action `Query`. It is the simplest form of a query, one with no query parameters,

src/backend/EdFi.DataManagementService.Backend.Postgresql/PostgresqlDocumentStoreRepository.cs
```diff
@@ -53,6 +54,24 @@ public async Task<GetResult> GetDocumentById(IGetRequest getRequest)
         }
     }
 
+    public async Task<GetResult> GetDocumentByResourceName(IGetRequest getRequest, int offset, int limit)
```
bradbanister, line 57: As an FYI, everything needed to represent a request to a backend action belongs in that backend action's request object. In this case you are actually implementing a slice of `Query` so you already have `IQueryRequest` available to you with `offset` and `limit` in it.

src/core/EdFi.DataManagementService.Core.External/Interface/IDocumentStoreRepository.cs
```diff
@@ -22,6 +22,11 @@ public interface IDocumentStoreRepository
     /// </summary>
     public Task<GetResult> GetDocumentById(IGetRequest getRequest);
 
+    /// <summary>
+    /// Entry point for get document by resource name requests
+    /// </summary>
+    public Task<GetResult> GetDocumentByResourceName(IGetRequest getRequest, int offset, int limit);
```
bradbanister, line 28: You'll be surprised to hear that you don't need this at all! You are actually implementing the `IQueryHandler` interface that is in the same folder.

src/backend/EdFi.DataManagementService.Backend.Postgresql/PostgresqlDocumentStoreRepository.cs
bradbanister: Really I should point out that you are actually implementing `public async Task<QueryResult> QueryDocuments(IQueryRequest queryRequest)` down at the bottom of the file.

src/core/EdFi.DataManagementService.Core.External/Interface/IApiService.cs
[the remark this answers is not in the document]
bradbanister: This is right, we need to make `IApiService` expose a more generic `Get`. But `Get` should look at the path and decide to either call the existing `GetById` implementation or a brand new `Query` implementation.

src/core/EdFi.DataManagementService.Core/ApiService.cs
bradbanister: `GetById` should stay as-is but not be exposed through the `IApiService` interface. A _new_ `Get` that is exposed should check for a document uuid and either call `GetById` or a new `Query` method. `ParsePathMiddleware` already does this with a regex so reuse that rather than do it yourself with the `segments` stuff.

Meadowlark.js has an example of this new `Get` behavior here:
https://github.com/Ed-Fi-Exchange-OSS/Meadowlark/blob/21eca1ca8eeb0362354f94304821eb73cf9910e7/Meadowlark-js/packages/meadowlark-core/src/handler/FrontendFacade.ts#L174

src/core/EdFi.DataManagementService.Core/Handler/GetByResourceNameHandler.cs
```diff
@@ -0,0 +1,56 @@
...
+using EdFi.DataManagementService.Core.Backend;
+using EdFi.DataManagementService.Core.External.Backend;
+using EdFi.DataManagementService.Core.External.Interface;
+using EdFi.DataManagementService.Core.Model;
+using EdFi.DataManagementService.Core.Pipeline;
+using Microsoft.Extensions.Logging;
+using static EdFi.DataManagementService.Core.External.Backend.GetResult;
+
+namespace EdFi.DataManagementService.Core.Handler;
+internal class GetByResourceNameHandler(IDocumentStoreRepository _documentStoreRepository, ILogger _logger)
```
bradbanister, line 15: This of course should now be `QueryHandler`

src/core/EdFi.DataManagementService.Core/Handler/GetByResourceNameHandler.cs
```diff
@@ -0,0 +1,56 @@
...
+    : IPipelineStep
+{
+    public async Task Execute(PipelineContext context, Func<Task> next)
+    {
+        _logger.LogDebug("Entering GetByResourceNameHandler - {TraceId}", context.FrontendRequest.TraceId);
+
+        int offset = 0;
+        int limit = 25;
+
+        if (context.FrontendRequest.QueryParameters.ContainsKey("offset"))
```
bradbanister, line 25: This is all behavior that belongs in a middleware. Trust me, it's going to get bigger and more complicated fast.
I suggest a middleware called `ValidateQueryMiddleware`. For now all it needs to do is validate the special query parameters `limit` and `offset`, and put them into a `PaginationParameters` object on `PipelineContext` if they are valid. We already have the `PaginationParameters` object. 

Example from Meadowlark.js at:
https://github.com/Ed-Fi-Exchange-OSS/Meadowlark/blob/21eca1ca8eeb0362354f94304821eb73cf9910e7/Meadowlark-js/packages/meadowlark-core/src/middleware/ValidateQueryMiddleware.ts#L30

which calls: 

https://github.com/Ed-Fi-Exchange-OSS/Meadowlark/blob/21eca1ca8eeb0362354f94304821eb73cf9910e7/Meadowlark-js/packages/meadowlark-core/src/validation/PaginationValidator.ts#L16

src/backend/EdFi.DataManagementService.Backend.Postgresql.Test.Integration/GetTest.cs
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        [Test]
+        public void It_should_be_found_by_get_by_key()
+        {
+            _getByKeyResult!.Should().BeOfType<QueryResult.QuerySuccess>();
+            (_getByKeyResult! as QueryResult.QuerySuccess)!.EdfiDocs.Length.Should().Be(1);
+        }
+    }
+}
```
bradbanister, line 68: I'd suggest two new tests
1) Insert two or three documents, check that query returns all of them.

2) Insert two documents that are different resources. Query on one of the resources. Should only return the one correct document.

src/core/EdFi.DataManagementService.Core/Middleware/ValidateQueryMiddleware.cs
```diff
@@ -0,0 +1,35 @@
...
+    public async Task Execute(PipelineContext context, Func<Task> next)
+    {
+        _logger.LogDebug("Entering ValidateQueryMiddleware - {TraceId}", context.FrontendRequest.TraceId);
+
+        int offset = 0;
+        int limit = 25;
+
+        if (context.FrontendRequest.QueryParameters.ContainsKey("offset"))
+        {
+            offset = int.TryParse(context.FrontendRequest.QueryParameters["offset"], out int o) ? o : 0;
```
bradbanister, line 22: Instead of the constant `0`, couldn't it be `offset`? Same below.

bradbanister (changes requested):

bradbanister (approved):

PR #125, opened 2024-06-07
[DMS-212] - Creation of the Scenarios of InvalidReferenceValidation

elopezgap:
Creation of the file InvalidReferenceValidation.feature.disabled.

Keep working on the first scenario because of unexpected response

### DIFF
--- a/src/tests/EdFi.DataManagementService.Tests.E2E/Features/InvalidReferenceValidation.feature.disabled
+++ b/src/tests/EdFi.DataManagementService.Tests.E2E/Features/InvalidReferenceValidation.feature.disabled
@@ -120,6 +120,6 @@ Feature: Invalid Reference Validation
                       "type": "urn:ed-fi:api:conflict:invalid-reference",
                       "title": "Resource Not Unique Conflict due to invalid-reference",
                       "status": 409,
-                      "correlationId": "76e6b75d-8df7-44ed-9ed5-94b4b06cee01"
+                      "correlationId": null
                   }
                   """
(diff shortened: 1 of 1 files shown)


src/tests/EdFi.DataManagementService.Tests.E2E/Features/InvalidReferenceValidation.feature.disabled
```diff
@@ -0,0 +1,125 @@
...
+                  """
+             Then it should respond with 409
+              And the response body is
+                  """
+                  {
+                      "detail": "The referenced 'Section' resource does not exist.",
+                      "type": "urn:ed-fi:api:conflict:invalid-reference",
+                      "title": "Resource Not Unique Conflict due to invalid-reference",
+                      "status": 409,
+                      "correlationId": "76e6b75d-8df7-44ed-9ed5-94b4b06cee01"
```
msilesgap, line 123: This should be null

elopezgap: Changed

msilesgap (approved):

src/tests/EdFi.DataManagementService.Tests.E2E/Features/InvalidReferenceValidation.feature.disabled
```diff
@@ -0,0 +1,125 @@
...
+                      },
+                      "studentIndicators": [
+                          {
+                              "indicatorName": "Fake",
+                              "indicator": "Fake"
+                          }
+                      ]
+                  }
+                  """
+             Then it should respond with 403
```
stephenfuqua, line 24: @elopezgap @msilesgap this is not the correct response, so we need another pull request. Yes, this is the correct response from the ODS/API, but it is not the correct response for the Data Management Service.

Use the next scenario to find the expected response. This is what I was trying to explain on Thursday, Esteban, using Course as an example.

PR #140, opened 2024-06-12
[DMS-241] Improve Workflow Permissions

chrisgocode:


vimayya (approved):

.github/workflows/on-pullrequest-dockerfile.yml
```diff
@@ -29,7 +29,6 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       pull-requests: write
-      security-events: write
```
stephenfuqua, line 32: I think this is required. @msilesgap please review and confirm one way or another.

stephenfuqua: NVM, I see that this is no longer required, because we had to disable the sarif file upload (lines 83-88).

stephenfuqua (approved):

PR #153, opened 2024-06-18
[DMS-219] Dockerfile based on NuGet package

msilesgap:
Added new docker file based on nuget package
Publish image to Docker Hub


src/Nuget.Dockerfile
```diff
@@ -0,0 +1,32 @@
+# SPDX-License-Identifier: Apache-2.0
+# Licensed to the Ed-Fi Alliance under one or more agreements.
+# The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+# See the LICENSE and NOTICES files in the project root for more information.
+
+FROM mcr.microsoft.com/dotnet/aspnet:8.0.3-alpine3.19-amd64@sha256:a531d9d123928514405b9da9ff28a3aa81bd6f7d7d8cfb6207b66c007e7b3075 AS runtimebase
+
+RUN apk --no-cache add curl=~8 unzip=~6 dos2unix=~7 bash=~5 gettext=~0
```
stephenfuqua, line 8: what is the purpose of installing `curl`, `dos2unix`, `bash`, and `gettext`?

src/Nuget.Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+WORKDIR /app
+
+RUN umask 0077 && \
+    wget -O /app/EdFi.DataManagementService.zip "https://pkgs.dev.azure.com/ed-fi-alliance/Ed-Fi-Alliance-OSS/_apis/packaging/feeds/EdFi/nuget/packages/EdFi.DataManagementService/versions/${VERSION}/content" && \
+    unzip /app/EdFi.DataManagementService.zip -d /app/ && \
+    rm -f /app/EdFi.DataManagementService.zip && \
+    chmod 700 /app/* -- **
+
+EXPOSE ${ASPNETCORE_HTTP_PORTS}
+CMD ["dotnet", "/app/DataManagementService/EdFi.DataManagementService.Frontend.AspNetCore.dll"]
```
stephenfuqua, line 32: We should be able to use the same `run.sh` file here as used in the original ("local") `Dockerfile`. That also means you'll need to copy appsettings.template.json into the container, with proper chmod commands on the template and run.sh. Also will need to add the `NEED_DATABASE_SETUP` env variable above.

src/Nuget.Dockerfile
stephenfuqua: Since I won't be here to respond tomorrow, I'll go ahead and partially answer my question.

* I have seen `curl` in some other container, but we don't need it: we have `wget`.
* `dos2unix` is used to change the line types on files we copy in, but we don't need it in our case. The run script and appsettings template are already saved in Git with the correct line endings (`\lf` without `\cr`).
* Alpine doesn't come with `bash` and that makes it a little harder to debug. For localhost development we definitely prefer Bash. But it does have `sh`, which is good enough. So I recommend skipping `bash` for this "production level image".
* I don't even remember what `gettext` is, and I don't see us using it anywhere here.

src/Nuget.Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+
+LABEL maintainer="Ed-Fi Alliance, LLC and Contributors <techsupport@ed-fi.org>"
+
+RUN apk --no-cache add bash=~5 gettext=~0 postgresql16-client=~16
+
+FROM runtimebase AS setup
+
+ENV LOG_LEVEL=${LOG_LEVEL}
+#replace with latest
+ARG VERSION=0.0.0-alpha.0.113
```
msilesgap, line 16: @stephenfuqua I think the answer is yes, but I would like to confirm if the Version value should be latest

src/Nuget.Dockerfile
msilesgap: - Removed `curl, dos2unix`
- `gettext` is required for this line in the sh file `envsubst < /app/appsettings.template.json > /app/appsettings.json`
- `bash` was removed

src/Nuget.Dockerfile
stephenfuqua: `latest` is not a valid tag for a NuGet package.

stephenfuqua: Is it possible to have an `ARG` without a default value? I can't remember. If so, might as well remove the default value above - since it will no longer be correct as soon as the next pull request merges.

src/Nuget.Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+
+RUN apk --no-cache add gettext=~0 postgresql16-client=~16
+
+FROM runtimebase AS setup
+
+ENV LOG_LEVEL=${LOG_LEVEL}
+#replace with latest
+ARG VERSION=0.0.0-alpha.0.113
+ENV ASPNETCORE_HTTP_PORTS=8080
+ENV OAUTH_TOKEN_ENDPOINT=${OAUTH_TOKEN_ENDPOINT}
```
stephenfuqua, line 18: This variable must be set at _runtime_, but it is not required when _building_ an image. In other words, you don't need it here. That's actually true of all the other environment variables below. They're not being used in any way.

src/Nuget.Dockerfile
stephenfuqua: As discussed in daily standup, we can leave this in place.

src/Nuget.Dockerfile
stephenfuqua: As discussed in daily standup, we can leave these in place.

stephenfuqua (approved):

docs/DOCKER.md
```diff
@@ -34,6 +34,7 @@ POSTGRES_PASSWORD=<Non-admin user password>
 POSTGRES_PORT=<Port for postgres server Eg. 5432>
 POSTGRES_HOST=<DNS or IP address of the PostgreSQL Server, i.e. sql.somedns.org Eg. 172.25.32.1>
 LOG_LEVEL=<serilog log level i.e. Information>
+OAUTH_TOKEN_ENDPOINT=<Authentication service url>
```
stephenfuqua, line 37: Please replace the first paragraph with this:

```none
The [Dockerfile](../src/Dockerfile) in `src` builds the Ed-Fi Data Management
Service (DMS) image directly from source code, and [nuget.Dockerfile](../src/nuget.Dockerfile)
build the image from the pre-built NuGet packages. Both versions include both the API
application and the database installer. The source code version is primarily used in local and 
end-to-end testing; the NuGet package version is intended for production-like environments
and for publishing to [Docker Hub](https://hub.docker.com/u/edfialliance).
```

msilesgap: Added

stephenfuqua (approved):

PR #195, opened 2024-07-17
[DMS-238] Automatically run integration tests on pull request

JBrenesSimpat:


.github/workflows/on-pullrequest.yml
```diff
@@ -130,6 +130,44 @@ jobs:
...
 
+  run-integration-tests:
+    name: Run Integration Tests
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: pwsh
+    services:
+      postgres:
+        image: postgres:14.12
```
JBrenesSimpat, line 141: This is the version that is included with ubuntu https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md#postgresql

msilesgap:
On this file [docs/RUNNING-LOCALLY.md](https://github.com/Ed-Fi-Alliance-OSS/Data-Management-Service/blob/d6c34dfb178c442325a9ef0c23579ae756e6290b/docs/RUNNING-LOCALLY.md) there is a section called `Running the EdFi.DataManagementService.Api.Tests.Integration`, could you check if the documentation needs to be updated.

.github/workflows/on-pullrequest.yml
stephenfuqua: Since it is included, I don't think you needed to add this `services` section. This section starts a Docker image running PostgreSQL.

To start the service directly inside the runner, instead of in another container, I think you just need to run `sudo systemctl start postgresql.service`.

stephenfuqua: However, I think you have to configure it yourself in that case. [Stack Overflow example](https://stackoverflow.com/a/78269163/30384). What you have already done is working fine, so let's keep it. However, in that case, let's update the PostgreSQL version to the same one used in our Kubernetes setup: `postgres:16.3-alpine3.20@sha256:e89da2c083a5405943408b6807cd1fd25dc9010c1294e30611b841778bedc653`

stephenfuqua (approved):

PR #198, opened 2024-07-19
[DMS-279] New ODS/API 7.2 error message format for key unification

JBrenesSimpat:


src/tests/EdFi.DataManagementService.Tests.E2E/StepDefinitions/StepDefinitions.cs
```diff
@@ -302,11 +302,22 @@ public void ThenTheResponseBodyIs(string expectedBody)
...
-            );
+            (responseJson as JsonObject)?.Remove("correlationId");
+            (expectedBodyJson as JsonObject)?.Remove("correlationId");
+
+            try
+            {
+                responseJson.Should().BeEquivalentTo(expectedBodyJson, options => options
+                    .WithoutStrictOrdering()
+                    .IgnoringCyclicReferences()
+                    .RespectingRuntimeTypes()
```
JBrenesSimpat, line 313: Testing I found that it always gave the evaluation of the expected result as a false positive, I found that adding `.RespectingRuntimeTypes()` solves the problem.

src/core/EdFi.DataManagementService.Core/Validation/EqualityConstraintValidator.cs
```diff
@@ -33,15 +41,36 @@ public string[] Validate(JsonNode? documentBody, IEnumerable<EqualityConstraint>
...
-                errors.Add($"Constraint failure: document paths {equalityConstraint.SourceJsonPath.Value} and {equalityConstraint.TargetJsonPath.Value} must have the same values");
+                string conflictValues = string.Join(
+                    ", ",
+                    sourceValues
+                        .Concat(targetValues)
+                        .Distinct(new JsonNodeEqualityComparer())
+                        .Select(x => $"'{x}'")
+                        .ToArray()
+                );
+                var targetSegment = targetPath.Segments[1].ToString().Replace(".", "");
```
simpat-adam, line 66: I don't know that this will always be the proper way to get the property name you're looking for. It looks like you're assuming a path like `{$.schoolReference.schoolId}` and extracting `schoolId`. Instead of hard coding the segment at index 1, it might be safer to take the segment at the _last_ index. Also instead of `.Replace` I would suggest either `TrimStart` or `[1..]`

simpat-adam (changes requested):

simpat-adam (approved):

PR #204, opened 2024-07-24
[DMS-215] E2E tests for managing References record

JBrenesSimpat:


src/tests/EdFi.DataManagementService.Tests.E2E/Features/General/DiscoveryAPI.feature
```diff
@@ -18,7 +18,7 @@ Feature: The Discovery API provides information about the application version, s
                       "urls": {
                           "dependencies": "{BASE_URL}/metadata/dependencies",
                           "openApiMetadata": "{BASE_URL}/metadata/specifications",
-                          "oauth": "{BASE_URL}/oauth/token",
+                          "oauth": "http://localhost/oauth/token",
```
JBrenesSimpat, line 21: `http://localhost` is the API response

stephenfuqua: The backend code for the step definition is supposed to be replacing the `{BASE_URL}` token with the actual base URL for the running instance. This way we can try running the E2E tests against a different URL in the future, and this test shouldn't fail. Since the test passed, I thought i had written the step definition code correctly. See [this line](https://github.com/Ed-Fi-Alliance-OSS/Data-Management-Service/blob/fed2f612104c0a353b9c8286cdf36a0f00e2b207/src/tests/EdFi.DataManagementService.Tests.E2E/StepDefinitions/StepDefinitions.cs#L415).

src/tests/EdFi.DataManagementService.Tests.E2E/Features/References/DeleteReferenceValidation.feature
```diff
@@ -1,150 +1,120 @@
...
                   {
-                      "detail": "The requested action cannot be performed because this item is referenced by an existing 'School' item.",
+                      "detail": "The requested action cannot be performed because this item is referenced by existing School item(s).",
                       "type": "urn:ed-fi:api:data-conflict:dependent-item-exists",
                       "title": "Dependent Item Exists",
                       "status": 409,
-                      "correlationId": null
+                      "correlationId": null,
+                      "validationErrors": null,
+                      "errors": null
```
stephenfuqua, line 49: Since `validationErrors` and `errors` are arrays, it would be better to return `[]` instead of `null`. It is unusual to have a JSON document with a `null` value.

src/tests/EdFi.DataManagementService.Tests.E2E/Features/References/UpdateReferenceValidation.feature
```diff
@@ -175,14 +172,16 @@ Feature: Update Reference Validation
...
-                      "title": "Unresolved Reference",
-                      "status": 409,
-                      "correlationId": null
+                      "detail":"Identifying values for the StudentSectionAssociation resource cannot be changed. Delete and recreate the resource item instead.",
+                      "type":"urn:ed-fi:api:bad-request:data-validation-failed:key-change-not-supported",
+                      "title":"Key Change Not Supported",
+                      "status": 400,
+                      "correlationId": null,
+                      "validationErrors":null,
+	                    "errors":null
```
stephenfuqua, line 185: Interestingly, `StudentSectionAssociation` is also one of the few entities where we will need to allow cascading updates in the future. Today, we do get this new error message above - but the old one is what we should get in the future. So we'll need to `@ignore` this test until we work on cascading updates in DMS-80.

Please restore the old expected response, except that it had an error. The detail should be `The referenced 'Section' item does not exist.`.

`key-change-not-supported` is an important scenario, and it is covered in scenario 4.

src/tests/EdFi.DataManagementService.Tests.E2E/Management/JsonComparer.cs
stephenfuqua, line 1: nice work!

src/tests/EdFi.DataManagementService.Tests.E2E/Features/General/DiscoveryAPI.feature
simpat-adam: The "oauth" endpoint is the only one that's hard coded. See `ContainerSetup` class.

stephenfuqua: OK, good point. We still need to be able to inject this value from configuration. Will need a different token other than `{BASE_URL}`.

stephenfuqua (approved):

PR #225, opened 2024-08-14
[DMS-286] Adding Kafka connectors

msilesgap:


stephenfuqua:
As a team we have decided to abandon the k8s approach in favor of Docker Compose.

PR #251, opened 2024-08-29
[DMS-49] Add a Healthcheck for DMS API image

JBrenesSimpat:


src/frontend/EdFi.DataManagementService.Frontend.AspNetCore/Program.cs
```diff
@@ -34,6 +35,9 @@
 
 app.MapRouteEndpoints();
 
+app.MapHealthChecks("/health");
+app.UseHealthChecks("/health", new HealthCheckOptions { Predicate = _ => true });
```
simpat-adam, line 39: I don't think you need the `HealthCheckOptions` from what I read, if predicate is null it will run all the checks, which is what we want. This will do the same thing but it's confusing to developers so I would leave it out if you can.

src/frontend/EdFi.DataManagementService.Frontend.AspNetCore/Infrastructure/HealthCheckServiceExtensions.cs
```diff
@@ -0,0 +1,81 @@
...
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+using Npgsql;
+
+namespace EdFi.DataManagementService.Frontend.AspNetCore.Infrastructure;
+
+public class ApplicationHealthCheck : IHealthCheck
+{
+    private readonly ILogger<ApplicationHealthCheck> _logger;
+
+    public ApplicationHealthCheck(ILogger<ApplicationHealthCheck> logger)
```
simpat-adam, line 16: Prefer to use primary constructor syntax.

simpat-adam (changes requested):

src/frontend/EdFi.DataManagementService.Frontend.AspNetCore/Infrastructure/WebApplicationBuilderExtensions.cs
```diff
@@ -45,6 +48,20 @@ public static void AddServices(this WebApplicationBuilder webAppBuilder)
         ConfigureDatastore(webAppBuilder, logger);
         ConfigureQueryHandler(webAppBuilder, logger);
 
+        webAppBuilder.Services.AddSingleton(
```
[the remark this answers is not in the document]
JBrenesSimpat: > Why do you add this singleton for `DbHealthCheck` but I don't see one for `ApplicationHealthCheck`? Are you sure this is necessary?

It's because DbHealthCheck needs the ConnectionString and the provider

PR #256, opened 2024-09-04
[DMS-301] E2E testing for OpenSearch queries

CSR2017:


src/tests/EdFi.DataManagementService.Tests.E2E/Features/Descriptors/ReadDescriptorsValidation.feature
```diff
@@ -103,24 +67,6 @@ Feature: Read a Descriptor
                   []
                   """
```
CSR2017, line 68: Comment from @stephenfuqua : @msilesgap when we merge this PR, some of the scenarios will move. We will need to update the spreadsheet accordingly.

Why? Suganya and I discussed and agreed that this would simplify our ability to only reset OpenSearch between tests if really needed. Thus we want to keep all of the OpenSearch-related tests together in one place. We thought about using tags, but there is also a setup step that needs to happen ("background"), and we did not want that setup step scattered in many different places.

src/tests/EdFi.DataManagementService.Tests.E2E/Features/ResourceQueries/DescriptorQueriesQueryString.feature
```diff
@@ -0,0 +1,74 @@
+Feature: Query String handling for GET requests for Descriptor Queries
```
CSR2017, line 1: @msilesgap 
All the descriptors tests with query string are moved to this new file

eng/docker-compose/start-all-services.ps1
```diff
@@ -0,0 +1,44 @@
...
+    "kafka-opensearch.yml"
+)
+
+if ($d) {
+    if ($v) {
+        Write-Host "Shutting down with volume delete"
+        docker compose $files down -v
+    }
+    else {
+        Write-Host "Shutting down"
```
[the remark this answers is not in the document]
stephenfuqua: For a long time I was trying to obey this warning, but I have found that these other commands do not work the same way as `Write-Host` when called _from an inside a function_. These other functions write to output streams, not to the console, where we often literally want to write to the console.

This warning [has been controversial for the last six years at least](https://github.com/PowerShell/PSScriptAnalyzer/issues/1118).

I'm going to dismiss these alerts, and will consider trying to setup a global setting to ignore this rule.

stephenfuqua (approved):

stephenfuqua (approved):

PR #267, opened 2024-09-10
[DMS-314] Keycloak developer setup

JBrenesSimpat:
New keycloak docker compose file


eng/docker-compose/keycloak.yml
```diff
@@ -0,0 +1,17 @@
...
+    image: quay.io/keycloak/keycloak:latest
+    environment:
+      - KEYCLOAK_ADMIN=admin
+      - KEYCLOAK_ADMIN_PASSWORD=admin
+    command:
+      - start-dev
+    ports:
+      - "8081:8080"
+    volumes:
+      - ./keycloak_data:/opt/keycloak/data
```
stephenfuqua, line 17: Suggestions:

1. line 8, put a specific version & SHA256 hash code instead of `latest`.
2. lines 10 and 11, make it possible to override with environment variable. For example: `KEYCLOAK_ADMIN: ${KEYCLOAK_ADMIN:-admin}` (if the variable is not set, it defaults to the value "admin")
3. Also set a variable for the port mapping on line 15. For default, I would choose something even more different than 8080, so that there is less chance of overlap with some other user interface. For example, choosing 8023 is random and less likely to be used by something else.
4. Finally, for the volume on line 17, setup a Docker volume rather than a local directory. See the `volumes` section in the Postgresql yml file.

eng/docker-compose/KEYCLOAK-SETUP.md
```diff
@@ -0,0 +1,68 @@
...
+### 6.	In the upper left of the scream, you must create a new realm, called “edfi”.
+![alt text](./images/image-4.png)
+
+### 7.	In this configuration, we will only be asked for the name and click on Create
+![alt text](./images/image-5.png)
+
+## 8.	The home screen will change this way
+![alt text](./images/image-6.png)
+
+### 9.	We will proceed to the configuration of the client by clicking Clients, which is located in the left panel (Note: make sure you are in edit and not in Master).
```
simpat-adam, line 41: Typeo "make sure you are in edfi..."

eng/docker-compose/KEYCLOAK-SETUP.md
```diff
@@ -0,0 +1,68 @@
...
+![alt text](./images/image-7.png)
+
+## #10.	In General settings, we will assign a Client ID
+![alt text](./images/image-8.png)
+
+### 11.	En Capability config, we will enable
```
simpat-adam, line 47: Typeo "In Capability config"

eng/docker-compose/KEYCLOAK-SETUP.md
```diff
@@ -0,0 +1,68 @@
...
+### 7.	In this configuration, we will only be asked for the name and click on Create
+![alt text](./images/image-5.png)
+
+## 8.	The home screen will change this way
+![alt text](./images/image-6.png)
+
+### 9.	We will proceed to the configuration of the client by clicking Clients, which is located in the left panel (Note: make sure you are in edit and not in Master).
+![alt text](./images/image-7.png)
+
+## #10.	In General settings, we will assign a Client ID
```
simpat-adam, line 44: I think the client id should be "DmsConfigurationService"

eng/docker-compose/KEYCLOAK-SETUP.md
```diff
@@ -0,0 +1,68 @@
...
+![alt text](./images/image-8.png)
+
+### 11.	En Capability config, we will enable 
+a.	Client authentication
+b.	Authorization
+c.	Authentication Flow, check-in Standard flow
+![alt text](./images/image-9.png)
+
+### 12.	In Login settings, we will indicate the Root URL
```
simpat-adam, line 53: I think you'll want this root URL to be the root of the Configuration Service. Right now that is localhost:5126

simpat-adam (changes requested):

simpat-adam (approved):

PR #269, opened 2024-09-11
[DMS-252] Adding E2E Tests Post Conventions

msilesgap:
Adding E2E Scenarios POST - REST API Conventions

- Scenario: 01 Verify user can send a POST using a valid data
- Scenario: 02 Verify user is able to execute upsert
- Scenario: 03 Verify invalid json is handled correctly
- Scenario: 04 Verify missing fields are handled correctly
- Scenario: 05 Verify user can send a POST using extra fields
- Scenario: 06 Verify clients cannot post a resource being unauthenticated
- Scenario: 07 Verify clients cannot POST a resource without permissions
- Scenario: 08 Verify clients cannot POST a resource with invalid token
- Scenario: 09 Validate boundary values during POST action
- Scenario: 10 Validate special characters values during POST action


src/tests/EdFi.DataManagementService.Tests.E2E/Features/RESTAPIConventions/POSTRequest.feature
```diff
@@ -0,0 +1,258 @@
+Feature: Validating POST Actions
+
+        @APIConventions @POST
+        Scenario: 01 Verify user can send a POST using a valid data
```
msilesgap, line 4: Some of these tests could be similar to the existing ones in other feature files, but in order to keep the order of the sections we are covering it would be a good idea to keep them.

src/tests/EdFi.DataManagementService.Tests.E2E/Features/Security/TokenIntrospection.feature.disabled
```diff
@@ -16,42 +16,51 @@ Feature: JWT Token Introspection
             Given the Data Management Service must receive a token issued by "http://localhost"
 
     Rule: Discovery API does not require an authorization token
+        @ignore
```
msilesgap, line 19: The `@ignore ` here will help us filtering the existing tests that are marked as ignore, even if the file is disabled is hard to track tests that are not being executed.

src/tests/EdFi.DataManagementService.Tests.E2E/Features/RESTAPIConventions/POSTRequest.feature
stephenfuqua: This seems like a lot of duplication of the features in `CreateResourcesValidation.feature`. If there are any scenarios here that we do not already cover, then please add them to `CreateResourcesValidation.feature` instead of having a new file.

The goal of these tickets is to make sure all of the functionality described in the API standards guidance has scenarios. If it is convenient to organize the scenarios to more closely match the structure of the API standards, then let's think about re-organizing the existing files instead of creating duplicate tests.

Remember: every test that runs take a little bit of time and means more waiting to see if all tests passed.

src/tests/EdFi.DataManagementService.Tests.E2E/Features/Resources/CreateResourcesValidation.feature
```diff
@@ -720,3 +709,134 @@ Feature: Resources "Create" Operation validations
...
+                  """
+             Then it should respond with 201
+              And the response headers includes
+                  """
+                    {
+                        "location": "/ed-fi/students/{id}"
+                    }
+                  """
+
+        @POST
```
msilesgap, line 812: Moved to the end of the file as a workaround of the `'\"'` issue

src/tests/EdFi.DataManagementService.Tests.E2E/Features/RESTAPIConventions/POSTRequest.feature
msilesgap: Make sense, the new scenarios were added in the `CreateResourcesValidation.feature `, also the scenarios were tagged with the @POST value.

stephenfuqua (approved):

PR #278, opened 2024-09-19
[DMS-364] Validate query parameter value type

JBrenesSimpat:
I also activated some E2E tests targeting finished tickets.


src/core/EdFi.DataManagementService.Core/Middleware/ValidateQueryMiddleware.cs
```diff
@@ -146,6 +147,8 @@ public async Task Execute(PipelineContext context, Func<Task> next)
 
         List<QueryElement> queryElements = [];
 
+        var validationErrors = new Dictionary<string, string[]>();
```
bradbanister, line 150: Modern C# is shorter: `Dictionary<string, string[]> validationErrors = [];`

bradbanister (changes requested):
Please review my commit. I changed the ApiSchemaBuilder because the query field mapping additions weren't right. There were no positive tests in your unit tests. When I added some, your boolean type check fails. Please add more positive tests for boolean "false" string and date-time.

bradbanister (approved):

PR #282, opened 2024-09-24
[DMS-65] Enrich Saved Document with Trace ID

JBrenesSimpat:


bradbanister (changes requested):
Looks good, but please add simple integration tests that check the LastModifiedTraceId is the same going in on insert and update as coming out.

src/backend/EdFi.DataManagementService.Backend.Postgresql/Operation/GetDocumentById.cs
```diff
@@ -37,7 +39,7 @@ NpgsqlTransaction transaction
 
         try
         {
-            JsonNode? edfiDoc = await _sqlAction.FindDocumentEdfiDocByDocumentUuid(
+            var document = await _sqlAction.FindDocumentEdfiDocByDocumentUuid(
```
bradbanister, line 42: Please don't switch to `var` when there is already a type. It took me a bit to figure out the type actually changed.

src/backend/EdFi.DataManagementService.Backend.Postgresql/Operation/SqlAction.cs
```diff
@@ -58,7 +58,7 @@ TraceId traceId
                 {
                     await using NpgsqlCommand command =
                         new(
-                            $@"SELECT EdfiDoc FROM dms.Document WHERE DocumentPartitionKey = $1 AND DocumentUuid = $2 AND ResourceName = $3 {SqlFor(lockOption)};",
+                            $@"SELECT * FROM dms.Document WHERE DocumentPartitionKey = $1 AND DocumentUuid = $2 AND ResourceName = $3 {SqlFor(lockOption)};",
```
bradbanister, line 61: You never know what big metadata columns might be added to the `Document` table later. Let's just select `EdfiDoc` and the LastModifieds to minimize data transfer/conversion and future surprises. Either create a new simple `record` type for the return, or just return a (`JsonNode`, `DateTime`, `TraceId`) tuple.

src/backend/EdFi.DataManagementService.Backend.Postgresql/Operation/GetDocumentById.cs
```diff
@@ -47,13 +49,17 @@ NpgsqlTransaction transaction
...
             {
                 return new GetResult.GetFailureNotExists();
             }
 
-            // TODO: Documents table needs a last modified datetime
-            return new GetResult.GetSuccess(getRequest.DocumentUuid, edfiDoc, DateTime.Now);
+            return new GetResult.GetSuccess(
+                getRequest.DocumentUuid,
+                document.EdfiDoc.Deserialize<JsonNode>()!,
+                document.LastModifiedAt!.Value,
```
bradbanister, line 60: Don't use `!` here. You can't guarantee those values won't be null. Use a default value instead. Or, better yet, use default values in `FindDocumentEdfiDocByDocumentUuid().`

bradbanister (changes requested):

PR #283, opened 2024-09-25
[DMS-251] - Addition of numeration tag to each scenario

elopezgap:
Naming convention for this tag is DMS-000


src/tests/EdFi.DataManagementService.Tests.E2E/Features/Constraints/EqualityConstraintValidation.feature
```diff
@@ -6,6 +6,7 @@ Feature: Equality Constraint Validation
     "targetJsonPath": "$.schoolReference.schoolId"
     }
     ]
+        @DMS-001
```
stephenfuqua, line 9: Suggestion to change to `API-001` and so forth instead of `DMS-001`. I have the idea that someone else building an Ed-Fi API could also benefit from our spreadsheet of requirements. In that case, they wouldn't have a "DMS". Might also use the spreadsheet with the ODS/API at some point.

msilesgap (approved):

PR #291, opened 2024-10-02
[DMS-361] Mask request data in DEBUG logging

JBrenesSimpat:


src/core/EdFi.DataManagementService.Core/Middleware/CoreLoggingMiddleware.cs
```diff
@@ -17,12 +17,10 @@ internal class CoreLoggingMiddleware(ILogger _logger) : IPipelineStep
 {
     public async Task Execute(PipelineContext context, Func<Task> next)
     {
-        _logger.LogDebug("FrontendRequest: {FrontendRequest}", context.FrontendRequest);
```
JBrenesSimpat, line 20: I remove it from this place, move it to a middleware that controls it.

docs/DOCKER.md
```diff
@@ -54,6 +55,7 @@ POSTGRES_ADMIN_PASSWORD=P@ssw0rd53
 POSTGRES_PORT=5432
 POSTGRES_HOST=localhost
 LOG_LEVEL=Information
+MASK_REQUEST_BODY_IN_LOGS:true
```
stephenfuqua, line 58: `=` instead of `:`

src/core/EdFi.DataManagementService.Core/Middleware/CoreLoggingMiddleware.cs
stephenfuqua: After moving this, what is the purpose of `CoreLoggingMiddleware`?

src/core/EdFi.DataManagementService.Core/Middleware/RequestDataBodyLoggingMiddleware.cs
```diff
@@ -0,0 +1,100 @@
...
+        {
+            // Deserialize the JSON body in a dynamic object
+            JsonDocument jsonDoc = JsonDocument.Parse(body);
+            JsonElement maskedJson = MaskJsonElement(jsonDoc.RootElement);
+            return JsonSerializer.Serialize(maskedJson);
+        }
+        catch (JsonException ex)
+        {
+            logger.LogError(ex, "Error while masking request body.");
+            return body; // In case of error, it returns the original body.
```
stephenfuqua, line 75: Let's return an empty string so that there is no chance of accidentally logging PII.

src/core/EdFi.DataManagementService.Core/Middleware/RequestDataBodyLoggingMiddleware.cs
```diff
@@ -0,0 +1,100 @@
...
+                );
+            }
+        }
+        await next();
+    }
+
+    private static string SanitizeInput(string input)
+    {
+        // Deletes new line, line feed, carriage return and tab characters
+        return Regex.Replace(
```
stephenfuqua, line 56: I think this function is actually "minifying" the JSON, rather than "sanitizing" it.

Since you have a regex in there already, then it might be useful to create a single regex, and change to a [source-generated regular expression](https://learn.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-source-generators?pivots=dotnet-8-0).

You might want to try out [this minification technique](https://stackoverflow.com/a/8913186/30384).

src/core/EdFi.DataManagementService.Core/Middleware/CoreLoggingMiddleware.cs
JBrenesSimpat: > After moving this, what is the purpose of `CoreLoggingMiddleware`?

Respond with statusCode 500 - Unknown Error in case of any uncontrolled errors occurring in the following Core...Middleware, Now renamed to CoreExceptionLoggingMiddleware

stephenfuqua (approved):

PR #304, opened 2024-10-16
[DMS-394] Fix missing audit data on update

simpat-adam:


src/dms/backend/EdFi.DataManagementService.Backend.Postgresql/Operation/SqlAction.cs
```diff
@@ -253,7 +253,7 @@ TraceId traceId
     {
         await using var command = new NpgsqlCommand(
             @"UPDATE dms.Document
-              SET EdfiDoc = $1, LastModifiedAt = NOW(), LastModifiedTraceId = $4
+              SET EdfiDoc = $1, LastModifiedAt = clock_timestamp(), LastModifiedTraceId = $4
```
simpat-adam, line 256: using `clock_timestamp()` instead of `now()` because the latter would use the same timestamp for an entire transaction, making it difficult to test.

stephenfuqua (approved):

PR #308, opened 2024-10-17
[DMS-386] Simple authorization in the DMS

CSR2017:


eng/docker-compose/local-dms.yml
```diff
@@ -39,3 +39,9 @@ services:
       MINIMUM_THROUGHPUT: ${MINIMUM_THROUGHPUT:-2}
       BREAK_DURATION_SECONDS: ${BREAK_DURATION_SECONDS:-30}
       OPENSEARCH_URL: ${OPENSEARCH_URL:-http://dms-opensearch:9200}
+      ENFORCE_AUTHORIZATION: ${IDENTITY_ENFORCE_AUTHORIZATION:-false}
+      SERVICE_ROLE: ${IDENTITY_SERVICE_ROLE:-dms-client}
+      AUTHORITY: ${IDENTITY_AUTHORITY:-http://dms-keycloak:8080/realms/dms}
+      AUDIENCE: ${IDENTITY_AUDIENCE:-account}
+      REQUIRE_HTTPS_METADATA: ${IDENTITY_REQUIRE_HTTPS_METADATA:-false}
+      ROLE_CLAIM_TYPE: ${IDENTITY_ROLE_CLAIM_TYPE:-http://schemas\\.microsoft\\.com/ws/2008/06/identity/claims/role}
```
stephenfuqua, line 47: Would it be useful for these parameter names to be the same as the environment variable names?

eng/docker-compose/KEYCLOAK-SETUP.md
```diff
@@ -149,6 +149,68 @@ Keycloak locally using docker-compose.
...
+   you. Make sure to store this for later use.
+3. Follow the section titled "Configuring service-specific realm roles" to
+   create a role specific to your service (e.g., dms-client for a DMS-specific
+   role).
+4. Once you have created the role, assign it to your client by navigating to the
+   `Service Account Roles` tab under the client settings.
+
+   ![alt text](./images/image-17.png)
+
+   ![alt text](./images/image-18.png)
```
stephenfuqua, line 167: Suggestion: instead of "alt text", use "Screenshot of page XYZ in KeyCloak".

eng/docker-compose/start-local-dms.ps1
```diff
@@ -23,7 +23,11 @@ param (
 
     # Enable KafkaUI and OpenSearch Dashboard
     [Switch]
-    $EnableOpenSearchUI
+    $EnableOpenSearchUI,
+
+    # Enforce Authorization
+    [Switch]
+    $EnforceAuthorization
```
stephenfuqua, line 30: I like this. Can we set an environment variable that would override the .env file setting? [Environment variables precedence in Docker Compose](https://docs.docker.com/compose/how-tos/environment-variables/envvars-precedence/):

> The order of precedence (highest to lowest) is as follows:
> 1. Set using [docker compose run -e in the CLI](https://docs.docker.com/compose/how-tos/environment-variables/set-environment-variables/#set-environment-variables-with-docker-compose-run---env).
> 2. Set with either the environment or env_file attribute but with the value interpolated from your [shell](https://docs.docker.com/compose/how-tos/environment-variables/variable-interpolation/#substitute-from-the-shell) or an environment file. (either your default [.env file](https://docs.docker.com/compose/how-tos/environment-variables/variable-interpolation/#env-file), or with the [--env-file argument](https://docs.docker.com/compose/how-tos/environment-variables/variable-interpolation/#substitute-with---env-file) in the CLI).

I think this means we can add `-e IDENTITY_ENFORCE_AUTHORIZATION=true` to the `docker compose` command below when `$EnforceAuthorization -eq $true`

src/dms/frontend/EdFi.DataManagementService.Frontend.AspNetCore/Modules/CoreEndpointModule.cs
```diff
@@ -11,9 +11,10 @@ public class CoreEndpointModule : IEndpointModule
 {
     public void MapEndpoints(IEndpointRouteBuilder endpoints)
     {
-        endpoints.MapPost("/data/{**dmsPath}", Upsert);
-        endpoints.MapGet("/data/{**dmsPath}", Get);
-        endpoints.MapPut("/data/{**dmsPath}", UpdateById);
-        endpoints.MapDelete("/data/{**dmsPath}", DeleteById);
+        bool enforceAuthorization = endpoints.ServiceProvider.GetRequiredService<IConfiguration>().GetValue<bool>("IdentitySettings:EnforceAuthorization");
```
stephenfuqua, line 14: Shouldn't we have a shared function somewhere so that we don't need to repeat `.GetValue<bool>("IdentitySettings:EnforceAuthorization")` and on `WebApplicationBuilderExtensions` line 129?

src/dms/frontend/EdFi.DataManagementService.Frontend.AspNetCore.Tests.Unit/AuthenticationConstants.cs
```diff
@@ -0,0 +1,13 @@
...
+// The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+// See the LICENSE and NOTICES files in the project root for more information.
+
+namespace EdFi.DataManagementService.Frontend.AspNetCore.Tests.Unit;
+
+public static class AuthenticationConstants
+{
+    public const string AuthenticationSchema = "TestScheme";
+    public const string Role = "test-role";
+    public const string Client_Id = "test_client";
```
simpat-adam, line 12: Did you mean to use an underscore in this variable name?

PR #314, opened 2024-10-23
[DMS-360] Add Token Endpoint proxy 

roberthunterjr:
Add a `/oauth/token` endpoint to the DMS to allow for proxying to configured IdP Service.

The handler at `/oauth/token` adapts to the functionality of ODS/API, where a client can request basic authentication by base64 encoding a clientId:clientSecret in the Authorization headers of the request, and by including a `{grant_type: client_credentials}` in the body. The Authorization header format is validated by the proxy, then sent to an upstream server using an `OAuthService`, which makes a  `POST` request that includes a properly formatted `Authorization: Basic <clientId>:<clientSecret>`  header. 

A formatting check is not performed for the client request body- instead a body of `{"grant_type": "client_credentials"}` is included automatically to the OAuth service.


src/dms/backend/EdFi.DataManagementService.Backend.OAuthService/OAuthManager.cs
```diff
@@ -0,0 +1,57 @@
...
+using System.Text;
+
+namespace EdFi.DataManagementService.Backend.OAuthService;
+
+public class OAuthManager() : IOAuthManager
+{
+    public async Task GetAccessTokenAsync(HttpContext httpContext, string upstreamUri)
+    {
+        // Create HttpClient to proxy request.
+        var httpClientFactory = httpContext.RequestServices.GetRequiredService<IHttpClientFactory>();
```
stephenfuqua, line 16: Although we do have access to `httpContext`, in order to get the headers and request body, what do you think about also letting dependency injection handling the `IHttpClientFactory`? That could probably be done at the class constructor level, along with the logger.

requests.http
```diff
@@ -7,29 +8,32 @@ Content-Type: "application/json"
 }
 
 ###
+### DMS Auth running locally Proper
 POST http://localhost:5198/oauth/token
-Authorization: Basic clientId:clientSecret
-Content-Type: application/json
+Authorization: Basic CSClientApp:test1@Secret
```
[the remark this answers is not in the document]
roberthunterjr: Ahh, didn't mean to commit theses were for my local dev. Regarding your side note however, I agree that we should accept `application/json` as well as `application/x-www-form-urlencode` as acceptable types. I created DMS-408 to for enabling the `x-www-form-urlencode`, which will unlock the ability to properly forward the requests in the proxy.

src/dms/backend/EdFi.DataManagementService.Backend.OAuthService/OAuthManager.cs
```diff
@@ -0,0 +1,58 @@
...
+            httpContext.Response.StatusCode = StatusCodes.Status400BadRequest;
+            await httpContext.Response.WriteAsync($"Bad Request - Malformed Authorization Header '{authHeader}'");
+        }
+        else
+        {
+            // Decode Base64 clientId and clientSecret.
+            var base64Credentials = authHeader.ToString().Substring("Basic ".Length).Trim();
+            var credentials = Encoding.UTF8.GetString(Convert.FromBase64String(base64Credentials)).Split(':');
+            var clientId = credentials[0];
+            var clientSecret = credentials[1];
```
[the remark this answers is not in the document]
roberthunterjr: Yes, this was originally here to ensure clientId and clientSecret were being parsed properly and added to the header. I have removed that logic in favor of forwarding the headers directly. The content-type and request body are still hard-coded for now until the `x-www-form-urlencoded` is resolved.

src/dms/backend/EdFi.DataManagementService.Backend.OAuthService/EdFi.DataManagementService.Backend.OAuthService.csproj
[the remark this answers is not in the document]
roberthunterjr: TBH I have a strong preference for keeping it in the main application. In addition to the overhead and over-engineering, it seems like its non-trivial to interact and test the TokenEndpointHandler. Specifically, adding the IOAuthManager using dependency injection it good for the testing, but the actual implementation requires a concrete class.  Additionally, the logger and IHttpClientFactory do not seem to work with the dependency injection in OAuthManager class.

I wonder if this is due to being in a seperate project? Either way, I think consolidating makes sense because keeping them separate is not giving us what we want currently either.

src/dms/backend/EdFi.DataManagementService.Backend.OAuthService/OAuthManager.cs
```diff
@@ -0,0 +1,58 @@
...
+            upstreamRequest!.Content = new StringContent("grant_type=client_credentials", Encoding.UTF8, "application/x-www-form-urlencoded");
+
+            // In case of 5xx Error, pass 503 Service unavailable to client, otherwise forward status directly to client.
+            var response = new HttpResponseMessage();
+            try
+            {
+                response = await client.SendAsync(upstreamRequest);
+                httpContext.Response.StatusCode = (int)response.StatusCode;
+            }
+            catch (Exception)
```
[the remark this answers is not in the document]
roberthunterjr: I had trouble adding the logger directly in this class, (See above regarding DI issues). I did add a custom exception that includes a message and status code that can be caught and logged in the TokenEndpointModule.

src/dms/frontend/EdFi.DataManagementService.Frontend.AspNetCore/Modules/TokenEndpointModule.cs
```diff
@@ -14,11 +16,16 @@ public void MapEndpoints(IEndpointRouteBuilder endpoints)
...
     {
-        var tokenDetails = new TokenResponse("temporary-fake-token", 1800, "bearer");
-        httpContext.Response.StatusCode = StatusCodes.Status200OK;
-        await httpContext.Response.WriteAsSerializedJsonAsync(tokenDetails);
+        try
+        {
+            await oAuthManager.GetAccessTokenAsync(httpContext, appSettings.Value.AuthenticationService);
+        }
+        catch (Exception)
+        {
```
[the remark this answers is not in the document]
roberthunterjr: Done!

src/dms/backend/EdFi.DataManagementService.Backend.OAuthService/OAuthManager.cs
roberthunterjr: Agreed! Tried adding the class constructor but the compiler complains about some Json serialization issue when the injecting the `IHttpClientFactory`:

```
2024-10-24 13:42:38.543 ERR Unknown Error - 0HN7KBDAG4JA3:00000001System.InvalidOperationException: Each parameter in the deserialization constructor on type 'EdFi.DataManagementService.Backend.OAuthService.OAuthManager' must bind to an object property or field on deserialization. Each parameter name must match with a property or field on the object. Fields are only considered when 'JsonSerializerOptions.IncludeFields' is enabled. The match can be case-insensitive.
```

I tried to avoid the pit of fixing by just calling the `GetRequiredService`, but it seems the DI with this class library / project will ultimately need to get solved.

src/dms/backend/EdFi.DataManagementService.Backend.OAuthService/EdFi.DataManagementService.Backend.OAuthService.csproj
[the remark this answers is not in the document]
roberthunterjr: Although merging back is still a path forward, resolving the DI issues and refactoring makes it less of a priority in my opinion. Should we still think about changing this?

src/dms/backend/EdFi.DataManagementService.Backend.OAuthService/OAuthManager.cs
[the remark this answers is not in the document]
roberthunterjr: Logging added after resolving DI issues.

src/dms/backend/EdFi.DataManagementService.Backend.OAuthService/OAuthManager.cs
roberthunterjr: After refactoring, the IHttpClientFactory is now injected into the module and passed to the OAuthManager as a parameter. This not only removes the larger dependency on HttpContext, but also allows for the Module to handle some of the other parsing activites. This results in a method signature that only requires an instance of the client, an authHeader string, and the url of the service configured in AppSettings.json.

src/dms/backend/EdFi.DataManagementService.Backend.OAuthService/EdFi.DataManagementService.Backend.OAuthService.csproj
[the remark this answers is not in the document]
stephenfuqua: Once everything is working cleanly, if it is only 10 minutes of refactoring to merge into the core code, then probably worth doing so. We just don't need the separate project and DLL.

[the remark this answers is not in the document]
roberthunterjr: Sounds good. Everything else seems to be working now from a implementation / unit testing standpoint. The E2E tests in our pipeline are failing because the references to IOAuthManager and OAuthManager seem to be missing. This is something this refactor will likely solve so pinging the team for suggestions.

requests.http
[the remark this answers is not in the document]
roberthunterjr: Removed this local requests file from the PR.

src/dms/backend/EdFi.DataManagementService.Backend.OAuthService/EdFi.DataManagementService.Backend.OAuthService.csproj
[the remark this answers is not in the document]
roberthunterjr: Refactor completed. It was closer to 15 min so hope you dont mind 😄

src/dms/core/EdFi.DataManagementService.Core/OAuthManager.cs
```diff
@@ -0,0 +1,61 @@
...
+        // Generate request message to send upstream
+        _logger.LogInformation("Generating Request message to send upstream");
+        HttpRequestMessage upstreamRequest = new HttpRequestMessage(HttpMethod.Post, upstreamUri);
+
+        // Verify the request header contains Authorization Basic.
+        _logger.LogInformation("Verifying Authorization Header contains 'Basic' Authentication formatting");
+        if (!authHeaderString.Contains("Basic"))
+        {
+            _logger.LogError("Malformed Authorization Header");
+            throw new OAuthIdentityException($"Malformed Authorization Header", HttpStatusCode.BadRequest);
```
stephenfuqua, line 29: Since this was ultimately a client error, perhaps it would make sense to simply return a `HttpResponseMessage` instead of throwing an exception? And we can construct it following the `problem+json` approach for error messages.

Add ` TraceId traceId` to the function arguments. Then when we detect the bad content:

```csharp
            return new HttpResponseMessage(HttpStatusCode.BadRequest)
            {
                Content = new StringContent(
                    FailureResponse.ForBadRequest("Malformed Authorization header", traceId, [], []).ToString(),
                    Encoding.UTF8,
                    "application/problem+json"
                ),
            };
```

Do not need to log the error here, since this is a client error, not a system error.

src/dms/frontend/EdFi.DataManagementService.Frontend.AspNetCore/Modules/TokenEndpointModule.cs
```diff
@@ -14,11 +17,28 @@ public void MapEndpoints(IEndpointRouteBuilder endpoints)
...
     }
 
-    internal static async Task GenerateToken(HttpContext httpContext)
+    internal static async Task GenerateToken(HttpContext httpContext, IOptions<AppSettings> appSettings, IOAuthManager oAuthManager, ILogger<TokenEndpointModule> logger, IHttpClientFactory httpClientFactory)
     {
-        var tokenDetails = new TokenResponse("temporary-fake-token", 1800, "bearer");
-        httpContext.Response.StatusCode = StatusCodes.Status200OK;
-        await httpContext.Response.WriteAsSerializedJsonAsync(tokenDetails);
+        // Create client for sending upstream request.
+        logger.LogInformation("Handling OAuth Token request.");
```
stephenfuqua, line 23: The following will be useful. Will require changing `ExtractTraceIdFrom` from private to public.

```csharp
        var traceId = AspNetCoreFrontend.ExtractTraceIdFrom(httpContext.Request, appSettings);
        logger.LogInformation("Received token request for forwarding to identity provider - {TraceId}", traceId);
```

src/dms/frontend/EdFi.DataManagementService.Frontend.AspNetCore/Modules/TokenEndpointModule.cs
```diff
@@ -14,11 +17,28 @@ public void MapEndpoints(IEndpointRouteBuilder endpoints)
...
+
+        // Extract the Authorization Headers.
+        httpContext.Request.Headers.TryGetValue("Authorization", out var authHeader);
+        try
+        {
+            logger.LogInformation("Requesting access token from OAuth Manager.");
+            var response = await oAuthManager.GetAccessTokenAsync(client, authHeader.ToString(), appSettings.Value.AuthenticationService);
+
+            logger.LogInformation("OAuth Manager access token request successful.");
+            await response.Content.CopyToAsync(httpContext.Response.Body);
```
stephenfuqua, line 34: Try this:

```csharp
            var response = await oAuthManager.GetAccessTokenAsync(client, authHeader.ToString(), appSettings.Value.AuthenticationService, traceId);

            if (logger.IsDebugEnabled())
            {
                logger.LogDebug("Get token response code {Code} - {TraceId}", response.StatusCode, traceId);
            }
            httpContext.Response.StatusCode = (int)response.StatusCode;
            await response.Content.CopyToAsync(httpContext.Response.Body);
```

src/dms/frontend/EdFi.DataManagementService.Frontend.AspNetCore/Modules/TokenEndpointModule.cs
```diff
@@ -14,11 +17,28 @@ public void MapEndpoints(IEndpointRouteBuilder endpoints)
...
+        // Create client for sending upstream request.
+        logger.LogInformation("Handling OAuth Token request.");
+        var client = httpClientFactory.CreateClient();
+
+        // Extract the Authorization Headers.
+        httpContext.Request.Headers.TryGetValue("Authorization", out var authHeader);
+        try
+        {
+            logger.LogInformation("Requesting access token from OAuth Manager.");
+            var response = await oAuthManager.GetAccessTokenAsync(client, authHeader.ToString(), appSettings.Value.AuthenticationService);
```
stephenfuqua, line 31: In testing, I ended up in an infinite loop, because I had not yet changed my appsettings file to use a different URL.

src/dms/frontend/EdFi.DataManagementService.Frontend.AspNetCore/Modules/TokenEndpointModule.cs
```diff
@@ -14,11 +17,28 @@ public void MapEndpoints(IEndpointRouteBuilder endpoints)
...
+            await response.Content.CopyToAsync(httpContext.Response.Body);
+        }
+        catch (OAuthIdentityException ex)
+        {
+            logger.LogError(ex, "Error from OAuthManager service");
+            httpContext.Response.StatusCode = (int?)ex.StatusCode ?? (int)HttpStatusCode.BadGateway;
+            await httpContext.Response.WriteAsync($"Error Getting Access Token Async: {ex.Message}");
+        }
     }
 }
```
simpat-adam, line 43: It looks like we're no longer using `TokenRespose` at the bottom of this file, you can remove it.

src/dms/frontend/EdFi.DataManagementService.Frontend.AspNetCore/Modules/TokenEndpointModule.cs
```diff
@@ -14,11 +17,28 @@ public void MapEndpoints(IEndpointRouteBuilder endpoints)
         endpoints.MapPost("/oauth/token", GenerateToken);
     }
 
-    internal static async Task GenerateToken(HttpContext httpContext)
+    internal static async Task GenerateToken(HttpContext httpContext, IOptions<AppSettings> appSettings, IOAuthManager oAuthManager, ILogger<TokenEndpointModule> logger, IHttpClientFactory httpClientFactory)
```
simpat-adam, line 20: Go ahead and install `CSharpier` and set it to auto format on save. It's very handy. The length of this method signature without wrapping clued me in that you don't have it installed.

simpat-adam, line 20: This can be `private`

src/dms/core/EdFi.DataManagementService.Core/OAuthManager.cs
```diff
@@ -0,0 +1,61 @@
...
+
+namespace EdFi.DataManagementService.Core;
+
+public interface IOAuthManager
+{
+    public Task<HttpResponseMessage> GetAccessTokenAsync(HttpClient httpClient, string authHeaderString, string upstreamUri);
+}
+public class OAuthManager(ILogger<OAuthManager> logger) : IOAuthManager
+{
+    private readonly ILogger<OAuthManager> _logger = logger;
```
simpat-adam, line 17: Can remove this line

src/dms/core/EdFi.DataManagementService.Core/OAuthManager.cs
```diff
@@ -0,0 +1,61 @@
...
+
+public class OAuthIdentityException : Exception
+{
+    public OAuthIdentityException(string message) : base(message) { }
+
+    public OAuthIdentityException(string message, HttpStatusCode statusCode) : base(message)
+    {
+        StatusCode = statusCode;
+    }
+    public HttpStatusCode? StatusCode { get; set; }
```
simpat-adam, line 60: I don't think this needs to be nullable. If you make it not nullable you can simplify line 49 on `TokenEndpointModule`

src/dms/core/EdFi.DataManagementService.Core/OAuthManager.cs
```diff
@@ -0,0 +1,61 @@
...
+// Licensed to the Ed-Fi Alliance under one or more agreements.
+// The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+// See the LICENSE and NOTICES files in the project root for more information.
+using Microsoft.Extensions.Logging;
+using System.Net;
+using System.Text;
+
+namespace EdFi.DataManagementService.Core;
+
+public interface IOAuthManager
```
simpat-adam, line 11: Suggestion for organization: Move into a folder/namespace called `OAuth` and break these 3 into their own files. Probably put one of those `<summary>` comment blocks on the interface at least.

simpat-adam (changes requested):

src/dms/frontend/EdFi.DataManagementService.Frontend.AspNetCore/Modules/TokenEndpointModule.cs
roberthunterjr: Removed. Good catch

simpat-adam (approved):

PR #319, opened 2024-10-25
Fix errors and provide better resilience

stephenfuqua:


eng/docker-compose/setup-connectors.ps1
```diff
@@ -41,70 +40,73 @@ catch {
...
-         }
-     }
-     catch {
+# Source connector
+if (IsReady($sourceBase)) {
+    try {
+        $sourceResponse = Invoke-RestMethod -Uri $sourceUrl -Method Get -SkipHttpErrorCheck
+
+        # only true if the response was 200
+        if ($null -ne $sourceResponse.name) {
```
stephenfuqua, line 57: With more manual testing, I realized that the 404 response when the connector does not yet exist was throwing an error, rather than being captured in `$sourceResponse`. But it wasn't really _throwing an exception_... just logging an error. Which became problematic, as it was never caught and handled. So I found this way (`-SkipHttpErrorCheck`) to stop writing the error to `stderr` and instead let it be captured in a variable.

eng/docker-compose/start-local-dms.ps1
```diff
@@ -60,20 +58,25 @@ if ($d) {
     }
 }
 else {
-    $upArg = ""
-    if ($r) { $upArg = "--build" }
+    $upArgs = @(
+        "--detach"
+    )
+    if ($r) { $upArgs += @("--build") }
```
stephenfuqua, line 64: The change I made earlier fixed the situation I had locally, wanting to run the script with `-r`. But it broke the script when run _without_ `-r`. It turns out that `$upArg = ""` caused the `docker compose .. up $upArg` command to fail. Docker Compose was getting the empty string and responding with error. Which I still don't get. Weird. Changing to this array approach - which is what I had replaced earlier, though even that was not implemented quite right - fixed the core E2E test failure that began with my initial PR.

eng/docker-compose/start-local-dms.ps1
```diff
@@ -60,20 +58,25 @@ if ($d) {
...
+        $env:IDENTITY_ENFORCE_AUTHORIZATION = $false
     }
 
-    docker compose $files --env-file $EnvironmentFile up -d $upArg
+    docker compose $files --env-file $EnvironmentFile up $upArgs
 
-    Start-Sleep 30
+    if ($LASTEXITCODE -ne 0) {
+        throw "Unable to start local Docker environment, with exit code $LASTEXITCODE."
+    }
```
stephenfuqua, line 78: If the `docker compose` command fails, then we want to stop the process instead of proceeding to setup connectors.

stephenfuqua (commented):
[The e2e tests are passing now](https://github.com/Ed-Fi-Alliance-OSS/Data-Management-Service/actions/runs/11510493736) on this branch.

simpat-adam (approved):

PR #322, opened 2024-10-28
[DMS-254] - Creation of UpdateValidation.feature file 

elopezgap:
Creation of UpdateValidation.feature file
Fix of Numeration of ExperimentalQueries.feature.disabled


src/dms/tests/EdFi.DataManagementService.Tests.E2E/Features/Resources/UpdateValidation.feature
```diff
@@ -0,0 +1,80 @@
...
+              And the response body is
+                  """
+                  {
+                      "studentUniqueId": "111111",
+                      "birthDate": "2014-08-14",
+                      "firstName": "Russella",
+                      "lastSurname": "Mayers"
+                  }
+                  """
+              And the ETag is in the response header
```
stephenfuqua, line 32: Once we add support for eTags, the `_etag` property will show up in all of the GET responses. We need to decide how to test for that. Options:

1. Add `_etag` to _all_ GET response bodies in all of the feature files. That would make the feature very visible to all readers. 
2. Follow the current approach used for `_lastModifiedDate`: we have C# code that looks for the value, but removes it before validating the response. Thus, we do not need to include `_lastModifiedDate` in the feature files. This is working well, but does not _communicate_ very well to readers.
3. A hybrid approach: look for `_etag` in the C# step definition, like we do for `_lastModifiedDate`. And also add a step definition here `And the response body contains an etag`. We would not need to add this step definition on _all_ of the GET requests in all feature files. We could add it to only the scenarios on this file.

I prefer either (1) or (3).

src/dms/tests/EdFi.DataManagementService.Tests.E2E/Features/Resources/UpdateValidation.feature
```diff
@@ -0,0 +1,80 @@
...
+                  """
+                  {
+                      "studentUniqueId": "111111",
+                      "birthDate": "2014-08-14",
+                      "firstName": "Russella",
+                      "lastSurname": "Mayorga"
+                  }
+                  """
+              And the ETag is in the request header
+              And If-Match value matches ETag
```
stephenfuqua, line 46: We will need to rewrite this scenario. The `When` clause generates the HTTP request, so the `And` on the next line is too late - the request has already been made. So we will need to have a new approach written just for this purpose. I am thinking about it how that should work.

src/dms/tests/EdFi.DataManagementService.Tests.E2E/Features/Resources/etag.feature.disabled
```diff
@@ -0,0 +1,77 @@
...
+            Given the system has these "students"
+                  | studentUniqueId | birthDate  | firstName | lastSurname |
+                  | 111111          | 2014-08-14 | Russella  | Mayers      |
+            Given the system has these "schools"
+                  | schoolId  | nameOfInstitution        | gradeLevels                                                                      | educationOrganizationCategories                                                                                   |
+                  | 255901044 | Grand Bend Middle School | [ {"gradeLevelDescriptor": "uri://ed-fi.org/GradeLevelDescriptor#Sixth grade"} ] | [ {"educationOrganizationCategoryDescriptor": "uri://ed-fi.org/EducationOrganizationCategoryDescriptor#School"} ] |
+            Given the system has these "studentSchoolAssociations"
+                  | entryDate  | schoolReference | studentReference | entryGradeLevelDescriptor                                                               |
+                  | 2024-10-28 | 255901044       | 111111           | [ {"entryGradeLevelDescriptor": "uri://ed-fi.org/GradeLevelDescriptor#Seventh grade"} ] |
```
stephenfuqua, line 17: @elopezgap since your scenarios are only using Students, you do not need to setup descriptors, schools, or studentSchoolAssociations in the background.

stephenfuqua (approved):

PR #326, opened 2024-10-30
[DMS-342] Allow reset credentials in configuration manager

simpat-adam:


src/config/frontend/EdFi.DmsConfigurationService.Frontend.AspNetCore/Modules/ApplicationModule.cs
```diff
@@ -197,4 +214,77 @@ ILogger<ApplicationModule> logger
...
+                    }
+                }
+
+                break;
+            case ApplicationApiClientsResult.FailureUnknown failure:
+                logger.LogError("Error fetching ApiClients: {failure}", failure);
+                return Results.Problem(statusCode: 500);
+        }
+
+        switch (response.Count)
```
simpat-adam, line 279: Our data model supports many `ApiClient` per `Application` but the AdminApi spec doesn't.

stephenfuqua: I'm going to talk to Jason and Vinaya about this. I don't understand the use case of 1-to-many on those database tables.

stephenfuqua: For now I'm assuming that we do not need to distinguish `Application` and `ApiClient` from each other, like the `EdFi_Admin` database does.

simpat-adam: After discussion, leaving the many-to-one in the database but only using `First()` in `reset-credential`

CSR2017 (approved):

PR #333, opened 2024-11-05
[DMS-397] Case insensitive limit, offset, totalCount params.

roberthunterjr:
Adds support for mixed case query params.

E.g. `/ed-fi/schools?liMIt=2`  becomes acceptable with the applied changes.

The original solutionwas to evaluate the toLower cases of each of the params from within the Context.FrontendRequest object. This resulted in performing the correct validation for the values, but validation failed downstream with 400 even after parsing.

This solution performs a check for the reserve params `totalCount`, `limit`, and `offset` by intercepting the query params and writing the correct form prior to creating the FrontendRequest object. This allows the validation to still occur while leaving the original HTTP Request Context intact.

### DIFF
--- a/src/dms/frontend/EdFi.DataManagementService.Frontend.AspNetCore/AspNetCoreFrontend.cs
+++ b/src/dms/frontend/EdFi.DataManagementService.Frontend.AspNetCore/AspNetCoreFrontend.cs
@@ -9,7 +9,6 @@ using EdFi.DataManagementService.Core.External.Interface;
 using EdFi.DataManagementService.Core.External.Model;
 using EdFi.DataManagementService.Frontend.AspNetCore.Infrastructure.Extensions;
 using Microsoft.Extensions.Options;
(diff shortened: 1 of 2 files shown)


src/dms/frontend/EdFi.DataManagementService.Frontend.AspNetCore/AspNetCoreFrontend.cs
```diff
@@ -9,6 +9,7 @@
 using EdFi.DataManagementService.Core.External.Model;
 using EdFi.DataManagementService.Frontend.AspNetCore.Infrastructure.Extensions;
 using Microsoft.Extensions.Options;
+using Microsoft.Extensions.Primitives;
```
stephenfuqua, line 12: Unnecessary `using` statement, please delete.

stephenfuqua (approved):

PR #353, opened 2024-11-20
[DMS-357] Manual test with Elasticsearch instead of OpenSearch

CSR2017:


eng/docker-compose/start-local-dms.ps1
```diff
@@ -27,18 +27,29 @@ param (
...
 )
 
+if($SearchEngine -eq "ElasticSearch")
+{
+    $files += @("-f", "kafka-elasticsearch.yml")
+}
+else {
+    $files += @("-f", "kafka-opensearch.yml")
+}
+
```
stephenfuqua, line 52: When I start with OpenSearch / Kafka UI enabled, I get this error: 

```
service "kafka-ui" depends on undefined service "kafka-opensearch-sink": invalid compose project
```

eng/docker-compose/kafka-elasticsearch.yml
```diff
@@ -0,0 +1,126 @@
...
+      - kafka-postgresql-source
+      - kafka-elasticsearch-sink
+    environment:
+      DYNAMIC_CONFIG_ENABLED: 'true'
+      KAFKA_CLUSTERS_0_NAME: kafka1
+      KAFKA_CLUSTERS_0_BOOTSTRAPSERVERS: kafka:9092
+
+  elasticsearch:
+    image: docker.elastic.co/elasticsearch/elasticsearch:8.5.1
+    container_name: dms-elasticsearch
```
stephenfuqua, line 89: Recommendation: change to `container_name: dms-search` and same for `hostname`. Make the same change for OpenSearch. That way the connection string does not need to change for the DMS.

eng/docker-compose/.env.elasticsearch.example
```diff
@@ -0,0 +1,65 @@
...
+
+POSTGRES_PASSWORD=P@ssW0rd
+POSTGRES_DB_NAME=edfi_datamanagementservice
+
+CONNECT_SOURCE_PORT=8083
+CONNECT_SINK_PORT=8084
+
+ELASTICSEARCH_HTTP_PORT=9200
+ELASTICSEARCH_ANALYZER_PORT=9300
+ELASTICSEARCH_DASHBOARD_PORT=5601
```
stephenfuqua, line 13: Recommendation: put the new settings in to the existing example file instead of creating a new one. Add a comment explaining that you only need either the elastic search or the open search settings.

JBrenesSimpat (approved):

PR #363, opened 2024-12-05
[DMS-450] Docker Compose for the Config Service, and authorization demonstration

stephenfuqua:


eng/docker-compose/.env.example
```diff
@@ -54,15 +75,33 @@ BREAK_DURATION_SECONDS=30
...
+DMS_CONFIG_IDENTITY_AUDIENCE=${IDENTITY_AUDIENCE}
+DMS_CONFIG_IDENTITY_IDENTITY_SERVER=http://dms-keycloak:8080
+DMS_CONFIG_IDENTITY_REALM=edfi
+DMS_CONFIG_IDENTITY_CLIENT_ID=DmsConfigurationService
+DMS_CONFIG_IDENTITY_CLIENT_SECRET=s3creT@09
+DMS_CONFIG_IDENTITY_REQUIRE_HTTPS=false
+DMS_CONFIG_IDENTITY_ROLE_CLAIM_TYPE=${IDENTITY_ROLE_CLAIM_TYPE}
+DMS_CONFIG_IDENTITY_SCOPE=scp:edfi_dms_configuration_service/full_access
+DMS_CONFIG_LOG_LEVEL=Information
+DMS_CONFIG_DEPLOY_DATABASE=true
```
stephenfuqua, line 107: @bradbanister this is a copy of the `.env` file that I used for my testing today.

eng/docker-compose/kafka-opensearch.yml
```diff
@@ -117,3 +127,7 @@ volumes:
   kafka-opensearch-sink-logs:
   kafka-opensearch-sink-config:
   opensearch-data1:
+
+networks:
+  dms:
+    external: true
```
stephenfuqua, line 133: I used this external network so that keycloak can be started on its own before (e.g. while manually running the applications from VS / VS Code), and the service will still be accessible when starting DMS and Config Service using their compose files.

bradbanister (approved):

PR #379, opened 2024-12-20
[DMS-454] Add E2E OpenSearch and ElasticSearch to scheduled job

simpat-adam:


JBrenesSimpat (approved):

build-dms.ps1
```diff
@@ -244,6 +248,18 @@ function RunE2E {
 function E2ETests {
     Invoke-Step { DockerBuild }
 
+    # Clean up all the containers and volumes
+    Invoke-Execute {
+        try {
+            Push-Location eng/docker-compose/
+            ./start-local-dms.ps1 -d -v
+            ./start-local-dms.ps1 -SearchEngine "ElasticSearch" -d -v
```
stephenfuqua, line 256: @simpat-adam does this really require both lines? I would have thought line 256 was sufficient.

Is this running for if the search engine is "OpenSearch"?

simpat-adam: Yes, I have both because we run both open and elastic variants back to back and I didn't want to be locked into which order we run them in.

stephenfuqua (approved):

PR #381, opened 2025-01-08
[DMS-92] Basic eTag creation

simpat-adam:


src/dms/core/EdFi.DataManagementService.Core/Middleware/InjectVersionMetadataToEdFiDocumentMiddleware.cs
```diff
@@ -20,6 +21,10 @@ public async Task Execute(PipelineContext context, Func<Task> next)
             DateTimeOffset utcNow = DateTimeOffset.UtcNow;
             string formattedUtcDateTime = utcNow.ToString("yyyy-MM-ddTHH:mm:ssZ");
             context.ParsedBody["_lastModifiedDate"] = formattedUtcDateTime;
+            context.ParsedBody["_etag"] = DateTime
```
simpat-adam, line 24: I mimic'd the way [ODS API generates etag. ](https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS/blob/427af1af0e747f89966e0667e6b889c2ed5eb6ad/Application/EdFi.Ods.Api/Providers/ETagProvider.cs#L38)

src/dms/core/EdFi.DataManagementService.Core/Middleware/InjectVersionMetadataToEdFiDocumentMiddleware.cs
```diff
@@ -3,12 +3,13 @@
...
 // See the LICENSE and NOTICES files in the project root for more information.
 
+using System.Globalization;
 using EdFi.DataManagementService.Core.Pipeline;
 using Microsoft.Extensions.Logging;
 
 namespace EdFi.DataManagementService.Core.Middleware
 {
-    internal class InjectLastModifiedDateToEdFiDocumentMiddleware(ILogger _logger) : IPipelineStep
+    internal class InjectVersionMetadataToEdFiDocumentMiddleware(ILogger _logger) : IPipelineStep
```
bradbanister, line 12: Good name change 😎

src/dms/core/EdFi.DataManagementService.Core.External/Model/Etag.cs
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: Apache-2.0
+// Licensed to the Ed-Fi Alliance under one or more agreements.
+// The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+// See the LICENSE and NOTICES files in the project root for more information.
+
+namespace EdFi.DataManagementService.Core.External.Model;
+
+// A string type branded as an Etag which uniquely identifies a specific version of a document
+public record struct Etag(string Value);
```
bradbanister, line 9: Nice

src/dms/core/EdFi.DataManagementService.Core/Backend/UpsertRequest.cs
```diff
@@ -42,4 +42,6 @@ internal record UpsertRequest(
     IUpdateCascadeHandler UpdateCascadeHandler
 )
     : UpdateRequest(ResourceInfo, DocumentInfo, EdfiDoc, TraceId, DocumentUuid, UpdateCascadeHandler),
-        IUpsertRequest;
+        IUpsertRequest
+{
+}
```
bradbanister, line 47: Do we need this change?

src/dms/core/EdFi.DataManagementService.Core.External/Backend/UpsertResult.cs
```diff
@@ -16,7 +16,8 @@ public record UpsertResult
     /// A successful upsert request that took the form of an insert
     /// </summary>
     /// <param name="NewDocumentUuid">The DocumentUuid of the new document</param>
-    public record InsertSuccess(DocumentUuid NewDocumentUuid) : UpsertResult();
+    /// <param name="Etag">The Etag of the new document</param>
+    public record InsertSuccess(DocumentUuid NewDocumentUuid, Etag Etag) : UpsertResult();
```
bradbanister, line 20: Does a backend really need to return it? Doesn't core set it?

bradbanister (changes requested):
I do wonder if it would be better to generate etags from a hash of the document. Since DMS could be running on multiple servers, there's a teeny-tiny chance the clocks are out of sync in just the right way for different documents to have the same timestamp. Just something to think about.

src/dms/core/EdFi.DataManagementService.Core/Backend/UpsertRequest.cs
simpat-adam: oops

src/dms/core/EdFi.DataManagementService.Core.External/Backend/UpsertResult.cs
simpat-adam: Good point.

simpat-adam:
> I do wonder if it would be better to generate etags from a hash of the document. Since DMS could be running on multiple servers, there's a teeny-tiny chance the clocks are out of sync in just the right way for different documents to have the same timestamp. Just something to think about.

Definitely could be done. Couple thoughts

- Is it important to be able to re-generate the hash once it's generated? Currently only do this for testing. I wonder about things like the order of nodes when it's serialized to/from json strings, also would have to remove the actual `_etag` jsonNode from the doc if we were going to try to re-gen the hash.
- Speed of hashing is a factor because of cascading updates. Not sure if there's a significant difference between hashing the entire document vs the single property.

bradbanister:
> 
> Definitely could be done. Couple thoughts
> 
> * Is it important to be able to re-generate the hash once it's generated? Currently only do this for testing. I wonder about things like the order of nodes when it's serialized to/from json strings, also would have to remove the actual `_etag` jsonNode from the doc if we were going to try to re-gen the hash.
> * Speed of hashing is a factor because of cascading updates. Not sure if there's a significant difference between hashing the entire document vs the single property.

Good points. Let's just keep it in the back of our minds.

bradbanister (approved):

PR #397, opened 2025-01-28
[DMS-485] Resource Claim Action Validation

CSR2017:


src/dms/core/EdFi.DataManagementService.Core/Middleware/ResourceAuthorizationMiddleware.cs
```diff
@@ -91,12 +90,46 @@ public async Task Execute(PipelineContext context, Func<Task> next)
...
+                    "ResourceAuthorizationMiddleware: No actions on the resource claim {ResourceClaim} - {TraceId}",
+                    resourceClaim.Name,
+                    context.FrontendRequest.TraceId.Value
+                );
+                RespondAuthorizationError();
+                return;
+            }
+            var actionName = ActionResolver.Translate(context.Method).ToString();
+            var isActionAuthorized =
+                resourceActions.Find(x =>
```
simpat-adam, line 110: You could simplify this:
```
 var isActionAuthorized =
     resourceActions.SingleOrDefault(x =>
         string.Equals(x.Name, actionName, StringComparison.InvariantCultureIgnoreCase)
         && x.Enabled
     ) != null;
```

Recommend using `SingleOrDefault` instead of `Find` because `Find` will just look for the first match but `Single` will throw an exception if there are more than one match (bad claimset data) which I think we want. Also `string.equals` will simplify the null checking a little bit.

src/dms/core/EdFi.DataManagementService.Core/Middleware/ResourceAuthorizationMiddleware.cs
```diff
@@ -105,19 +138,29 @@ public async Task Execute(PipelineContext context, Func<Task> next)
...
+                    Headers: [],
+                    ContentType: "application/problem+json"
+                );
+            }
         }
-        catch (ConfigurationServiceException ex)
+        catch (Exception ex)
         {
-            _logger.LogError(ex, "Error while retrieving claim sets");
+            _logger.LogError(ex, "Error while authorizing the request");
```
simpat-adam, line 154: Should we put the trace id in this log statement?

simpat-adam (changes requested):

simpat-adam (approved):

PR #405, opened 2025-02-04
[DMS-417] Namespace Authorization for Descriptors and Resources - PUT POST

CSR2017:


simpat-adam:
Some of the E2E tests in `NamespaceAuthorization.feature` can be un-ignored following this change. Also could consider adding the error message validation to those tests. They were omitted when the tests were written.

PR #413, opened 2025-02-14
[DMS-514] Update Kafka to v3.9.0 and use KRaft instead of Zookeeper

axelmarquezh:


eng/docker-compose/kafka-elasticsearch.yml
```diff
@@ -1,32 +1,24 @@
...
+      # Default values for the following environment variables:
+      KAFKA_NODE_ID: 1
+      KAFKA_PROCESS_ROLES: broker,controller
+      KAFKA_CONTROLLER_LISTENER_NAMES: CONTROLLER
+      KAFKA_LISTENER_SECURITY_PROTOCOL_MAP: CONTROLLER:PLAINTEXT,PLAINTEXT:PLAINTEXT
+      KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR: 1
+      KAFKA_TRANSACTION_STATE_LOG_REPLICATION_FACTOR: 1
+      KAFKA_TRANSACTION_STATE_LOG_MIN_ISR: 1
+      KAFKA_GROUP_INITIAL_REBALANCE_DELAY_MS: 0
+      KAFKA_NUM_PARTITIONS: 1
```
axelmarquezh, line 21: If any configuration is overridden, none of the default configurations will be used; so we need to specify the rest of the required configurations with their values.
More info [here](https://hub.docker.com/r/apache/kafka).

PR #421, opened 2025-02-24
[DMS-541] DMS errors on optional array types

JBrenesSimpat:
The reason for this change is because the bulkLoader is sending to DMS arrays with an empty object, e.g.` “gradeLevels”: [{}]`


bradbanister:
Don't merge this yet

bradbanister (approved):

PR #429, opened 2025-02-28
Fix docker image - dll version conflict

CSR2017:


CSR2017:
Did not solve the issue, package update completed with https://github.com/Ed-Fi-Alliance-OSS/Data-Management-Service/pull/430

PR #433, opened 2025-03-04
[DMS-475] EducationOrganizationHierarchy Opensearch query

simpat-adam:


eng/docker-compose/postgresql_connector.json
simpat-adam, line 1: After these changes, the OpenSearch `_id` field is still a `String` where it should be `Long`, but this does fix it where the id is the correct value. ie `"_id": "60010",`

Also, the `hierarchy` field after the expandDocumentJson looks like this (notice "array"). I tried to fix it but it was taking too long. 
```
"hierarchy": {
            "array": [
              400,
              60010
            ]
          }
```

CSR2017 (approved):

PR #457, opened 2025-03-20
[DMS-563] Commit hook for CSharpier using Husky.Net

JBrenesSimpat:
To test it, you must first run ./setup-dev-environment.ps1 .
Suggestion: It may be a best idea to make a branch of this PR.
Then make any changes that modify the format of an existing class (do not use the Format Document or Format with CSharpier option).
Commit your change. You will see the file autoformats instantly and without the need for a new commit.


.config/dotnet-tools.json
JBrenesSimpat, line 1: This file is used to define other the tasks that Husky.Net will run.

.husky/task-runner.json
JBrenesSimpat, line 1: This file is used to define other the tasks that Husky.Net will run.

.husky/pre-commit
```diff
@@ -0,0 +1,25 @@
...
+# If there are modified files, run CSharpier, otherwise skip formatting.
+if [ -n "$CHANGED_FILES" ]; then
+    echo "Running CSharpier on changed files..."
+    dotnet csharpier $CHANGED_FILES || {
+        echo "CSharpier detected unformatted files. Please run 'dotnet csharpier .' to fix."
+        exit 1
+    }
+else
+    echo "No C# files changed, skipping formatting."
+fi
```
stephenfuqua, line 25: I was surprised that this bash script worked while running in Powershell. But it did work perfectly 😄 . I am watching this because we probably should apply this tool in projects on the Tools team as well.

simpat-jesus (approved):

.husky/pre-commit
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+. "$(dirname "$0")/_/husky.sh"
+
+echo 'Formatting Code'
+
+dotnet husky run --name dotnet-format-staged-files
```
simpat-jesus, line 7: 👍  Using husky tasks alows to test the logic without the need to make a commit

stephenfuqua (approved):

PR #471, opened 2025-04-03
[DMS-578] StudentSchoolAssociation Authorization backend

simpat-adam:
Create StudentSchoolAssociationAuthorization table,
Handle Create/Update (DELETE handled via cascade)

### DIFF
--- a/src/dms/core/EdFi.DataManagementService.Core.External/Model/AuthorizationPathway.cs
+++ b/src/dms/core/EdFi.DataManagementService.Core.External/Model/AuthorizationPathway.cs
@@ -13,9 +13,9 @@ public record AuthorizationPathway
     /// <summary>
     /// Authorization Pathway that uses StudentSchoolAssociation to calculate the EducationOrganizations that can reach the related Securable Document(s).
     /// </summary>
-    /// <param name="StudentId">Extracted from the request body. Has its default value if the request method is GET or DELETE.</param>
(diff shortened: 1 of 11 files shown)


src/dms/core/EdFi.DataManagementService.Core.External/Model/AuthorizationPathway.cs
```diff
@@ -13,9 +13,9 @@ public record AuthorizationPathway
     /// <summary>
     /// Authorization Pathway that uses StudentSchoolAssociation to calculate the EducationOrganizations that can reach the related Securable Document(s).
     /// </summary>
-    /// <param name="StudentId">Extracted from the request body. Has its default value if the request method is GET or DELETE.</param>
-    /// <param name="SchoolId">Extracted from the request body. Has its default value if the request method is GET or DELETE.</param>
-    public record StudentSchoolAssociation(StudentId? StudentId, EducationOrganizationId? SchoolId)
+    /// <param name="StudentId">Extracted from the request body.</param>
+    /// <param name="SchoolId">Extracted from the request body.</param>
+    public record StudentSchoolAssociation(StudentId StudentId, EducationOrganizationId SchoolId)
```
simpat-adam, line 18: Correct me if I'm wrong, but we don't need this construct on GET/DELETE. DELETEs are cascaded and GETs would use 'authorizationSecurable' construct.

PR #472, opened 2025-04-03
[DMS-612] Docker Compose Optimization

stephenfuqua:
I had to run the PR action manually, since there were no C# changes to trigger it.

* [Config Service PR workflow](https://github.com/Ed-Fi-Alliance-OSS/Data-Management-Service/actions/runs/14316990890)
* [DMS PR workflow](https://github.com/Ed-Fi-Alliance-OSS/Data-Management-Service/actions/runs/14317168591)

### DIFF
--- a/eng/docker-compose/start-local-config.ps1
+++ b/eng/docker-compose/start-local-config.ps1
@@ -55,7 +55,7 @@ else {
 
     Write-Output "Starting locally-built DMS config service"
 
-    docker compose $files --env-file $EnvironmentFile up $upArgs
+    docker compose $files --env-file $EnvironmentFile -p dms-local up $upArgs
 
     if ($LASTEXITCODE -ne 0) {
         throw "Unable to start local Docker environment, with exit code $LASTEXITCODE."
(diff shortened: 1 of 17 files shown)


eng/docker-compose/start-local-config.ps1
```diff
@@ -55,7 +55,7 @@ else {
 
     Write-Output "Starting locally-built DMS config service"
 
-    docker compose $files --env-file $EnvironmentFile up $upArgs
+    docker compose $files --env-file $EnvironmentFile -p dms-local up $upArgs
```
stephenfuqua, line 58: `-p` is for project name, and it is what the group of containers is called when viewing in Docker Desktop. Without this, the group is simply called "docker-compose" (the directory name), which is confusing if there are other repositories with a `docker-compose` directory.

stephenfuqua:
I had to run the PR action manually, since there were no C# changes to trigger it. ✅[all good](https://github.com/Ed-Fi-Alliance-OSS/Data-Management-Service/actions/runs/14314891761).

eng/docker-compose/start-local-config.ps1
vimayya: This one is the deploying local config service and its dependencies, can we name it cms-local?

stephenfuqua: done.

And I realized that i needed to test start and stop both for the DMS and CS powershell scripts. Running those made me realize that I had a problem that needed to be solved. Done.

vimayya (approved):

PR #485, opened 2025-04-12
[DMS-598] Add workarounds to make bulk load client succeed

axelmarquezh:


src/dms/core/EdFi.DataManagementService.Core/Middleware/ResourceActionAuthorizationMiddleware.cs
```diff
@@ -6,12 +6,14 @@
 using System.Diagnostics;
 using System.Net;
 using System.Text.Json.Nodes;
+using EdFi.DataManagementService.Core.Configuration;
```
simpat-adam, line 9: These new usings are not needed.

simpat-adam (approved):

PR #545, opened 2025-05-15
[DMS-672] Staff Authorization: E2E Test Scenarios

semalaiappan:
4 defects created 

https://edfi.atlassian.net/browse/DMS-696
https://edfi.atlassian.net/browse/DMS-697
https://edfi.atlassian.net/browse/DMS-703
https://edfi.atlassian.net/browse/DMS-704

### DIFF
--- a/src/dms/tests/EdFi.DataManagementService.Tests.E2E/Features/Authorization/RelationshipsWithEdOrgsAndStaffs.feature
+++ b/src/dms/tests/EdFi.DataManagementService.Tests.E2E/Features/Authorization/RelationshipsWithEdOrgsAndStaffs.feature
@@ -1,887 +1,264 @@
 Feature: RelationshipsWithEdOrgsAndStaff Authorization
 
-
         Background:
-            Given the claimSet "EdFiSandbox" is authorized with educationOrganizationIds "255901001"
+            Given the claimSet "EdFiSandbox" is authorized with educationOrganizationIds "255901,255901001"
               And the system has these descriptors
-                  | descriptorValue                                                          |
-                  | uri://ed-fi.org/PostSecondaryEventCategoryDescriptor#College Application |
-                  | uri://ed-fi.org/SexDescriptor#Female                                     |
-                  | uri://ed-fi.org/GradeLevelDescriptor#Tenth Grade                         |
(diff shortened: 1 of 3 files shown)


simpat-adam (changes requested):
I looked at the first few scenarios but not all of them. Please take this back to in progress and use the recently merged `RelationshipsWithEdOrgsAndContacts.feature` as your guide. We are interested in testing staff based authorization with this file. Remember there are two ways staff can be associated with an ed-org `StaffEducationOrganizationEmploymentAssociation` and `StaffEducationOrganizationAssignmentOrganization`. Please test POST/PUT/GET/DELETE/SEARCH for these and verify both authorized (success) and unauthorized outcomes. When verifying an unauthorized outcome, please include the response message in the assertion. 

Also as general feedback, please make use of the background setup for things like descriptors, schools etc to keep the scenarios tight.

src/dms/tests/EdFi.DataManagementService.Tests.E2E/Features/Authorization/RelationshipsWithEdOrgsAndStaffs.feature
```diff
@@ -0,0 +1,723 @@
+Feature: RelationshipsWithEdOrgsAndStaff Authorization
+
+        Background:
+            Given the claimSet "EdFiSandbox" is authorized with educationOrganizationIds "255901,255901001"
+              And the system has these descriptors
+                  | descriptorValue                                                                |
+                  | uri://ed-fi.org/GradeLevelDescriptor#Tenth Grade                               |
```
simpat-adam, line 7: One more note, we never discussed this during knowledge transfer but we have a mechanism in place where you don't need to explicitly create descriptors if they are referenced in setup steps. When we process the tables such as `And the system has these "localEducationAgencies"` we automatically create any descriptors that we encounter while processing the table. Since the E2E runs are getting so long, every wasted millisecond adds up especially since these background steps are ran over and over. 

You still must create any descriptor that is used in the actual testing portion of the test (if it is not already created by setup). I think in this file only `ProgramAssignmentDescriptor` would need to be manually created. 

I will bring this up during stand up so the rest of the team is aware.

simpat-adam (approved):

PR #549, opened 2025-05-19
[DMS-682] Added dorny/test-reporter

semalaiappan:


simpat-adam (commented):
How does this report failure if zero tests are run? Is that by default?

semalaiappan:
> How does this report failure if zero tests are run? Is that by default?

Yes  https://github.com/dorny/test-reporter 
image

PR #558, opened 2025-05-22
[DMS-440] Deny update with mismatched etag

semalaiappan:


src/dms/frontend/EdFi.DataManagementService.Frontend.AspNetCore/AspNetCoreFrontend.cs
```diff
@@ -37,6 +37,43 @@ public static class AspNetCoreFrontend
...
 
+    /// <summary>
+    /// Takes an HttpRequest and returns a deserialized request Headers
+    /// </summary>
+    private static async Task<string?> ExtractJsonIfMatchFromAsync(HttpRequest request)
+    {
+        // Try to get If-Match from headers
+        var ifMatch = request.Headers["If-Match"].FirstOrDefault();
+
+        if (string.IsNullOrEmpty(ifMatch))
```
simpat-jesus, line 48: The error response mentions that the stored etag  does not match the If-Match header etag; if a user does not pass the header intentionally or not, but they pass the etag as part of the body, this can cause an un expected response.

I think this method should only get headers from request and return them as a Dictionary to avoid vaing to serialize and deserialize a json headers string.

Once modified, should be renamed too reflecting the changes in the logic

src/dms/core/EdFi.DataManagementService.Core/Middleware/InjectVersionMetadataToEdFiDocumentMiddleware.cs
```diff
@@ -21,10 +23,23 @@ public async Task Execute(PipelineContext context, Func<Task> next)
             DateTimeOffset utcNow = DateTimeOffset.UtcNow;
             string formattedUtcDateTime = utcNow.ToString("yyyy-MM-ddTHH:mm:ssZ");
             context.ParsedBody["_lastModifiedDate"] = formattedUtcDateTime;
-            context.ParsedBody["_etag"] = DateTime
-                .ParseExact(formattedUtcDateTime, "yyyy-MM-ddTHH:mm:ssZ", DateTimeFormatInfo.InvariantInfo)
-                .ToBinary()
-                .ToString(CultureInfo.InvariantCulture);
+
+            if (!string.IsNullOrEmpty(context.FrontendRequest.Header))
```
simpat-jesus, line 27: Adding the header etag as an IfMatch property will cause that hashing the parsedBody results in a different etag, even if the resource was not modified. 

This middleware should only add the etag based on hashing the parsedBody as is

simpat-jesus (changes requested):

simpat-jesus (approved):

PR #573, opened 2025-05-30
[DMS-709] Grant access to Contacts after their StudentSchoolAssociation is recreated

axelmarquezh:
Why the issue happened:

```
               School                                  
                 │                                     
                 ▼                                     
StudentSchoolAssociationAuthorization ─┐               
                 │                     │               
                 ▼                     │               
        StudentSchoolAssociation ──────┤ Cascade delete
                 │                     │               
                 ▼                     │               
               Student                 │               
                 │                     │               
                 ▼                     │               
    ContactStudentSchoolAuthorization ─┘               
                 │                                     
                 ▼                                     
        StudentContactAssociation                      
                 │                                     
                 ▼                                     
               Contact                                 
```
When we delete a `StudentSchoolAssociation`, the cascade delete also deletes the related `StudentSchoolAssociationAuthorization` and `ContactStudentSchoolAuthorization`.
When we recreate the `StudentSchoolAssociation`, the trigger doesn't grant access to the Contacts because the `ContactStudentSchoolAuthorizations` don't exist anymore, so there's no way to get to the `StudentContactAssociations`.

To fix it, I changed the `ON DELETE CASCADE` to `ON DELETE SET NULL`, which keeps the `ContactStudentSchoolAuthorizations` after the `StudentSchoolAssociation` is deleted.
This brings the side effect that when the same `StudentSchoolAssociation` gets deleted multiple times, we end up with many `ContactStudentSchoolAuthorizations` with empty EdOrgIds (garbage). We could consider cleaning them up in an asynchronous recurrent process.

I recommend revisiting this implementation after the RC release, we might consider the next options:
- Add a table similar to `ContactStudentSchoolAuthorization` with the only purpose of having an index toward the `StudentContactAssociation` (by Student Id).
- Investigating if we can add an index directly to the Documents table for the `StudentContactAssociation` StudentId ([see jsonb indexing](https://www.postgresql.org/docs/current/datatype-json.html#JSON-INDEXING)).

### DIFF
--- a/src/dms/backend/EdFi.DataManagementService.Backend.Postgresql/Deploy/Scripts/1001_Create_ContactStudentSchoolAuthorization_Table.sql
+++ b/src/dms/backend/EdFi.DataManagementService.Backend.Postgresql/Deploy/Scripts/1001_Create_ContactStudentSchoolAuthorization_Table.sql
@@ -10,13 +10,13 @@ CREATE TABLE dms.ContactStudentSchoolAuthorization(
     ContactStudentSchoolAuthorizationEducationOrganizationIds JSONB NOT NULL,
     StudentContactAssociationId BIGINT NOT NULL,
     StudentContactAssociationPartitionKey SMALLINT NOT NULL,
-    StudentSchoolAssociationId BIGINT NOT NULL,
-    StudentSchoolAssociationPartitionKey SMALLINT NOT NULL,
+    StudentSchoolAssociationId BIGINT,
+    StudentSchoolAssociationPartitionKey SMALLINT,
     CONSTRAINT FK_ContactStudentSchoolAuthorization_SSA_Document FOREIGN KEY (StudentSchoolAssociationId, StudentSchoolAssociationPartitionKey)
-        REFERENCES dms.Document(Id, DocumentPartitionKey) ON DELETE CASCADE,
+        REFERENCES dms.Document(Id, DocumentPartitionKey) ON DELETE SET NULL,
 
     CONSTRAINT FK_ContactStudentSchoolAuthorization_SCA_Document FOREIGN KEY (StudentContactAssociationId, StudentContactAssociationPartitionKey)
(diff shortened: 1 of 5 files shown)


src/dms/backend/EdFi.DataManagementService.Backend.Postgresql/Deploy/Scripts/1001_Create_ContactStudentSchoolAuthorization_Table.sql
```diff
@@ -10,13 +10,13 @@ CREATE TABLE dms.ContactStudentSchoolAuthorization(
...
-    StudentSchoolAssociationPartitionKey SMALLINT NOT NULL,
+    StudentSchoolAssociationId BIGINT,
+    StudentSchoolAssociationPartitionKey SMALLINT,
     CONSTRAINT FK_ContactStudentSchoolAuthorization_SSA_Document FOREIGN KEY (StudentSchoolAssociationId, StudentSchoolAssociationPartitionKey)
-        REFERENCES dms.Document(Id, DocumentPartitionKey) ON DELETE CASCADE,
+        REFERENCES dms.Document(Id, DocumentPartitionKey) ON DELETE SET NULL,
 
     CONSTRAINT FK_ContactStudentSchoolAuthorization_SCA_Document FOREIGN KEY (StudentContactAssociationId, StudentContactAssociationPartitionKey)
-        REFERENCES dms.Document(Id, DocumentPartitionKey) ON DELETE CASCADE
+        REFERENCES dms.Document(Id, DocumentPartitionKey) ON DELETE CASCADE DEFERRABLE INITIALLY DEFERRED
```
axelmarquezh, line 19: Had to add `DEFERRABLE INITIALLY DEFERRED` so that the foreign key gets checked at the end of the transaction.
Otherwise it was failing during E2E's teardown `delete from dms.document` because it was checking that the SCA existed in the AFTER DELETE TRIGGER after the SCA was deleted (weird).

bradbanister (approved):

PR #580, opened 2025-06-03
[DMS-706] LONG edOrgId support in kafka

simpat-adam:
Jsonb columns holding simple arrays cannot mix int32 and int64, no matter the sort order or else kafka will error and the data will not make it to opensearch. The fix here is to make them arrays of strings always.

### DIFF
--- a/src/dms/backend/EdFi.DataManagementService.Backend.Postgresql.Tests.Integration/Configuration.cs
+++ b/src/dms/backend/EdFi.DataManagementService.Backend.Postgresql.Tests.Integration/Configuration.cs
@@ -7,7 +7,7 @@
 using System.Data;
 using Microsoft.Extensions.Configuration;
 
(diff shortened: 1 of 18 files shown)


src/dms/backend/EdFi.DataManagementService.Backend.Postgresql.Tests.Integration/Configuration.cs
```diff
@@ -7,7 +7,7 @@
 using System.Data;
 using Microsoft.Extensions.Configuration;
 
-namespace EdFi.DataManagementService.Backend.Postgresql.Test.Integration
+namespace EdFi.DataManagementService.Backend.Postgresql.Tests.Integration
```
simpat-adam, line 10: Several misaligned namespaces fixed in this PR.

PR #602, opened 2025-06-13
[DMS-741] Fix nested array issue with array uniqueness validation

bradbanister:


src/dms/core/EdFi.DataManagementService.Core/Middleware/ArrayUniquenessValidationMiddleware.cs
```diff
@@ -0,0 +1,387 @@
...
+            return;
+        }
+
+        await next();
+    }
+}
+
+/// review!
+/// then
+/// other tests
```
JBrenesSimpat, line 387: Those line could be removed

semalaiappan (approved):

PR #614, opened 2025-06-20
[DMS-606] Add extensions to .env file to enable them conditionally

CSR2017:


CSR2017:
Changes covered by https://github.com/Ed-Fi-Alliance-OSS/Data-Management-Service/pull/625

PR #616, opened 2025-06-20
[DMS-711] Fix `Invoke-BulkLoad` call

axelmarquezh:


eng/DatabaseTemplates/Template-Management.psm1
```diff
@@ -570,7 +570,8 @@ function Build-Template {
         -SampleDataDirectory $MinimalSampleDataDirectory `
         -Extension $Extension `
         -BulkLoadClientPaths $bulkLoadClientPaths `
-        -ForceReloadData $true
```
axelmarquezh, line 573: `ForceReloadData` is a switch parameter, so it doesn't need to be set to `$true` (its presence is enough). Therefore, PowerShell was assigning `$true` to the next parameter in the list (`MaxConcurrentConnections`), which, when casted to int, results in `1`; hence, the performance degradation compared to the ODS counterpart.

eng/DatabaseTemplates/Template-Management.psm1
```diff
@@ -266,7 +266,7 @@ function Invoke-DatabaseDump {
     & docker @options | Out-File -FilePath $backupPath -Encoding utf8
 
     # Read CREATE PUBLICATION SQL from the file
-    $publicationSqlFilePath = Join-Path -Path $PSScriptRoot -ChildPath "../../src/dms/backend/EdFi.DataManagementService.Backend.Postgresql/Deploy/Scripts/0099_Configure_Replication.sql"
+    $publicationSqlFilePath = Join-Path -Path $PSScriptRoot -ChildPath "../../src/dms/backend/EdFi.DataManagementService.Backend.Postgresql/Deploy/Scripts/0009_Configure_Replication.sql"
```
axelmarquezh, line 269: [#609](https://github.com/Ed-Fi-Alliance-OSS/Data-Management-Service/pull/609) renamed the script files but missed to update this reference which is making the workflow fail (in [this run](https://github.com/Ed-Fi-Alliance-OSS/Data-Management-Service/actions/runs/15934688082), for example).

PR #633, opened 2025-07-03
[DMS-757] Promote template packages to release

semalaiappan:


vimayya:
@semalaiappan Can you also update on-prerelease.yml to create SBOM and publish the latest pre-release packages ? testsdk doesn't have to be part of the release artifact.

PR #671, opened 2025-08-04
[DMS-793] DMS discovery endpoint doesn't use reverse proxy headers

axelmarquezh:


src/dms/frontend/EdFi.DataManagementService.Frontend.AspNetCore/Program.cs
```diff
@@ -32,6 +32,22 @@
...
+    builder.Services.Configure<ForwardedHeadersOptions>(options =>
+    {
+        options.ForwardedHeaders =
+            ForwardedHeaders.XForwardedFor
+            | ForwardedHeaders.XForwardedHost
+            | ForwardedHeaders.XForwardedProto;
+
+        // Accept forwarded headers from any network and proxy
+        options.KnownNetworks.Clear();
+        options.KnownProxies.Clear();
```
axelmarquezh, line 47: This is the main change.

By default, `UseForwardedHeaders()` only applies the reverse proxy headers when the request comes from localhost's network (see [here](https://learn.microsoft.com/en-us/aspnet/core/host-and-deploy/proxy-load-balancer?view=aspnetcore-9.0#forwarded-headers)).

I originally set `KnownNetworks = { }` to set a blank List and allow for all networks, but after decompiling the .dll I realized that the compiler removed the statement:

image

After testing explicitly setting a new List I found out that the property doesn't have a setter so I now call `Clear()` on the existing List instead (which is how we do it in the [SwaggerUI project](https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-Implementation/blob/c21012c08d695fdc3575485632ec597cb5fe1449/Application/EdFi.Ods.SwaggerUI/SwaggerStartupBase.cs#L57)).

I confirmed that the change works by starting DMS within Docker.

PR #678, opened 2025-08-07
[DMS-402] Exclude domains from OpenApi

simpat-adam:


src/dms/core/EdFi.DataManagementService.Core/OpenApi/OpenApiDocument.cs
```diff
@@ -13,8 +13,77 @@ namespace EdFi.DataManagementService.Core.OpenApi;
...
+                if (domainNode != null)
+                {
+                    if (
+                        domainNode.AsValue().TryGetValue<string>(out string? domainName)
+                        && !string.IsNullOrWhiteSpace(domainName)
+                    )
+                    {
+                        if (_excludedDomains.Contains(domainName, StringComparer.OrdinalIgnoreCase))
+                        {
+                            return true;
```
axelmarquezh, line 45: Haven't tested this, but considering a resource that appears in multiple domains, this seems to exclude it if one of the domains is excluded (instead of only excluding it if all its domains are excluded, per the ticket description).

Could you please confirm?

simpat-adam: Good catch. Fixed logic + added test.

simpat-jesus (approved):

PR #719, opened 2025-09-23
[DMS-817] Improve logging and traceability

simpat-adam:
- **Logging enhancements**
- **Move await statement outside try catch**


src/dms/core/EdFi.DataManagementService.Core/Middleware/ProvideAuthorizationFiltersMiddleware.cs
```diff
@@ -82,8 +82,6 @@ public async Task Execute(RequestInfo requestInfo, Func<Task> next)
             }
 
             requestInfo.AuthorizationStrategyEvaluators = [.. authorizationStrategyEvaluators];
-
-            await next();
```
simpat-adam, line 86: `next()` statement inside the `try` means we are asking to catch any exception downstream. Need to move outside as a final statement and add `return` statements to our catch blocks so we don't call `next()` when we fail.

bradbanister (approved):

PR #720, opened 2025-09-24
[DMS-830] Flattening Metadata Design

bradbanister:


reference/design/flattening-metadata-design.md
```diff
@@ -0,0 +1,1315 @@
...
+    ContactUniqueId VARCHAR(32) NOT NULL,
+    FirstName VARCHAR(75),
+    LastSurname VARCHAR(75),
+    Document_Id UUID NOT NULL,
+    Document_PartitionKey SMALLINT NOT NULL,
+    UNIQUE (ContactUniqueId)
+);
+
+-- Extension root table (one per extension project)
+CREATE TABLE edfi.ContactExtension (
```
vimayya, line 1213: Can you address extension schema in this document.  Core tables go to 'edfi' schema and extension tables go a specific schema for the extension.

reference/design/flattening-metadata-design.md
```diff
@@ -0,0 +1,1315 @@
...
+    OperationalStatusDescriptorId BIGINT REFERENCES dms.Descriptor(Id),
+    ShortNameOfInstitution VARCHAR(75),
+    WebSite VARCHAR(255),
+    -- School-specific fields
+    SchoolTypeDescriptorId BIGINT REFERENCES dms.Descriptor(Id),
+    CharterStatusDescriptorId BIGINT REFERENCES dms.Descriptor(Id),
+    AdministrativeFundingControlDescriptorId BIGINT REFERENCES dms.Descriptor(Id),
+    LocalEducationAgency_Id BIGINT,
+    -- Metadata
+    Document_Id BIGINT NOT NULL,
```
vimayya, line 845: Include CreateDate and LastModifiedDate (in root tables)

reference/design/flattening-metadata-design.md
axelmarquezh: You are likely aware, but for the sake of completeness, I want to remind us that we need to add the project name somewhere, as extensions allow creating resources with the same name as core resources (such as `Homograph.Student`).

reference/design/flattening-metadata-design.md
```diff
@@ -0,0 +1,1315 @@
...
+
+### Descriptor Table Structure
+
+PostgreSQL:
+```sql
+CREATE TABLE dms.Descriptor (
+    Id BIGINT GENERATED ALWAYS AS IDENTITY PRIMARY KEY,
+    -- Lookup keys for integration with DMS
+    ReferentialId UUID NOT NULL,
+    ReferentialPartitionKey SMALLINT NOT NULL,
```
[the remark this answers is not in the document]
bradbanister: Good question! Maybe it's a premature performance optimization? But my thinking is that descriptors are special because they are all on one table per ODS/API. So, it seemed like a good way to speed up descriptor reference resolution by avoiding a join. They are very popular in documents, so much so I initially thought about denormalizing them into their referring tables and not having a dms.Descriptor table at all, but figured that strayed too far from the ODS/API for our purposes here.

A very secondary reason would be for the unique index on referentialIds, ensuring no duplicate descriptors. I don't know that it's really necessary but might be nice since we are shoving them all into one table and descriptors are weird with their case-insensitivity.

PR #721, opened 2025-09-24
[DMS-817] Add indexes for on delete foreign keys

axelmarquezh:


src/dms/backend/EdFi.DataManagementService.Backend.Postgresql/Deploy/Scripts/0002_Create_Alias_Table.sql
```diff
@@ -41,6 +41,7 @@ BEGIN
         ALTER TABLE dms.Alias
         ADD CONSTRAINT FK_Alias_Document FOREIGN KEY (DocumentPartitionKey, DocumentId)
         REFERENCES dms.Document (DocumentPartitionKey, Id) ON DELETE CASCADE;
+
+        CREATE INDEX IF NOT EXISTS IX_FK_Alias_Document ON dms.Alias (DocumentPartitionKey, DocumentId);
```
axelmarquezh, line 45: These new indexes are the main change.
There's a new index per delete/update cascade foreign key.

Per [PgSQL documentation](https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-FK):

> Since a DELETE of a row from the referenced table or an UPDATE of a referenced column will require a scan of the referencing table for rows matching the old value, it is often a good idea to index the referencing columns too. Because this is not always needed, and there are many choices available on how to index, the declaration of a foreign key constraint does not automatically create an index on the referencing columns.

PR #728, opened 2025-09-29
[DMS-819] Run smoke tests on selected pull requests

jleiva-gap:
Update github action to execute if OpenApiDocument or if some relevant packages were changed

### DIFF
--- a/.github/workflows/scheduled-smoke-test.yml
+++ b/.github/workflows/scheduled-smoke-test.yml
@@ -9,6 +9,10 @@ on:
   workflow_dispatch:
   schedule:
     - cron: "30 8 * * 6" # Scheduled the job to run every Saturday at 8 AM UTC
+  pull_request:
+    paths: # Only run on PRs that change relevant files
+      - 'src/Directory.Packages.props'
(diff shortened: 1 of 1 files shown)


.github/workflows/scheduled-smoke-test.yml
```diff
@@ -27,11 +31,67 @@ jobs:
     secrets:
       AZURE_ARTIFACTS_PERSONAL_ACCESS_TOKEN: ${{ secrets.AZURE_ARTIFACTS_PERSONAL_ACCESS_TOKEN }}
 
+  check-changes:
```
simpat-adam, line 34: This is fancy! 
My expectation was that we would just run the smoke tests on any changes to those two files. Here is my reasoning: 

1. Probably > 90% of the time directory.packages.props is touched, it is to bump apiSchema version. 
2. There is no harm in running smokeTest when we occasionally update another library. It might even be wise. 
3. In the future we might change a naming convention that could make this code stop working but we would never know because it is unlikely anyone would remember to look for the test to run on their PR.

simpat-adam (changes requested):

PR #753, opened 2025-11-25
[DMS-880] Tenant support in Config Service

simpat-adam:


jleiva-gap (changes requested):
- Please update eng/docker-compose/.env.example to include the new appSettings parameter, as well as eng/docker-compose/local-config.yml.
 
- Add documentation referencing the new field and the required tenant header value.
 
- I couldn’t find any test cases covering multitenancy.
Will this be addressed in other tickets, or should we add them here as part of this change?

- I manually added the parameter in the .env file with value true, but the DMS Instance container is failing because, when trying to obtain the token, it requires the tenant header value. 
```FTL Critical failure: Unable to load DMS instances from Configuration Service. 

DMS cannot start without proper instance configuration.System.InvalidOperationException: Unable to connect to Configuration Service at [http://dms-config-service:8081/.⁠](http://dms-config-service:8081/) 

Verify that the service is running and the ConfigurationServiceSettings are configured correctly. 

Error: Error response from [http://dms-config-service:8081/connect/token.⁠](http://dms-config-service:8081/connect/token.) Error message: {"error":"Bad Request","message":"The 'Tenant' header is required when multi-tenancy is enabled"}. ```

JBrenesSimpat (approved):

PR #757, opened 2025-12-03
[DMS-886] DMS Multi-tenancy

simpat-adam:


src/dms/tests/EdFi.InstanceManagement.Tests.E2E/Configuration/TestConstants.cs
```diff
@@ -0,0 +1,64 @@
...
+
+/// <summary>
+/// Compile-time constants for E2E test infrastructure.
+/// Database connection strings and names are known at compile time because
+/// they are created by the setup script before tests run.
+/// </summary>
+public static class TestConstants
+{
+    // Database names (created by setup script)
+    public const string Database1Name = "edfi_datamanagementservice_d255901_sy2024";
```
simpat-adam: The tricky thing about the InstanceManagement e2e tests is that we have to create separate databases for separate instances as part of environment setup. We do this in a powershell script before running tests. That is why these hard coded database variables are what they are.

src/dms/tests/EdFi.InstanceManagement.Tests.E2E/Infrastructure/InstanceInfrastructureManager.cs
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    /// <summary>
+    /// Create complete infrastructure for an instance:
+    /// 1. Create Kafka topic
+    /// 2. Clean up any stale replication resources
+    /// 3. Create Debezium connector
+    /// 4. Wait for connector to be ready
+    /// </summary>
+    public async Task SetupInstanceInfrastructureAsync(
```
simpat-adam: See prior comment about this change. This greatly increases the runtime of these e2e tests but they are still completed before the much larger e2e test suite.

PR #766, opened 2025-12-09
[DMS-879] Document Query Indexing Design

bradbanister:


reference/design/document-query-indexing-design.md
```diff
@@ -0,0 +1,218 @@
...
+
+## C# changes to support the design
+- Projection:
+  - Add a projection utility that takes `ResourceSchema.QueryFields` + `EdfiDoc` and returns the deterministic JSONB described above.
+  - Invoke during upsert/update before calling the DB proc, and pass the projection as `p_QueryFields`.
+  - For backfill, reuse the same projector to avoid drift.
+- Query execution:
+  - Adjust `SqlAction` query builders to target `DocumentIndex.QueryFields` instead of `Document.EdfiDoc` for filter predicates, building the `@>` filter JSON from `QueryElements` by `QueryFieldName` (document paths are no longer needed in SQL).
+  - Keep pagination on `DocumentIndex.CreatedAt` and join to `Document` only for the selected page IDs.
+  - Apply `totalCount` on `DocumentIndex` with identical WHERE.
```
axelmarquezh: Note that `totalCount`, `offset`, and `limit` should all be calculated after applying authorization.
We'll need to get creative if we want to merge these changes before the new auth design is implemented.
E2E tests will fail if we disable auth, so we might want to keep these changes in a branch and merge them after the auth redesign is implemented.
Another option is to keep the old read logic alongside the new one, and enable/disable them with a feature flag (an entry in appsettings.json?).

PR #767, opened 2025-12-09
[DMS-887] Tenant selection in swagger ui

simpat-adam:


eng/docker-compose/start-local-dms.ps1
```diff
@@ -49,9 +49,9 @@ param (
     [ValidateSet("keycloak", "self-contained")]
     $IdentityProvider="self-contained",
 
-    # Add initial DMS Instance to Configuration Service
+    # Skip creating initial DMS Instance in Configuration Service
     [Switch]
-    $AddDmsInstance = $true,
+    $NoDmsInstance,
```
simpat-adam: Instead of a switch with a default `$true` which is stylistically discouraged, I changed the name and got rid of the default.

jleiva-gap (approved):

PR #1011, opened 2026-06-08
[DMS-1210] Restore Sample and Homograph smoke-test coverage on relational backend

simpat-adam:
## Summary

Restores Sample and Homograph generated-SDK smoke endpoint sweep coverage for the relational scheduled smoke path.

- Adds a smoke-only four-package relational environment (`DS52 + TPDM + Sample + Homograph`).
- Points the scheduled smoke workflow at the smoke-only relational environment/template package.
- Keeps product/template publication guarded against accidentally publishing the smoke-only package.
- Adds the Sample namespace prefix to smoke credentials so the generated SDK sweep can authorize Sample endpoints.

## Base / dependency

Rebased onto `main` (`b9934026`). The now-merged DMS-1212 (#1007) and DMS-1213 (#1008) commits have been dropped from the stack; only the unmerged DMS-1159 smoke-test stack (15 commits) remains beneath the 6 DMS-1210 commits, because DMS-1210 edits workflow/env files introduced by that stack. The diff vs `main` is the pure smoke/template/workflow surface plus the DMS-1159 relational write-flattener fix. Rebase again once the DMS-1159 stack lands if a mainline-only diff is required.

## Test Plan

- Not rerun locally during the rebase/PR retarget.
- Existing branch workflow history includes `scheduled-smoke-test.yml` validation on `DMS-1210` before this rebase; PR CI should now run against `main` with DMS-1209 included.

### DIFF
--- a/.github/workflows/build-populated-template.yml
+++ b/.github/workflows/build-populated-template.yml
@@ -92,11 +92,27 @@ jobs:
           $useRelationalBackend = [bool]::Parse('${{ inputs.use_relational_backend }}')
           $verifyRestore = [bool]::Parse('${{ inputs.verify_restore }}')
           $requirePopulatedData = [bool]::Parse('${{ inputs.require_populated_data }}')
+          $publishPackage = [bool]::Parse('${{ inputs.publish_package }}')
           $environmentFile = '${{ inputs.environment_file }}'
+          $packageName = '${{ inputs.package_name }}'
 
-          if ($useRelationalBackend -and $environmentFile -ne './.env.template.relational') {
-            throw "use_relational_backend requires environment_file './.env.template.relational'; got '$environmentFile'."
+          $relationalEnvFiles = @('./.env.template.relational', './.env.smoke.relational')
+          if ($useRelationalBackend -and $environmentFile -notin $relationalEnvFiles) {
+            throw "use_relational_backend requires environment_file to be one of: $($relationalEnvFiles -join ', '); got '$environmentFile'."
           }
+
+          # The four-package smoke surface (DS52+TPDM+Sample+Homograph) is an in-run
+          # smoke artifact only. It must never publish under the product template
+          # contract, and must not reuse a product package name.
+          if ($environmentFile -eq './.env.smoke.relational') {
+            if ($publishPackage) {
+              throw "environment_file './.env.smoke.relational' is smoke-only and must not be published; set publish_package to false."
+            }
+            if ($packageName -notlike 'EdFi.Dms.Smoke.Template.PostgreSQL*') {
+              throw "environment_file './.env.smoke.relational' requires a smoke package_name (prefix 'EdFi.Dms.Smoke.Template.PostgreSQL'); got '$packageName'."
+            }
+          }
+
           if ($requirePopulatedData -and -not $verifyRestore) {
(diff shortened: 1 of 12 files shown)


simpat-adam:
## Bundled CI-health fixes (beyond the smoke-test scope)

Heads-up for review: this PR now carries **two commits that fix pre-existing scheduled-CI breakage** unrelated to the Sample/Homograph smoke deliverable. They surfaced while validating the smoke changes through dispatched `scheduled-build` / `scheduled-pre-image-test` runs, and are bundled here so the scheduled lanes go green together rather than in a separate PR.

### 1. `72af20f9` — Drop the legacy E2E lane from scheduled workflows
`scheduled-build` and `scheduled-pre-image-test` have been red on `main` since the 2026-06-20 scheduled runs. Their **legacy E2E lane** (`./.env.e2e`, `Category!=@relational-backend`) fails on Sample-extension scenarios (`POST data/sample/buses` → 404 not-found; `_ext.sample` members absent from responses). The DMS binary itself (unit tests + OpenAPI spec job) and the **relational** lane were green throughout — only the legacy lane tripped the `Fail if an E2E lane failed` guard.

Per the decision to retire that lane, this removes the legacy step + its log/result-upload steps from both scheduled workflows and renames the guard to key solely off the relational lane (plus drops the "Legacy lane" fields from the pre-image Slack payloads). PR builds (`on-dms-pullrequest.yml`) were already relational-only and are unaffected; no workflow references `.env.e2e` any longer.

**Coverage note:** this drops the document/DLL-backed Sample/Homograph extension E2E scenarios from scheduled CI.

### 2. `83896364` — Make the Discovery API root golden tolerant of `informationalVersion`
With the legacy lane gone, `scheduled-pre-image-test`'s relational lane still failed on **1 of 751** tests — the Discovery API root document scenario. The golden pinned `"informationalVersion": "8.0.0"`, but the **published image** under test reported `"0.7.1-alpha.0.230"` (every other field — `applicationName`, `dataModels`, `urls` — matched byte-for-byte).

Root cause is a **version-source divergence**, not a product bug:
- A locally/CI-built image carries no `VERSION` build-arg and falls back to `src/dms/Directory.Build.props` (`InformationalVersion` = `8.0.0`, set by the DMS-1192 rebrand) — so `scheduled-build` passes.
- Published images are built by `on-prerelease.yml` with `VERSION` set to the **MinVer**-derived semver (the `0.7.x-alpha` line), so the app-level `informationalVersion` can never equal the static `8.0.0` the golden demanded.

Pinning a build-time version string in an E2E golden is inherently brittle against published artifacts. The fix follows the step's **own established pattern** — `ThenTheDiscoveryApiRootResponseBodyIs` already strips the top-level `version` field "as it varies between environments" — and extends that to the top-level `informationalVersion` on both expected and actual. The schema-driven `dataModels[].informationalVersion` values (e.g. `"The Ed-Fi Data Standard v5.2.0"`) are unaffected and remain asserted.

### Status
- `scheduled-build`: relational lane was already green; with the legacy lane removed it should go green.
- `scheduled-pre-image-test`: needed **both** fixes (legacy removal + Discovery tolerance).
- Re-validation runs are in flight on this branch.

JBrenesSimpat (approved):

bradbanister (approved):

PR #1043, opened 2026-06-23
[DMS-1136] Data Standard 6 Support (multi-version DS support)

samuellugo-ship-it:
## DMS-1136 — Data Standard 6 Support (multi-version DS support)

Jira: [DMS-1136](https://edfi.atlassian.net/browse/DMS-1136)

> **Draft / WIP.** Landing as a small, reviewable commit series. DS 5.2 remains the default everywhere; the goal is a repeatable mechanism where adding/dropping a data standard version is a cheap, obvious change.

### Goal

Officially support every Ed-Fi-maintained data standard version (DS 5.2 and DS 6.1 now, up to three concurrently) with full schema artifacts, security metadata, sample/template packages, CI/CD coverage, local-dev support, and lifecycle docs.

### Approach (per approved spec)

- **Runtime schema selection** rides the existing `USE_API_SCHEMA_PATH=true` + `SCHEMA_PACKAGES` download path (`src/dms/run.sh`); no DMS runtime C# change.
- **Security metadata**: embedded **multi-version** claims under `Claims/Standards/ds<NN>/`, selected at runtime by `ClaimsOptions.DataStandardVersion` (`DMS_CONFIG_DATA_STANDARD_VERSION`, default `5.2`). ResourceClaim metadata is seeded at load time from the version-selected hierarchy (insert-missing; the static `0009` dbup seed stays the DS 5.2 baseline).
- **CI**: per-workflow `standard_version` matrices (no parsing-heavy config file); distinct per-version package/artifact/SBOM/provenance names. Each multi-version lane delegates to a reusable workflow so one matrix leg = one full build → SBOM → provenance pipeline (per-version outputs stay correlated).
- **Local dev**: `.env.ds52`/`.env.ds61` overlays + a `-DataStandardVersion` parameter composed onto the base env.

### Commit series

> Status is maintained by the PR author as commits land — these markers are intentionally not clickable task-list checkboxes. ✅ = landed · ⬜ = pending. (The branch has been rebased onto `main` more than once; inline hashes would be stale, so they are omitted — see `git log` for current ones.)

- ✅ **1. CMS security-metadata version structure + `ClaimsProvider` selection** — DS 5.2 default preserved; selected by `ClaimsOptions.DataStandardVersion`.
- ✅ **2. Schema artifacts: DS 6.1 `Directory.Packages.props` + `.env.ds61`** — Core/Sample/Homograph `ApiSchema` @ `1.0.332`; **no TPDM** (in 6.1 TPDM is folded into core, so there is no separate TPDM extension).
- ✅ **3. Local scripts: `-DataStandardVersion` overlay composition + Pester tests** — adds the `.env.ds<NN>` overlay + `.derived/` composition, threaded across provisioning/seed/configure/startup.
- ✅ **4. DS 6.1 security-metadata content + runtime ResourceClaim seed** — embedded `Claims/Standards/ds61/Claims.json` (auto-embedded by the `**/Claims.json` glob, no csproj edit); `ClaimsDataLoader` seeds ResourceClaim rows from the version-selected hierarchy (PostgreSQL; no-op on MSSQL), with `0009` left as the DS 5.2 baseline. On reload/upload the seed runs **only after** the claims document is durably applied, so a failed/rolled-back replace cannot orphan ResourceClaim rows.
- ✅ **5a. CI: populated-template lane `standard_version` matrix** — `5.2.0` + `6.1.0` legs via the reusable `build-populated-template.yml`; adding a version = one `include` entry.
- ✅ **5b. CI: SDK + Minimal lanes multi-version** — extracted reusable `build-sdk-package.yml` / `build-minimal-template.yml`; the SDK (`Pkg EdFi.Api.Sdk` / `TestSdk`) and Minimal product workflows are now thin matrix callers with `5.2.0` + `6.1.0` legs and distinct package/provenance names. The reusable extraction keeps each leg's build → SBOM → provenance self-contained so provenance reads the correct leg's `hash-code` (GitHub Actions does not correlate matrix legs across jobs for outputs).
- ✅ **5c. CI: scheduled smoke test `6.1.0` leg** — new `.env.smoke.relational.ds61` (Core + Sample + Homograph). The ODS-published-SDK sweep stays `5.2.0`-only (it consumes the DS-5.2-specific `EdFi.Suite3.OdsApi.TestSdk.Standard.5.2.0`), gated by a `run_ods_sdk_tests` matrix flag; the `6.1.0` API surface is covered by the NonDestructive API tests + the DMS-generated SDK.
- ✅ **6. CI: per-PR E2E DS 6.1 (version-coupled)** — dedicated `run-e2e-tests-ds61` job runs the version-coupled scenarios (XSD metadata + Discovery root, tagged `@StandardVersion-6_1`) against a 6.1 stack; the DS 5.2 four-shard matrix is unchanged. Full DS 6.1 E2E coverage stays in the scheduled smoke test.
- ✅ **7. Lifecycle + local-dev docs** — `docs/DATA-STANDARD-VERSIONS.md` + cross-links from README / RUNNING-LOCALLY / CLAIMS-LOADING-GUIDE.

### Open questions (for review) — all resolved

1. ✅ DS 6.1 = Core + Sample + Homograph `ApiSchema` @ `1.0.332`, **no TPDM** (folded into core). 6.1 sample data confirmed at `Ed-Fi-Data-Standard@v6.1.0`; the 6.1 `SecurityMetadata` export was authored into `ds61/Claims.json`.
2. ✅ `0009_Insert_ResourceClaim.sql` mechanism — `ResourceClaim` is seeded from the version-selected metadata at load time (insert-missing, idempotent), keeping the static dbup seed as the DS 5.2 baseline.
3. ✅ DS 6.1 per-PR relational E2E — runs the version-coupled scenarios per-PR in a dedicated lean lane; full coverage stays in the scheduled smoke test.

[DMS-1136]: https://edfi.atlassian.net/browse/DMS-1136?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


samuellugo-ship-it:
## 🚧 Blocked on 3 answers to continue

We've finished everything that didn't depend on outside input (commits 1, 3, 5a, 7). To keep going we need answers to three questions. **One needs facts, two need a decision.** Each is written to be understandable without deep technical context.

---

### ❓ Q1 — We need the official Data Standard 6.1 package details *(facts)*

**Plain version:** Our software pulls in "Data Standard" content as published software packages, each with an exact name and version number. For Data Standard **5.2** we already know these; for **6.1** we don't — and we can't guess them, they have to come from whoever publishes the Ed-Fi 6.1 packages.

**Why it's blocking:** Without the exact names/versions we can't point the system at the 6.1 content. This one answer unblocks the most work (6.1 schema setup, 6.1 security data, and enabling 6.1 in automated builds).

**What we need back — please fill in the 6.1 column:**

| What it is | What we use for 5.2 (example) | What is it for 6.1? |
|---|---|---|
| Core schema package | `EdFi.DataStandard52.ApiSchema` — version `1.0.332` | ❓ name + version |
| TPDM add-on | `EdFi.DataStandard52.TPDM.ApiSchema` — `1.0.332` | ❓ does it exist? name + version |
| Sample add-on | `EdFi.DataStandard52.Sample.ApiSchema` — `1.0.332` | ❓ does it exist? name + version |
| Homograph add-on | `EdFi.DataStandard52.Homograph.ApiSchema` — `1.0.332` | ❓ does it exist? name + version |
| Database template | `EdFi.Dms.Populated.Template.PostgreSql.5.2.0` | ❓ 6.1 equivalent |

(If any add-on simply doesn't exist for 6.1, just say so — that's a valid answer.)

---

### ❓ Q2 — How should per-version security data be loaded? *(decision / approval)*

**Plain version:** Part of the security configuration is currently loaded into the database by a fixed, run-once setup script that only knows about Data Standard 5.2. To support a second version cleanly, we need to agree on *how* that security data gets loaded so each version gets the right data without breaking 5.2.

**Why it's blocking:** Getting this wrong risks duplicating data or quietly changing 5.2's existing security setup, so we want a decision before building it.

**What we need back — pick one (we recommend the first):**
- ✅ **Recommended:** Load this security data at service startup based on the selected version (matching how the *rest* of the security config already loads), and retire the old fixed script — keeping 5.2's data exactly as it is today.
- **Alternative A:** Keep a setup script, but make a single version-aware script.
- **Alternative B:** Do nothing here for now — leave this particular data at the 5.2 version only, revisit later.

---

### ❓ Q3 — When 6.1 arrives, how aggressively do we test it? *(process / risk)*

**Plain version:** Every time someone proposes a code change we automatically run a heavy test suite. We can either run the **full 6.1 suite on every change** right away, or **hold 6.1 testing to the nightly/scheduled runs** for now and switch it on for every-change testing later.

**Why it matters:** That every-change test lane is already slow and occasionally flaky; adding 6.1 to it now roughly doubles that cost and noise.

**What we need back — pick one (we recommend the first):**
- ✅ **Recommended:** Stage it. Test 6.1 in nightly/scheduled and package builds now; keep the every-change lane on 5.2 for the moment. It's built so turning on every-change 6.1 testing later is a one-line flip.
- **Alternative:** Turn on full 6.1 every-change testing immediately (more confidence, but slower and noisier for everyone).

---

Answering **Q1** unblocks the schema + security content and the 6.1 build entries; **Q2** unblocks the security-data load; **Q3** unblocks the per-change test wiring. Happy to hop on a quick call if that's faster.

samuellugo-ship-it:
## Update — Q1 resolved from the feed ✅ (thanks for the pointer)

Confirmed the DS 6.1 package surface directly against the Ed-Fi feed rather than waiting:

| Package | DS 6.1 | Version |
|---|---|---|
| `EdFi.DataStandard61.ApiSchema` (Core) | ✅ exists | `1.0.332` |
| `EdFi.DataStandard61.Sample.ApiSchema` | ✅ exists | `1.0.332` |
| `EdFi.DataStandard61.Homograph.ApiSchema` | ✅ exists | `1.0.332` |
| `EdFi.DataStandard61.TPDM.ApiSchema` | ❌ **not found** | — (5.2 has it) |

*Method check so this isn't a false negative:* the **5.2** TPDM id (`EdFi.DataStandard52.TPDM.ApiSchema`) resolves with the exact same query, and the feed search for `EdFi.DataStandard61` returns **only** the three packages above. So the TPDM gap is real, not a query/casing artifact. (`1.0.332` is the ApiSchema build version — shared with 5.2 — since the data-standard version lives in the package *name*.)

**Commit 2 landed (`9b6024dd`):** added the three DS 6.1 `PackageVersion` entries + the `.env.ds61` overlay, with TPDM intentionally excluded. DS 5.2 is unchanged.

### One residual question (quick yes/no, non-blocking)

Is TPDM **intentionally** not part of DS 6.1 right now, or is `EdFi.DataStandard61.TPDM.ApiSchema` simply **not published yet**? It doesn't block anything — adding TPDM later is a one-line overlay/props edit — but it decides whether 6.1's template/test surface is *Core only* vs *Core + TPDM*.

### Still open (these are decisions, not lookups)

- **Q2** — how per-version security data should be loaded (we recommend loading it at startup from the version-selected metadata and retiring the static seed script). *Blocks commit 4.*
- **Q3** — whether DS 6.1 joins the per-PR relational-E2E now or stays staged to scheduled/templates/SDK (we recommend staging). *Blocks commit 6.*

samuellugo-ship-it:
## TPDM-in-core confirmed ✅ — commit 2 follow-up

Thanks. TPDM being part of the **core model** in 6.1 (not a separate extension) matches what the feed showed and confirms commit 2 is correct. Refined the `.props` / `.env.ds61` comments and cataloged the version-coupled tests in `2d28c83c`.

**What's done vs. what's still needed for the rest of 6.1:**

- ✅ **Schema artifacts (commit 2):** 6.1 = Core + Sample + Homograph `ApiSchema` @ `1.0.332`. Correct and landed.
- ⏳ **6.1 populated-template build:** needs the 6.1 **sample dataset** (noted as not-ready-yet). Holding the 6.1 template-matrix leg until it lands; its template surface will be **Core only** (no separate TPDM).
- ⏳ **DS 6.1 security metadata (commit 4):** needs the **DS 6.1 SecurityMetadata XML export** (the input `eng/CmsHierarchy` turns into `Claims/Standards/ds61/Claims.json`) **+ the Q2 decision** on the `ResourceClaim` seed mechanism.
- ⏳ **6.1 E2E (commit 6):** the hardcoded-version tests (`XSDMetadata.feature`, `DiscoveryAPI.feature`) need their expected bodies **rewritten for 6.1** — not a `5.2.0`→`6.1.0` search-replace, since 6.1 has **no separate `tpdm` XSD entry/endpoint** and different namespaces/file lists. That needs a running 6.1 stack + the **Q3 decision** on whether 6.1 joins the per-PR lane. Cataloged in `docs/DATA-STANDARD-VERSIONS.md`.

**Net:** the two remaining external inputs are the **6.1 sample dataset** and the **6.1 SecurityMetadata XML export**; the two remaining decisions are **Q2** (ResourceClaim seed) and **Q3** (PR E2E scope).

samuellugo-ship-it:
## 6.1 sample data confirmed ✅ — and what it reveals about the critical path

The DS 6.1 sample data is present and verified at the **`Ed-Fi-Data-Standard@v6.1.0` tag** (`Descriptors/` (266) + `Samples/Sample XML/`, including the now-in-core TPDM entities like `Candidate.xml`). The template workflow already checks out `v${standard_version}`, so that input is wired.

**However, sample data alone doesn't unblock the 6.1 template build.** The populated-template build *loads* that sample data through the **authorized DMS API**, so CMS has to hold **6.1 claims**:

- with `DMS_CONFIG_DATA_STANDARD_VERSION=6.1` and no `Claims/Standards/ds61/Claims.json`, `ClaimsProvider` fail-fasts → **CMS won't start**;
- with 5.2 claims, the 6.1-only (core-TPDM) resources in the sample data **won't be authorized**.

So the **6.1 `SecurityMetadata` XML export is now the critical path** — it's what authors `ds61/Claims.json`, which **both** the 6.1 claims work (commit 4) **and** a green 6.1 populated-template build depend on. I'm intentionally **not** wiring the 6.1 template-matrix leg until that exists, so we don't add CI that's guaranteed to go red.

**Remaining to fully land 6.1:** the **6.1 SecurityMetadata XML** (external input, critical path) + the **Q2** decision (ResourceClaim seed) + the **Q3** decision (PR E2E scope).

samuellugo-ship-it:
## Critical path cleared ✅ — 6.1 security metadata landed, Q2 approach implemented, branch rebased

Picking up from the last update (where the **6.1 `SecurityMetadata` XML export** was the critical path): we sourced it ourselves, so the two items that depended on it are now in the branch. The branch is also **rebased onto `main`** and the PR conflict is resolved.

### ✅ 6.1 claims (`ds61/Claims.json`) — the critical-path input is no longer pending — `d14029e6`

Rather than wait on an external export, we generated the 6.1 security metadata from a local Ed-Fi ODS/API install: `initdev` on the 7.3.2 line builds the `EdFi_Security` database (which *is* the DS 6.1 security DB — there's no separate `_6_1` database), exported it to XML with the Ed-Fi `Security-Metadata-to-XML` script, and ran it through our `eng/CmsHierarchy` converter to produce `Claims/Standards/ds61/Claims.json`. Claim names were normalized to DMS form and the 6.1 surface verified (incl. the now-in-core TPDM domain, `epdm`). It auto-embeds via the existing glob — **no csproj change** — and is covered by new unit tests.

**This unblocks the 6.1 populated-template build's claims dependency:** CMS now starts under `DMS_CONFIG_DATA_STANDARD_VERSION=6.1` and can authorize the 6.1 sample-data load.

### ✅ Q2 — implemented the recommended approach — `37082fdd`

Per the Q2 recommendation (load per-version security data at runtime, keep 5.2 byte-stable):

- **`0009_Insert_ResourceClaim.sql` stays exactly as-is** — the DS 5.2 baseline, so existing ResourceClaim IDs are untouched.
- At initial claims load, the loader now **derives the resource-claim rows from the version-selected `Claims.json` and inserts only the missing ones** (`ON CONFLICT (ClaimName) DO NOTHING`). That's how a version contributes its own resources (e.g. 6.1's in-core TPDM) without editing the dbup script.
- PostgreSQL-only (ResourceClaim endpoints are already unsupported on MSSQL); the new dependency is optional and defaults to a no-op, so nothing else needs wiring.

Happy to switch to Alternative A/B if preferred, but this matches how the rest of the security config already loads.

### ✅ Rebased onto `main` — PR conflict resolved

`main` had moved ahead (incl. DMS-912 and the smoke-test changes). Rebased the branch; the **only** conflict was `scheduled-smoke-test.yml`, where `main` renamed the smoke package/env file and we had matrixed the same jobs — resolved by keeping our `standard_version` matrix structure and feeding `main`'s renamed values into it. PR is **mergeable** again, and the full CMS unit suite is green (313).

### ⏳ Still open

- **Q3** *(decision)* — whether DS 6.1 joins the per-PR relational-E2E now or stays staged to scheduled/templates/SDK. We still recommend staging. *Blocks the per-change test wiring (commit 6).*
- **Validate a real 6.1 populated-template build** — sample data (`v6.1.0`) ✅ and claims ✅ are both in place now, so this is finally runnable; we'll dispatch it (publish gated off) before adding the 6.1 matrix leg, since it triggers real builds.
- **6.1 E2E expectations (commit 6)** — `XSDMetadata.feature` / `DiscoveryAPI.feature` still need their expected bodies rewritten for 6.1 (no separate `tpdm` entry), which needs a running 6.1 stack + the Q3 decision.

**Net:** the last remaining external input (6.1 SecurityMetadata) is resolved and Q2 is implemented; what's left is the **Q3 decision** plus validating and wiring the 6.1 build and E2E lanes.

samuellugo-ship-it:
## Rebased onto the Ed-Fi API rebrand ✅ — DS 6.1 in the populated-template matrix, review feedback addressed

Caught the branch up to `main` (which just merged the **DMS-1192** rebrand, `EdFi.Dms.* → EdFi.Api.*`) and handled the latest review. The PR is mergeable again.

### ✅ Rebased onto `main` (the DMS-1192 rebrand)

`main` renamed the template/SDK artifacts and **renamed the populated-template workflow file itself**
(`EdFi.Dms.Populated.Template.PostgreSQL.yml` → `EdFi.Api.Populated.Template.PostgreSQL.yml`), also
normalizing the package casing to `PostgreSql`. Resolved the two conflicts (the renamed populated
workflow + `scheduled-smoke-test.yml`) by keeping our `standard_version` matrix structure and taking
main's rebranded values, then aligned every DS-version artifact this branch added to the new naming
(`2f0dab29`): the 6.1 matrix leg, the `DATABASE_TEMPLATE_PACKAGE` in `.env.ds52` / `.env.ds61` /
`.env.template.relational.ds61`, the env-overlay Pester fixtures (12/12 green), and the docs. No
`EdFi.Dms.*` template/SDK artifact names remain.

### ✅ DS 6.1 populated-template build leg (`e57d1005`)

The DS 6.1 populated package is now in the **product** workflow matrix
(`EdFi.Api.Populated.Template.PostgreSql.6.1.0`), driven by a new Core-only
`.env.template.relational.ds61` (`DMS_CONFIG_DATA_STANDARD_VERSION=6.1`, allowlisted in the reusable
workflow). Validated via `workflow_dispatch` with publishing gated off — re-dispatched on the rebased
tip. The scheduled smoke test stays `5.2.0`-only until that build is proven end-to-end, so the PR
lane never goes red on an unvalidated leg. SDK + Minimal multi-version remains the deferred
reusable-workflow extraction.

### ✅ Review feedback addressed

- **`-DataStandardVersion` consistency** (`3315833f`): `build-dms.ps1` now composes the `.env.ds<NN>`
  overlay **once** and threads the derived env file through provisioning, seed, configure, and
  startup. Previously only startup composed it, so provisioning/seed could mix DS 6.1 runtime schema
  with a DS 5.2 template/seed.
- **ResourceClaim seeding integration tests** (`3e574282`): live-PostgreSQL coverage that the load
  path inserts the version's resource claims and that re-seeding is idempotent
  (`ON CONFLICT (ClaimName) DO NOTHING`) — the path that previously had no DB-level test.
- **Lifecycle docs** (`fc507202`): now describe the implemented runtime seeding (0009 stays the 5.2
  baseline; the loader inserts the missing version rows) instead of "not yet versioned."

### ⏳ Still open

- **Q3** *(decision)* — whether DS 6.1 joins the per-PR relational E2E now or stays staged to
  scheduled/templates; recommend staging.
- **6.1 E2E expectations** — `XSDMetadata.feature` / `DiscoveryAPI.feature` rewrites (no separate
  `tpdm` entry under 6.1), which need a running 6.1 stack + the Q3 decision.
- **SDK + Minimal multi-version** — the reusable-workflow extraction (a separate, release-critical
  change validated via test tag/dispatch).

**Net:** the branch is rebased onto the rebrand and mergeable; DS 6.1 is built by the populated-template
product workflow (validating now); what remains is the Q3 decision and the E2E / SDK / Minimal follow-ups.

samuellugo-ship-it:
## ✅ Post-rebase CI fix — ResourceClaim seeding test isolation

After the rebase onto `main`, the `Run Integration Tests` check flipped from green to red. Root-caused and fixed — that job is green again (`ff65189d`).

### What broke
`ResourceClaimRepositoryTests` (from main's `[DMS-1148]` resource-claims read endpoints, pulled in by the rebase) started failing — first `ResourceClaimListResult.FailureProjectionIntegrity`, then `ResourceClaimGetResult.FailureNotFound` after a partial fix.

### Why — a test-isolation bug on our side, *surfaced* by the rebase (not the product, not the rebrand)
- The config PostgreSQL integration tests **share one database**, and the common `[SetUp]` (`ClearClaimsTablesAsync`) clears `ClaimSet` + `ClaimsHierarchy` but **not** `dmscs.ResourceClaim` — that table holds the static `0009` seed the read-endpoints projection depends on (every hierarchy claim must resolve to a ResourceClaim row, else `FailureProjectionIntegrity`).
- Our new ResourceClaim-seeding integration test ran `DELETE FROM dmscs.ResourceClaim` to get a clean baseline and didn't restore it. When `ResourceClaimRepositoryTests` ran **after** it, the projection failed. NUnit test order isn't fixed and the rebase changed it — which is exactly why this was green before and red after; the rebase exposed a latent bug, it didn't introduce one.
- A snapshot/restore attempt fixed the integrity error but surfaced a second one: `dmscs.ResourceClaim` has an **identity PK**, so delete + re-insert restored the rows with **new Ids**, and the projection's node Ids derive from those rows → GET-by-id returned `FailureNotFound`. Mutating this shared baseline at all is fragile.

### Fix (`ff65189d`)
Removed the redundant load-path test instead of mutating the shared baseline. Coverage is intact:
- a **unit test** proves the load path calls `SeedResourceClaims` with the rows derived from the hierarchy (FakeItEasy); and
- an **integration test** proves the real PostgreSQL insert + `ON CONFLICT (ClaimName)` idempotency using synthetic claim names it cleans up, so it never touches the `0009` baseline.

### Takeaway for this codebase
`dmscs.ResourceClaim` is a **shared, un-cleared baseline** (the `0009` seed) that the resource-claim read projection depends on; integration tests must not `DELETE`/mutate it without full restore (and even restore is unsafe because of the identity PK). Treat it as read-only from tests.

**Net:** the red check was a latent test-isolation bug in our new integration test, exposed by the test-ordering change the rebase brought in — not a regression in the product or the rebrand. `Run Integration Tests` is green again.


[DMS-1148]: https://edfi.atlassian.net/browse/DMS-1148?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

samuellugo-ship-it:
## Q3 — decided: stage DS 6.1 E2E ✅

Going with the recommended option — **stage it.** With this, all three original open questions are resolved: **Q1** (package facts) ✅, **Q2** (ResourceClaim runtime seeding) ✅ implemented, **Q3** ✅.

### What "stage" means here
- **Per-PR (every-change) relational E2E stays DS 5.2.** `on-dms-pullrequest.yml` runs `build-dms.ps1 E2ETest … -EnvironmentFile './.env.e2e.relational'` (sharded, no `-DataStandardVersion`) — unchanged. The slow, occasionally-flaky per-PR lane is not doubled.
- **DS 6.1 rides the scheduled / package-build lanes.** The populated-template product matrix already builds the 6.1 leg; the scheduled smoke 6.1 leg follows once that build is validated end-to-end.

### Turning per-PR 6.1 on later (the "one-line flip")
Add a `standard_version` dimension to the E2E matrix and pass `-DataStandardVersion 6.1`. `build-dms.ps1` now composes the `.env.ds<NN>` overlay consistently across provisioning, seed, configure, and startup (the consistency fix from the review), so runtime schema / template / security stay in lockstep.

The wiring is the easy part. The actual gating prerequisite is **authoring the version-coupled E2E expectations against a running 6.1 stack** — `XSDMetadata.feature` / `DiscoveryAPI.feature` embed the version, the `http://ed-fi.org/<ver>` namespace, and the per-extension XSD file lists, and 6.1 has **no separate `tpdm` entry** (TPDM folded into core), so those bodies must be rewritten from real 6.1 output, not search-replaced.

Recorded in `docs/DATA-STANDARD-VERSIONS.md` ("Per-PR E2E scope"), commit `23a8bbbf`.

**Net:** per-PR E2E stays 5.2; 6.1 is covered by the scheduled / package-build lanes; the per-PR 6.1 enablement is a small, well-understood flip gated only on a running 6.1 stack for the E2E expectation rewrites.

samuellugo-ship-it:
## DS 6.1 populated-template build — educator-prep root cause + fix 🩺

The 6.1 populated-template `workflow_dispatch` was failing at **"Generate Populated Template package (relational backend)"** while the 5.2 leg stayed green. Root-caused and fixed; **not** a claims bug.

### Why it failed
DS 6.1 folds the educator-preparation model (TPDM) **into core**, so `Ed-Fi-Data-Standard@v6.1.0`'s `Samples/Sample XML` ships **12 educator-prep files that don't exist in v5.2.0**: `Candidate.xml`, `PerformanceEvaluation.xml`, `Path.xml`, `ProfessionalDevelopment.xml`, `RecruitmentAndStaffing.xml`, and the `*-EdPrep.xml` cluster (AssessmentMetadata / EducationOrgCalendar / EducationOrganization / MasterSchedule / StaffAssociation / StudentGradebook / Survey).

Those records are scoped to an **Educator Preparation Provider** organization (`EdOrg 5` + schools `50/60/70`, defined only in `EducationOrganization-EdPrep.xml`). The populated load runs as the `EdFiSandbox` account scoped to the Grand Bend district (`255901`/`19255901`) with **relationship-based authorization** (`RelationshipsWithEdOrgsOnly`), so it has no relationship to the EPP org → the API correctly returns **403** for those records, and the BulkLoadClient treats that as fatal. Two 403 shapes appeared, both inside the excluded files: edorg-relationship-denied (`PerformanceEvaluation` → EdOrg 5) and `Candidate.xml` `StudentSectionAssociation` "SchoolId required".

5.2 is green purely because those files don't exist at v5.2.0. **The ds61 claims are correct** — `epdm` uses `RelationshipsWithEdOrgsOnly`, identical behavior to ds52's `domains/tpdm`; weakening them would be wrong.

### Fix — exclude educator-prep samples (parity with 5.2)
- **`814c0bed`** — `Build-Template -ExcludeEducatorPreparation` loads the populated sample data from a filtered copy that omits the educator-prep files (`*-EdPrep.xml` glob + the 5 standalone `epdm` interchanges; **keeps** core `SpecialEducation.xml`, which references only the in-scope district). New `exclude_educator_preparation` input, set `true` on the 6.1 leg only. Lives in the shared `Template-Management.psm1` so local 6.1 builds get the same behavior. Verified: 12 excluded / 50 kept; no kept file references schools `50/70` (nothing orphaned); restore-verification is count-comparison-based so it isn't affected.
- **`50be1be1`** — optional `standard_version` dispatch input so a single DS leg can be validated without re-running the others (via a small `configure-matrix` prep job). Blank/absent input, and all release/tag runs, build **every** leg.

**Parity:** both versions load their full *core* sample set; neither loads educator-prep populated rows. The future scheduled-smoke 6.1 leg will also set `exclude_educator_preparation: true`.

### Validation
The educator-prep 403s are **resolved** — the 6.1 load now runs through the full core sample set (core `Survey.xml` etc. creating at `201`). A **separate** issue then surfaced deeper in the load: a relational **namespace-based authorization** probe erroring in Postgres with `could not determine data type of parameter $1` (`SELECT CASE WHEN ($1 IS NULL OR $1 = '') THEN throw_error('AUTH1','ns1|0|r') WHEN $1 LIKE ANY($2) ... END`). That's a different dimension from the edorg-relationship auth fixed here — investigating now; follow-up to come.

samuellugo-ship-it:
Pushed `65fe10e1..bda4882f` — first round addressing the review feedback (the two **Major** items).

**`1e08e2b4` — Seed ResourceClaim metadata on claims reload/upload (Major: ClaimsDataLoader.cs:344)**
- `UpdateClaimsAsync` (the upload/reload path, called from `ClaimsUploadService.cs:101` and `:235`) replaced the claim sets + hierarchy but never seeded ResourceClaim metadata, unlike `LoadInitialClaimsAsync`. After an upload/reload that introduced a new claim URI, `ResourceClaimRepository.BuildProjectedNode` (line 83) failed because the hierarchy claim had no matching ResourceClaim row.
- Fix: seed resource-claim metadata derived from the new hierarchy before replacing the document, mirroring initial load. Seeding is insert-missing (`ON CONFLICT (ClaimName) DO NOTHING`) and a no-op on MSSQL, so it's safe/idempotent on every update. It intentionally does **not** delete rows dropped from the hierarchy — extra metadata rows are harmless to the projection (which only fails on *missing* rows). Also added a `DatabaseOperationException` catch so a seeding failure surfaces as `DatabaseFailure`, matching initial load.
- Production code touched: yes (CMS backend). Verified: new unit test + full backend unit suite **314/314 pass**; csharpier-clean.

**`bda4882f` — Tests for the sample-data ed-org parser (Major: Template-Management.psm1:652)**
- Added `eng/DatabaseTemplates/tests/Template-Management.Tests.ps1` covering `Get-EducationOrganizationIdsFromSampleData`: distinct/sorted extraction across `EducationOrganization*.xml`, **DS 6.1 EPP hierarchy coverage** (provider/schools/post-secondary), ignoring non-`EducationOrganization` files, excluding non-edorg ids (e.g. `ProgramId`), and the empty-directory case.
- Wired it into the `Run Bootstrap Pester Tests` job, and also registered `DataStandardEnvironmentFile.Tests.ps1` (added earlier on this branch but never added to the CI list).
- Production code touched: no (tests + CI wiring). Verified: 17/17 Pester pass locally (Pester 5.7.1), PSScriptAnalyzer clean.

Still in progress for this review round: the SDK/TestSdk + scheduled-smoke DS 6.1 legs (Blockers 1 & 2), the docs refresh (Minor), and a question on the per-PR E2E blocker (Blocker 3) — see my next comment.

samuellugo-ship-it:
Pushed `bda4882f..2f38ddf2` — Blocker 1 (SDK/TestSdk DS 6.1) and the docs Minor.

**`60d95b6a` — Matrix the SDK package lanes over Data Standard version (Blocker: Pkg EdFi.Api.Sdk.yml:17 / Pkg EdFi.Api.TestSdk.yml:17)**
- Extracted a shared reusable workflow `build-sdk-package.yml` (mirroring `build-minimal-template.yml` / `build-populated-template.yml`) and converted both `Pkg EdFi.Api.Sdk.yml` and `Pkg EdFi.Api.TestSdk.yml` into thin matrix callers with a `5.2.0` and a `6.1.0` leg.
- The SDK is generated from the OpenAPI document a running DMS serves, so the `6.1.0` leg starts DMS with `-DataStandardVersion 6.1` (the `.env.ds61` overlay) and packages with `-StandardVersion 6.1.0`; the `5.2.0` leg keeps the base `.env.e2e` behavior. Confirmed `configure-local-data-store.ps1` only registers a data store via the CMS API (no `docker compose` recreate), so it doesn't disturb the leg's schema.
- Per-leg `provenance_file` and package names are distinct (`core.sdk.intoto.jsonl` / `core.sdk.6.1.0.intoto.jsonl`, `…Standard.5.2.0` / `…Standard.6.1.0`) so SLSA/SBOM/NuGet artifacts never collide.
- One behavior change to call out: `workflow_dispatch` now defaults to **not** publishing (`publish_package` input, default false), matching the template package callers — so a 6.1 validation dispatch won't push to the feed. Push/release still publish.
- Production code touched: no (CI workflows only). These lanes are tag/release/dispatch-triggered (not PR), so validated **statically**: YAML parse + caller↔reusable input alignment (no undeclared/missing inputs) + unique artifact keys per leg. End-to-end still needs a test tag/dispatch.

**`2f38ddf2` — Refresh DS-versions docs (Minor: docs/DATA-STANDARD-VERSIONS.md:129)**
- Updated the CI section and the "adding a version" step 5: SDK + Minimal are no longer single-version/deferred — both are now matrix callers over per-lane reusable workflows with 5.2.0 + 6.1.0 legs. Noted the SDK legs select the standard via the `.env.ds<NN>` overlay (no template env file) and that the 6.1 SDK/Minimal legs are dispatch-validated before release.

**Deferred for now (per your call):**
- **Blocker 2 (scheduled-smoke 6.1)** and **Blocker 3 (per-PR E2E 6.1)** — skipped pending discussion. Worth flagging: both depend on the 6.1 **populated-template** build, and the latest single-leg `6.1.0` dispatch (`28228124163`) still **fails in the build job**, so adding a smoke leg today would create a perpetually-red scheduled lane (matches the existing "stay 5.2-only until 6.1 is proven" note in the docs). Blocker 3 also reverses the recorded Q3 "stage" decision and needs the version-coupled E2E rewrites. Happy to pick either up once you decide direction.

samuellugo-ship-it:
## DS 6.1 populated template — corrected approach, validated locally ✅

Supersedes the earlier educator-prep-exclusion note. Final approach in commit `4e648b9d` (on top of the sample-edorg parser `65fe10e1`):

**Goal: keep the 6.1 populated template comparable to DS 5.2, not the DS 6.1 educator-prep firehose.** DS 6.1 folds TPDM into core, so its sample set adds ~58 MB of educator-prep data (`ProfessionalDevelopment` ~30 MB, `Candidate` ~16 MB, `PerformanceEvaluation` ~8 MB, the `*-EdPrep` variants) the DS 5.2 template never carried.

- **Filter the educator-prep files** out of the populated load, but **keep `AssessmentMetadata-EdPrep.xml`** — self-contained, namespace-authorized assessment metadata that core `StudentAssessment` sample data references (dropping it orphaned those refs, which was the bug in the first attempt).
- **Scope the sandbox to every sample edorg** (the `65fe10e1` parser) so relationship-authorized resources all load.
- **Tolerate unresolved-reference (409) stragglers**: DS 6.1 interleaves a few educator-prep records into core files (e.g. `StudentAssessmentSample.xml` has assessments for educator-prep students defined in the excluded `Candidate.xml`). The populated load now tolerates **only** HTTP 409 unresolved-reference failures (capped at 50; any 403/500/validation error still fails). The bootstrap load stays strict.

**Validated on a local DS 6.1 relational stack:** 104,487 records load with **0 × 403 and 0 × 500** — every original auth / relational-write blocker is fixed; the 10 expected unresolved-reference 409s are tolerated.

6.1 CI leg dispatched (publish gated off) for the final dump + restore-verify confirmation.

samuellugo-ship-it:
## ✅ 6.1 populated template build — green in CI

Confirms the approach above (commit `4e648b9d`, tests `514cb68a`). 6.1-only dispatch (publish gated off) [run 28245685573](https://github.com/Ed-Fi-Alliance-OSS/Data-Management-Service/actions/runs/28245685573):

- **Generate Populated Template package (relational backend): success** — the load of the lean kept set + tolerated educator-prep unresolved-reference 409s
- **Verify template package restore: success** — dump + restore-verify with `require_populated_data`
- SBOM + Provenance: success (Publish/Promote skipped — publish off)

Pester coverage added for the new filter helpers (`Get-EducatorPreparationSampleFileNames`, `New-EducatorPreparationFilteredSampleDirectory`), 10/10.

This was the gate for the remaining 6.1 work — **scheduled-smoke 6.1 (Blocker 2)** and **per-PR E2E 6.1 (Blocker 3)** are now unblocked.

samuellugo-ship-it:
## DS 6.1 leg added to the scheduled smoke test (commit `3ebaebfa`)

Builds on the now-green 6.1 populated build. Mirrors the DS 5.2 smoke leg:

- **New `.env.smoke.relational.ds61`** — Core + Sample + Homograph (`DataStandard61`), `DMS_CONFIG_DATA_STANDARD_VERSION=6.1`. No separate TPDM (folded into core in 6.1).
- **`6.1.0` leg in both matrices** (build-populated-template + smoke-test); the new env file is allowlisted in `build-populated-template.yml` and remains smoke-only (never published).
- The **educator-prep filter + unresolved-reference tolerance are inherited from `Build-Template`**, so the 6.1 smoke populated build gets them automatically.
- **ODS SDK sweep gated to 5.2 only** via a `run_ods_sdk_tests` matrix flag — it uses the published `EdFi.Suite3.OdsApi.TestSdk.Standard.5.2.0` (DS 5.2-specific). DS 6.1 stays covered by the NonDestructive API tests and the DMS-generated SDK tests.

Production code untouched (CI workflow + env file + Build-Template inheritance only). Validation: this changes `scheduled-smoke-test.yml` (a PR trigger path), so the smoke workflow runs on this PR — the local Docker host is in use by another session, so the CI run is the runtime validation. Watching the 6.1 leg.

samuellugo-ship-it:
## ✅ Scheduled smoke test — 6.1 leg green

[Run 28252051929](https://github.com/Ed-Fi-Alliance-OSS/Data-Management-Service/actions/runs/28252051929) (PR-triggered), confirming commit `3ebaebfa`:

| Leg | Build Populated Template | Smoke Tests |
|---|---|---|
| 5.2.0 | ✅ success | ✅ success |
| 6.1.0 | ✅ success | ✅ success |

The 6.1 **Smoke Tests** job (restore → DMS-SDK generation → NonDestructive API + DMS SDK tests; ODS SDK sweep skipped as DS 5.2-specific) passed, so the leg is validated end-to-end. SBOM + provenance succeeded; publish/promote skipped (smoke is never published).

**Blocker 2 done.** Remaining 6.1 item is **Blocker 3 — per-PR E2E 6.1** (the version-coupled `XSDMetadata`/`DiscoveryAPI` feature rewrites from the captured golden values + a `standard_version` matrix dimension).

samuellugo-ship-it:
## Resolved the 6 CodeQL (PSScriptAnalyzer) findings — commit `b072eaa5`

All 6 code-scanning comments were style nits on the new template helpers (no security/correctness issues), consistent with patterns already in the module. Cleaned them up:

- **Singular nouns:** `Get-EducatorPreparationSampleFileNames` → `Get-EducatorPreparationSampleFileName`; test helper `Invoke-EdPrepNames` → `Invoke-EdPrepName`.
- **`ShouldProcess` + `Write-Host`:** suppressed on `New-EducatorPreparationFilteredSampleDirectory` (it returns the directory path, so `Write-Output` would corrupt the pipeline and `Write-Information` is hidden by default) and `Write-Host` on `Build-Template`'s sandbox-scope progress line — using the `SuppressMessageAttribute` pattern already established in this module.

No behavior change. PSScriptAnalyzer no longer flags these helpers; Pester 10/10. The alerts should auto-resolve on the next scan.

samuellugo-ship-it:
## Per-PR DS 6.1 version-coupled E2E lane (`a68e856a`)

Adds per-PR DS 6.1 E2E coverage, reversing the earlier "stage DS 6.1 E2E" decision (Q3). DS 6.1 now runs on every PR, but **lean** — only the version-coupled scenarios — while full DS 6.1 E2E coverage stays in the scheduled smoke test.

### What changed
- **`XSDMetadata.feature`** — appended DS 6.1 variants of scenarios 01–04 and 07–08 (expected bodies captured from a running 6.1 stack). The two `tpdm` scenarios (05/06) get **no** 6.1 variant — TPDM is folded into core in 6.1, so there's no `tpdm` extension or `/metadata/xsd/tpdm/...` endpoint.
- **`DiscoveryAPI.feature`** — appended a DS 6.1 variant of scenario 01 (root document, no TPDM data model). Scenarios 02–04 are version-independent and unchanged.
- All 6.1 variants are tagged `@relational-backend @StandardVersion-6_1` with **no shard tag**.
- **`.github/workflows/on-dms-pullrequest.yml`** — new dedicated `run-e2e-tests-ds61` job brings up a 6.1 stack (`-DataStandardVersion 6.1`) and runs only those scenarios via `-TestFilter 'Category=@relational-backend&Category=@StandardVersion-6_1'`. Wired into `relational-e2e-summary`. **The DS 5.2 four-shard matrix is byte-for-byte untouched.**
- **`docs/DATA-STANDARD-VERSIONS.md`** — documents the per-PR 6.1 version-coupled lane in place of the staged plan.

### Why the tagging works
The 5.2 shard lanes filter on `@relational-ci-shard-N`, so they never select the (shard-less) 6.1 variants; the 6.1 job filters on the version tag, which exactly **7** scenarios carry (6 XSDMetadata + 1 Discovery, confirmed in the generated code-behind). The tag uses hyphen/underscore because `:` and `.` break the dotnet/NUnit filter. The no-shard filter yields the suffix-less TRX (`EdFi.DataManagementService.Tests.E2E.relational.trx`), which the new job's reporter/upload steps reference.

### Production code
**None.** Changes are limited to E2E feature files (tests), one CI workflow, and docs.

### Verification
- E2E project compiles; the workflow YAML parses; generated code confirms exactly 7 scenarios carry `@StandardVersion-6_1` (none with a shard tag).
- **Ran locally on docker-compose against a real DS 6.1 relational stack** (image rebuilt from this branch; schema provisioned from `EdFi.DataStandard61.*` @ 1.0.332): the exact CI command produced **7 passed / 0 failed**. Live `GET /metadata/xsd` returns ed-fi 6.1.0 + homograph + sample (no tpdm) and `/metadata/xsd/tpdm/files` is 404, matching the new scenarios.

samuellugo-ship-it:
## Follow-up: guardrail fix for the version-coupled lane (`08ec8ba9`)

The first push's CI surfaced a **unit-test guardrail** failure (in `Run Unit Tests`) — the E2E feature-file guardrail `Given_E2E_Feature_File_Guardrails` requires every `@relational-backend` scenario to carry exactly one `@relational-ci-shard-N` tag (so the 5.2 shards stay balanced and non-overlapping). The new DS 6.1 variants are intentionally `@relational-backend` with **no** shard tag (they run in the dedicated `run-e2e-tests-ds61` lane, selected by `@StandardVersion-6_1`), so they tripped it. The guardrail was doing its job — it just didn't know about the new version-lane pattern yet.

### What changed
- `Given_E2E_Feature_File_Guardrails.cs`: the shard-tag rule now **exempts** `@StandardVersion-<NN>` scenarios (they route to a per-version lane, not the shards).
- Added a companion guardrail: every `@StandardVersion-<NN>` scenario **must** be `@relational-backend` and **must not** carry a shard tag — so a version-coupled scenario can never be dropped from CI or run in two lanes at once.

### Production code
**None** — unit test (guardrail) only.

### Verification
Full unit suite green locally: **39 passed / 0 failed** (was 37/1). CSharpier-clean and IDE0055 build passes (both formatter and analyzer agree).

> Note: in the previous run, all relational E2E lanes — including the new `run-e2e-tests-ds61` and the `relational-e2e-summary` gate — already passed; only this unit-test guardrail was red.

samuellugo-ship-it:
## Review-feedback round — 3 Majors + 2 Minors addressed (`d533320d`, `8a1be831`, `9ccc21ed`)

### [Major] Empty Azure Artifacts username on the release-promotion path — `d533320d`
The reusable `build-sdk-package.yml` / `build-minimal-template.yml` (and `build-populated-template.yml`, which shares the pattern) read `secrets.AZURE_ARTIFACTS_USER_NAME` into `env.ARTIFACTS_USERNAME`, but a `workflow_call` workflow only sees secrets declared in `on.workflow_call.secrets` and passed by the caller — only the PAT was. Declared `AZURE_ARTIFACTS_USER_NAME` (`required: false` — consumed only on the publish path; restore uses the PAT) in all three reusable workflows and passed it from every caller (Pkg SDK/TestSdk, Minimal, Populated, scheduled-smoke).

### [Major] Non-atomic ResourceClaim seed in `UpdateClaimsAsync` — `8a1be831`
Seeding committed on its own connection **before** `ReplaceClaimsDocument`'s transaction, so a failed/rolled-back replace orphaned ResourceClaim rows for a document that was never applied. Reordered: the document is applied first, and metadata is seeded **only on success**. The replace never reads ResourceClaim and the metadata is consumed by later projections, so post-commit seeding is correct. The rare post-commit seed failure is surfaced (the idempotent load can be re-run) rather than leaving the system half-applied.

### [Major] Integration tests used the no-op ResourceClaim repository — `8a1be831`
Wired the real `ResourceClaimMetadataRepository` into the integration loader so `LoadInitialClaimsAsync` / `UpdateClaimsAsync` exercise the production seeding path. Added a test asserting an upload that introduces a new claim seeds its ResourceClaim row **through the loader** (also covers the reorder above). Per the known ResourceClaim test-isolation gotcha, the shared `dmscs.ResourceClaim` baseline is snapshotted/restored around each test so seeded rows never leak.

### [Minor] Stale docs — `9ccc21ed`
`docs/DATA-STANDARD-VERSIONS.md` no longer says scheduled-smoke is 5.2.0-only; it now describes both legs and the 5.2-only ODS-SDK gating.

### [Minor] Stale PR description
Updated: commits 4 (security metadata + ResourceClaim seed), 5b (SDK/Minimal multi-version), and 6 (per-PR E2E) marked landed; obsolete "deferred" note removed; open questions closed.

### Production code touched
Yes — `ClaimsDataLoader.UpdateClaimsAsync` (atomicity reorder) and the reusable CI workflows. The rest is tests + docs.

### Verification
- **Backend unit suite: 314/314.**
- **Config PostgreSQL integration (`ClaimsDataLoaderTests` + `ResourceClaimRepositoryTests`): 74/74** against a real Postgres — confirms the reorder, the new loader-path seeding test, and that snapshot/restore keeps the shared ResourceClaim baseline intact for the projection-integrity fixture.
- CSharpier + IDE0055 clean; all 8 changed workflow YAML files parse.

samuellugo-ship-it:
## Scheduled-smoke review fixes (`08d2ae36`)

### [Major] PR trigger missed the DS 6.1 smoke-path dependencies
`scheduled-smoke-test.yml`'s `pull_request.paths` only covered the SDK/smoke tooling, so changes to the files the DS 6.1 smoke lane depends on would skip the PR validation the ticket relies on. Added:
- `.github/workflows/build-populated-template.yml` (the reusable workflow the smoke build calls)
- `eng/docker-compose/.env.smoke.relational*` (per-version smoke template env files, incl. `.ds61`)
- `eng/DatabaseTemplates/**` (template build/restore/filtering logic + its tests — e.g. the DS 6.1 educator-prep filter and bulk-load tolerance)

### [Minor] Stale Slack notification text
The success/failure messages described one Core+TPDM+Sample+Homograph run with ODS SDK tests. Rewritten to reflect the matrix:
- **DS 5.2** — Core + TPDM + Sample + Homograph; DMS NonDestructive API/SDK, DMS Destructive SDK, ODS NonDestructive/Destructive SDK
- **DS 6.1** — Core (TPDM in-core) + Sample + Homograph; DMS NonDestructive API/SDK, DMS Destructive SDK (ODS SDK skipped — DS-5.2-specific)

### Production code
None — CI workflow only (trigger filter + notification text).

### Verification
YAML parses. Note: a wider `paths` filter only ever *adds* PR runs of this workflow (fail-safe — it cannot cause a missed validation), and the Slack `notify-results` job runs only on the scheduled cron (`github.event_name != 'pull_request'`), so the text change has no PR-run effect.

samuellugo-ship-it:
## Review round — 2 Majors addressed (`c5779352`, `dfbcca66`)

### [Major] InstanceManagement E2E provisioning ignored `-DataStandardVersion` — `c5779352`
`setup-local-dms.ps1` forwarded `-DataStandardVersion` to the stack startup but read the raw `./.env.routeContext.e2e` for the post-startup schema provisioning. It now reads the **data-standard-resolved** env (`Resolve-DataStandardEnvironmentFile`) there too, so a non-default version is honored consistently (with no `-DataStandardVersion` it resolves to the base file unchanged — DS 5.2 default preserved).

Worth noting from the investigation: the one-shot Installer this wrapper runs deploys the **legacy/relational-disabled document-store schema, which is identical across DS versions** — it takes only a connection string and reads no ApiSchema (verified: the Installer project has zero ApiSchema/SCHEMA_PACKAGES references; `run.sh` downloads the per-version ApiSchema in a *separate* step that feeds only the running frontend). So there is no per-version relational DDL to thread into the Installer; the running stack already serves the correct versioned ApiSchema against these version-independent databases. The fix makes the env-read consistent with startup and robust if a future overlay adds a value this step reads, and the comment documents why no schema is passed to the Installer.

### [Major] Populated-template unresolved-reference tolerance applied to every version — `dfbcca66`
`Build-Template` passed `-AllowUnresolvedReferences` unconditionally for populated loads, weakening DS 5.2 release validation. It is now gated on whether the educator-prep filter actually excluded files: `New-EducatorPreparationFilteredSampleDirectory` returns the source directory **unchanged** when there is nothing to exclude (DS 5.2) and a new filtered directory **only** when it drops educator-prep files (DS 6.1). So `-AllowUnresolvedReferences:$educatorPrepFiltered` keeps **DS 5.2 (and any version with nothing filtered) strict** — any non-zero BulkLoad exit fails the build — while only the filtered DS 6.1 load tolerates the expected unresolved-reference 409s.

### Production code touched
`Build-Template` / `setup-local-dms.ps1` are build/test tooling (no DMS runtime C#).

### Verification
- Both files parse; PSScriptAnalyzer shows no new findings (pre-existing Write-Host/BOM warnings only, already accepted by the CI PSScriptAnalyzer check).
- Template-Management Pester suite **10/10** — including the `New-EducatorPreparationFilteredSampleDirectory` "returns the source directory unchanged when there is nothing to exclude" test, which is exactly the signal the DS 5.2 strict-mode gate relies on.

samuellugo-ship-it:
## Failure-path coverage for the ResourceClaim atomicity fix (`8b24d57f`)

### [Major] Prove `SeedResourceClaims` is skipped when the replace fails
The existing failure tests used the default loader (no `IResourceClaimMetadataRepository`), and the fake-repository test only covered the success path — so the orphaning failure mode wasn't actually proven.

Added a parameterized test in `Given_UpdateClaimsAsync_is_called` (via `TestCaseSource`) over all three failure results the reviewer named — **DatabaseFailure, ValidationFailure, MultiUserConflict**. Each case:
- wires the **fake `IResourceClaimMetadataRepository`** into the loader,
- mocks `ReplaceClaimsDocument` to return that failure,
- asserts the mapped `ClaimsDataLoadResult` failure type, **and**
- asserts `SeedResourceClaims(...).MustNotHaveHappened()`.

That directly proves a failed/rolled-back replace never seeds ResourceClaim metadata, so it cannot orphan `dmscs.ResourceClaim` rows for a claims document that was not applied — locking in the ordering fix from `8a1be831` (`ClaimsDataLoader.UpdateClaimsAsync` seeds only after `ReplaceClaimsDocument` returns Success).

### Production code touched
None — test-only.

### Verification
Full backend unit suite **317/317** (the `UpdateClaimsAsyncIntegrationTests` fixture is 10/10: the 7 existing tests + the 3 new failure cases). CSharpier-clean, IDE0055 build passes.

samuellugo-ship-it:
## Rebased onto latest `main`

Force-pushed (`--force-with-lease`) after a clean rebase onto current `main` (`368cfd6a`) — picks up the recently merged DMS-1233 / DMS-1229 / DMS-1224 + the nuget-docker change. **No conflicts.** Branch tip is now `4811483b`; commit hashes from earlier comments are rewritten by the rebase.

All the review-feedback commits (workflow username secret, claims-update atomicity + its failure-path test, loader integration coverage, scheduled-smoke paths/Slack, InstanceManagement E2E provisioning, populated-template tolerance gating) are included.

**Post-rebase verification (local):** config Backend unit **317/317**, DMS E2E test project compiles, DMS unit (E2E feature-file guardrails) **39/39**. No production code touched by the rebase itself.

GitHub reports the PR `MERGEABLE` (the `BLOCKED` state is branch-protection: required checks/reviews, not a conflict).

samuellugo-ship-it:
## Review round — 2 Majors (`fab8608d`, `5ad660c6`)

### [Major] `InstanceE2ETest -DataStandardVersion 6.1` was ignored — `fab8608d`
`build-dms.ps1`'s `InstanceE2ETests` didn't accept `-DataStandardVersion`, the `InstanceE2ETest` dispatcher didn't pass it, and **both** `setup-local-dms.ps1` calls omitted it — so a DS 6.1 request still produced a DS 5.2 route-context stack, even though `setup-local-dms.ps1` already honors the parameter (the startup + provisioning fix from the prior round). Added the parameter to `InstanceE2ETests`, threaded it from the dispatcher, and forwarded it to both setup-script calls (`-SkipDockerBuild` and the plain branch).

This is the **front half of the same `-DataStandardVersion` chain** whose back half I fixed earlier (setup-local-dms.ps1 reading the resolved env for provisioning) — they're complementary links, not a re-patch. The full chain is now: dispatcher → `InstanceE2ETests` → `setup-local-dms.ps1` (startup + resolved-env provisioning). The default (empty version) is byte-identical to the prior `& $instanceSetupScript` call, so DS 5.2 behavior is unchanged, and the `E2ETest` path is untouched.

### [Major] Post-commit ResourceClaim seed-failure branch was untested — `5ad660c6`
`UpdateClaimsAsync` applies the document first, then seeds ResourceClaim; if the post-commit seed throws `DatabaseOperationException` it returns `DatabaseFailure` (not Success). Coverage existed for successful seeding and for a failed replace skipping the seed, but not for **replace succeeds → seed fails**. Added a unit test: `ReplaceClaimsDocument` returns Success, the fake `SeedResourceClaims` throws `DatabaseOperationException`, and the loader returns `ClaimsDataLoadResult.DatabaseFailure` (asserting the replace ran and seeding was attempted once).

### Production code touched
Only `build-dms.ps1` (a new optional parameter + forwarding). Issue 2 is test-only.

### Verification
- Full backend unit suite **318/318** (the `UpdateClaimsAsyncIntegrationTests` fixture now covers all three branches: success-seeds, failed-replace-skips-seed, succeeded-replace-seed-fails).
- `build-dms.ps1` parses; PSScriptAnalyzer reports no Error-severity findings; both `setup-local-dms.ps1` calls confirmed to forward the version. CSharpier-clean.

### Cross-check for regressions (you flagged the patch-loop risk)
Issues 1 and 2 touch disjoint code (`build-dms.ps1` vs a test), and neither touches the prior fixes' files, so resolving one can't reopen the other. The InstanceE2E chain is now complete end-to-end rather than half-wired.

samuellugo-ship-it:
## Proactive hardening + full audit (`c0ad7ab9`)

To get ahead of the one-at-a-time findings, I audited the two recurring themes end-to-end rather than just patching the reported spots.

### `-DataStandardVersion` threading — audited every stack command in `build-dms.ps1`
| Command | Honors `-DataStandardVersion`? |
|---|---|
| `E2ETest` | ✅ `Get-E2ETestEnvironmentContext` composes the overlay → resolved env |
| `InstanceE2ETest` | ✅ fixed this round (dispatcher → `InstanceE2ETests` → both setup calls → setup-local-dms.ps1 startup + provisioning) |
| `StartEnvironment` | ✅ `Start-DockerEnvironment` composes the overlay standalone (build-dms.ps1:890-905) |
| `Run` | n/a — bare `dotnet run` of the frontend; no Docker stack / DB / schema provisioning |
| `DockerRun` | n/a — single container with a fixed `./src/dms/.env`; no compose / DB / provisioning |

All three **provisioning** paths now honor the version; `Run`/`DockerRun` have nothing to provision (threading them would be speculative and touch an unrelated base env). No remaining gaps.

### Claims failure-branch coverage — now exhaustive
`UpdateClaimsAsync` has three behavioral branches, all now covered in `UpdateClaimsAsyncIntegrationTests`:
1. **replace fails → no seed** — parameterized over **all four** `ClaimsDocumentUpdateResult` failure types (DatabaseFailure, ValidationFailure, MultiUserConflict, and now **UnexpectedFailure** — this commit).
2. **replace succeeds, seed succeeds → Success** — existing test.
3. **replace succeeds, seed throws → DatabaseFailure** — added in `5ad660c6`.

### Production code touched
None this commit (test-only). The InstanceE2ETest fix (`fab8608d`) was the only production change this round (a build-script parameter + forwarding).

### Verification
`UpdateClaimsAsyncIntegrationTests` 12/12; full backend unit suite green. CSharpier-clean.

samuellugo-ship-it:
## Review round — 3 Majors (`cd57f54c`, `4b05e5bd`)

### [Major] `.nupkg` discovery broken in the SDK + Minimal publish steps — `cd57f54c`
`build-minimal-template.yml:374` and `build-sdk-package.yml:324` used `Get-ChildItem -Path $_ -Name -Include *.nupkg`, where `$_` is null outside a pipeline. Dropped the spurious `-Path $_` so both match the working `build-populated-template.yml` form (`Get-ChildItem -Name -Include *.nupkg`), making the DS 6.1 SDK/TestSDK/Minimal publish lanes discover the built package on release/tag runs.

### [Major] `-AllowUnresolvedReferences` classifier had no Pester coverage — `4b05e5bd`
Extracted the tolerance decision from `Invoke-BulkLoad` into `Get-BulkLoadFailureClassification` and added Pester tests. `Invoke-BulkLoad` still streams the (possibly hundreds-of-MB) log via `Select-String` and passes only the parsed 4xx/5xx lines to the classifier, so memory use and behavior are unchanged. New tests cover all the cases you named plus boundaries: tolerated-within-cap, **cap boundary**, **mixed fatal failures**, **fatal 5xx**, **over-cap unresolved references**, **zero parsed failures**, and **strict DS 5.2** (flag off). Template-Management Pester now **17/17**.

### Thoroughness / breaking the loop
I audited the publish pattern across **all** workflows, not just the two flagged:
- `build-populated-template.yml:483` — already correct (no `-Path $_`); no change.
- **`on-prerelease.yml:181` and `:518` carry the identical `-Path $_` pattern** (DMS + Config package push). That file is **not part of this PR's diff** (a pre-existing release workflow) and the pattern works in pwsh 7 via the `-Path $null`→cwd fallback, so I did **not** modify it here — changing an unrelated, release-critical workflow I can't validate is riskier than the benefit. **Flagging it for a separate cleanup PR.** Happy to include it here if you'd prefer.

### Production code touched
`Template-Management.psm1` (behavior-preserving extraction) and the two reusable workflows. The classifier behavior is unchanged — the green DS 6.1 populated build (10×409 tolerated) classifies identically.

### Verification
Both PS files parse; Template-Management Pester **17/17**; PSScriptAnalyzer reports no Error findings and no new Warning findings (the only `PSUseSingularNouns` hit is the pre-existing `Get-UserSchemaNames`); both edited workflow YAMLs parse.

samuellugo-ship-it:
## Tracking note — same `.nupkg` discovery bug in `on-prerelease.yml` (separate PR)

While fixing the SDK/Minimal publish steps in this PR (`cd57f54c`), I audited every workflow for the same pattern and found a **pre-existing** instance of the identical bug in `on-prerelease.yml`, which is **not part of this PR's diff**:

- `on-prerelease.yml:181` — DMS package push
- `on-prerelease.yml:518` — Config package push

Both use `$artifact = (Get-ChildItem -Path $_ -Name -Include *.nupkg)`, where `$_` is null outside a pipeline. It currently works only by accident (in pwsh 7, `-Path $null` falls back to the current directory), so it's fragile rather than visibly broken.

**Decision: intentionally NOT fixing it in this PR.** It's a release-critical workflow unrelated to DMS-1136, and it can only be validated by an actual prerelease tag — so changing it here would mix scope and add unvalidatable risk. Tracking it for a **separate cleanup PR**.

**Fix when that PR is opened** (mirror what this PR did to the reusable workflows): drop the spurious `-Path $_` so both lines read `$artifact = (Get-ChildItem -Name -Include *.nupkg)`.

For reference, `build-populated-template.yml` was already correct, and `build-sdk-package.yml` / `build-minimal-template.yml` are fixed here.

samuellugo-ship-it:
## Review round — 2 Majors (`ef377c16`, `95f572ac`)

### [Major] BulkLoad tolerance now enforces HTTP 409 — `ef377c16`
`Get-BulkLoadFailureClassification` keyed only on the `unresolved-reference` marker, so a non-409 failure line containing that text (e.g. a 500) could have been tolerated within the cap — contradicting the documented "HTTP 409 unresolved-reference" contract. It now requires **both** `- 409 -` **and** the marker; everything else is fatal:
- a non-409 line (even one mentioning `unresolved-reference`),
- a 409 without the marker (e.g. a duplicate-key conflict),
- any other 4xx/5xx.

The caller pre-filters to `- <code> - {` lines and real unresolved-reference failures are 409, so the green DS 6.1 populated build (10×409) classifies identically — the change only closes the hypothetical non-409-with-marker hole. Added the two missing Pester cases (non-409 line **with** the marker → fatal; 409 **without** the marker → fatal). **Template-Management Pester now 19/19.**

### [Major] DMS PR workflow now triggers on `src/config` changes — `95f572ac`
The DMS E2E lanes (incl. `run-e2e-tests-ds61`) build and run the Config Service from `src/config`, but the path filter omitted `src/config/**` — so a config-only change (ds61 `Claims.json`, `ClaimsProvider`, ResourceClaim seeding) skipped the DS 6.1 DMS validation. Added `src/config/**` to **both** the `push` and `pull_request` filters (symmetric with the existing `src/dms/**`). A wider path filter only ever **adds** PR runs, so it cannot cause a missed validation.

### Production code touched
`Template-Management.psm1` (classifier tightening). The workflow + tests are CI/test-only.

### Verification
Template-Management Pester **19/19**; `Template-Management.psm1` parses with no PSScriptAnalyzer errors; `on-dms-pullrequest.yml` parses and `src/config/**` is present in both triggers. The classifier tightening is behavior-preserving for genuine 409 unresolved-reference loads.

### Audit note
This is the same classifier extracted last round (`4b05e5bd`); the 409 gap was a latent issue carried from the original inline logic, now fixed and regression-tested. No new patterns of this kind remain in the tolerance path.

samuellugo-ship-it:
## Fix: DS 6.1 populated-template build failure (regression I introduced) — `510689dc`

### What failed
`Build Populated Template 6.1.0 / Build Populated Template Package` (scheduled-smoke run 28274591369) failed at "Generate Populated Template package": the bulk load hit the expected 10× HTTP 409 unresolved-reference failures (kept records referencing the excluded educator-prep entities), but they were treated as **fatal** instead of tolerated.

### Did we introduce it? Yes — in the round-2 gating (`dfbcca66`)
That commit made the unresolved-reference tolerance conditional:
```powershell
$populatedSampleDataDirectory = New-EducatorPreparationFilteredSampleDirectory -SourceDirectory $PopulatedSampleDataDirectory
...
$educatorPrepFiltered = $populatedSampleDataDirectory -ne $PopulatedSampleDataDirectory
```
**PowerShell variable names are case-insensitive**, so `$populatedSampleDataDirectory` and `$PopulatedSampleDataDirectory` are the *same variable*. The filtered result was assigned back into the parameter, and the comparison was **always `$false`** → `-AllowUnresolvedReferences` was silently off for every version → the 6.1 build's 409s became fatal.

It slipped through because the populated build was last validated green (`28245685573`) *before* that gating (which had `-AllowUnresolvedReferences` unconditionally on), and the local Pester covers the filter helper + the classifier but not Build-Template's inline gating.

### The fix
Derive the gate from the unit-tested `Get-EducatorPreparationSampleFileName` helper (excluded-file count > 0) **before** filtering — no fragile path comparison:
```powershell
$educatorPrepFiltered = @(Get-EducatorPreparationSampleFileName -SourceDirectory $PopulatedSampleDataDirectory).Count -gt 0
```
DS 6.1 (educator-prep present) → tolerate; DS 5.2 (none) → strict. This restores the exact behavior of the last green populated build for 6.1.

### Verification
- Reproduced the case-insensitivity bug (self-comparison returns `$false`).
- Confirmed the fix end-to-end locally: gate **True** for a DS-6.1-like dir, **False** for core-only; the gate forwards through the switch to the classifier, which **tolerates** a real 409 log line.
- Pester **19/19**, no PSScriptAnalyzer errors.
- The scheduled-smoke workflow re-runs on this push (it triggers on `eng/DatabaseTemplates/**`), which will confirm the 6.1 build end-to-end.

### Honest note on the pattern
This bulk-load tolerance path has now been changed across several rounds (gating, classifier extraction, 409 enforcement) without a real populated-build run, since the green build predated them. The classifier is now unit-tested and the gate rests on a tested helper, but the integration is only truly exercised by the scheduled-smoke build — which is now wired to re-run on changes to that code.

samuellugo-ship-it:
## Rebased onto `main` to resolve the merge conflict

Rebased `DMS-1136` onto current `origin/main` (picked up #1045 DMS-1133 NuGet package security, #1042 ReadChanges authorization, #1053 abstract-identity reference column naming, and #1062 design spec). Kept the branch linear rather than introducing a merge commit, consistent with prior rebases on this PR.

### Conflict
One file, one region: `.github/workflows/on-dms-pullrequest.yml` — both `main` (#1045) and this branch appended to the bootstrap Pester test list. Resolved as a **union**: `LockFileParity.Tests.ps1` (main) + `DataStandardEnvironmentFile.Tests.ps1` and `Template-Management.Tests.ps1` (ours).

The path filters (`src/nuget.config` from main vs. `src/config/**` from us) and `build-dms.ps1` (main's `--locked-mode` restore vs. our `-DataStandardVersion` threading) auto-merged cleanly; verified both sides coexist.

### One follow-up commit
`a700e4c1` adds `src/nuget.config` to the `run-e2e-tests-ds61` job's buildx cache key. That job was modeled on the pre-#1045 5.2 E2E job, so it missed the cache-key update that #1045 applied to the other Docker-build jobs — now the DS 6.1 image cache also invalidates when `nuget.config` changes.

### Verification (local)
- `build-dms.ps1` parses with 0 errors; main's locked-mode logic intact alongside the `-DataStandardVersion` threading
- All 5 changed workflows parse as valid YAML
- backup→HEAD diff contains only main's incoming work — no DMS-1136 content lost
- No production application code touched in the rebase resolution (CI workflow + build script only)

CI re-running on the rebased tip.

simpat-adam (approved):

PR #1045, opened 2026-06-23
[DMS-1133] Improve NuGet Package Security

JBrenesSimpat:
## Summary

  - Lock the full dependency graph (direct + transitive). Enables RestorePackagesWithLockFile in both src/dms/Directory.Build.props and src/config/Directory.Build.props (no shared parent props exists, so it must live in both), and commits a packages.lock.json for every project under src/. This gives an auditable, diffable record of exactly what gets restored — closing the gap where Central Package Management pins only direct versions.
  - Enforce the lock in PR CI. Both PR workflows gain a verify-lock-files gate that runs dotnet restore <solution> --locked-mode, failing fast (NU1004) if a committed lock file drifts from the .csproj / Directory.Packages.props state.
  - Enforce the lock in the Docker image builds (the security-critical artifact). Both Dockerfiles now COPY each project's packages.lock.json alongside its .csproj and pass --locked-mode to every dotnet restore.
  - Dependabot cooldown. .github/dependabot.yml adds a cooldown block (patch 5d / minor 7d / major 14d) so freshly published versions sit long enough for vendors to flag poisoned releases — version updates only; security updates still arrive immediately.
  - Auto-regenerate lock files on Dependabot PRs. New dependabot-lock-file.yml runs dotnet restore --force-evaluate on both solutions and commits/pushes regenerated lock files, covering transitively-affected projects that Dependabot's native updater misses (dependabot-core #13950 (https://github.com/dependabot/dependabot-core/issues/13950)). Pushes with the built-in GITHUB_TOKEN — no PAT/secret required.
  - Keep the lock file out of published output. The Web SDK auto-includes *.json as Content, so the two frontend .csproj files add <Content Remove="packages.lock.json" /> to avoid publishing/packing it into the .nupkg.
  - Guard the two hand-maintained mirrors. New LockFileParity.Tests.ps1 (Pester) asserts (T1) each Dockerfile's lock-file COPY list mirrors its .csproj COPY list, and (T2) the Directory.Build.props that build-dms.ps1 / build-config.ps1 regenerate reproduces the committed props' restore-relevant content (RestorePackagesWithLockFile + analyzer PackageReferences). The build-script here-strings were updated to match.
  - Developer docs. New docs/NUGET-LOCK-FILES.md explains the scheme end-to-end, the Dockerfile COPY-list maintenance hotspot, how to regenerate locally, and the recovery steps for the GITHUB_TOKEN-push recursion edge case.

##  Test plan

  - [ ] On a clean checkout, dotnet restore ./src/dms/EdFi.DataManagementService.sln --locked-mode and dotnet restore ./src/config/EdFi.DmsConfigurationService.sln --locked-mode both succeed.
  - [ ] Negative check: add a PackageReference (or bump a version) without regenerating, then run --locked-mode → restore fails with NU1004, confirming the gate actually catches drift.
  - [ ] dotnet restore <sln> --force-evaluate on both solutions produces no git diff (committed lock files are already current).
  - [ ] Both Docker images build cleanly: ./build-dms.ps1 / ./build-config.ps1 (or docker build on src/dms/Dockerfile and src/config/Dockerfile) — the in-image --locked-mode restores succeed.
  - [ ] Pester parity suite passes: Invoke-Pester -Path eng/docker-compose/tests/LockFileParity.Tests.ps1 -Output Detailed (both dms and config cases, T1 + T2).
  - [ ] Confirm the lock file is excluded from output for both frontends: publish/pack the two AspNetCore frontend projects and verify packages.lock.json is absent from the publish dir and the .nupkg.
  - [ ] In PR CI, confirm the verify-lock-files gate runs on both workflows, the config run-parity-pester-tests job runs, and DMS includes LockFileParity.Tests.ps1 in its bootstrap Pester run — and that all are wired into the downstream needs: (gating event_file / summary).
  - [ ] .github/dependabot.yml parses (cooldown uses the current *-days keys, not the deprecated semver-patch: short form).
  - [ ] Manual / first live Dependabot PR (cannot be triggered on demand): the dependabot-lock-file workflow runs only for dependabot[bot] on src/Directory.Packages.props changes, regenerates + pushes lock files, and --locked-mode gates pass — exercising the recovery path in docs/NUGET-LOCK-FILES.md if the GITHUB_TOKEN push doesn't re-trigger CI.

### DIFF
--- a/.github/workflows/dependabot-lock-file.yml
+++ b/.github/workflows/dependabot-lock-file.yml
@@ -1,0 +1,54 @@
+# SPDX-License-Identifier: Apache-2.0
+# Licensed to the Ed-Fi Alliance under one or more agreements.
+# The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+# See the LICENSE and NOTICES files in the project root for more information.
+
+# Dependabot bumps src/Directory.Packages.props but does not update the lock files for
+# projects affected only transitively via <ProjectReference> (dependabot-core #13950).
+# This workflow regenerates the full lock graph on Dependabot PRs. It pushes with the
+# built-in GITHUB_TOKEN (contents: write) -- no PAT / Dependabot secret is required.
+# See docs/NUGET-LOCK-FILES.md for the recovery steps when the push does not auto-run the gate.
+
+name: Update Lock File on Dependabot PR
+
+on:
+  pull_request:
+    branches:
+      - main
+    paths:
+      - "src/Directory.Packages.props"
+
+permissions:
+  contents: write
+
+env:
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+jobs:
+  update-lock-file:
+    name: Regenerate packages.lock.json
+    if: github.actor == 'dependabot[bot]'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout PR branch
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          ref: ${{ github.head_ref }}
+
+      - name: Setup dotnet
+        uses: actions/setup-dotnet@67a3573c9a986a3f9c594539f4ab511d57bb3ce9 # v4.3.1
+        with:
+          dotnet-version: "10.0.x"
+
+      - name: Restore packages and regenerate lock files
+        run: |
+          dotnet restore src/dms/EdFi.DataManagementService.sln --force-evaluate
+          dotnet restore src/config/EdFi.DmsConfigurationService.sln --force-evaluate
+
+      - name: Commit updated lock files
+        run: |
+          git config user.name "dependabot[bot]+commit"
+          git config user.email "noreply@ed-fi.org"
+          find src -name "packages.lock.json" -print0 | xargs -0 --no-run-if-empty git add
+          git diff --staged --quiet || git commit -m "Update packages.lock.json after Dependabot package update"
+          git push
--- a/.github/workflows/dependabot-lock-file.yml
+++ b/.github/workflows/dependabot-lock-file.yml
@@ -17,6 +17,7 @@ on:
       - main
     paths:
       - "src/Directory.Packages.props"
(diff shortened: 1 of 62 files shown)


bradbanister (approved):

.github/workflows/dependabot-lock-file.yml
```diff
@@ -0,0 +1,68 @@
...
+  cancel-in-progress: true
+
+jobs:
+  update-lock-file:
+    name: Regenerate packages.lock.json
+    if: github.actor == 'dependabot[bot]'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout PR branch
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
stephenfuqua, line 38: This is a very old and deprecated version of checkout.

.github/workflows/dependabot-lock-file.yml
```diff
@@ -0,0 +1,68 @@
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          git config user.name "dependabot[bot]+commit"
+          git config user.email "noreply@ed-fi.org"
+          find src -name "packages.lock.json" -print0 | xargs -0 --no-run-if-empty git add
+          if git diff --staged --quiet; then
+            echo "No lock file changes to commit."
+          else
+            git -c core.hooksPath=/dev/null commit --no-verify -m "Update packages.lock.json after Dependabot package update"
+            git -c core.hooksPath=/dev/null push --no-verify "https://x-access-token:${GITHUB_TOKEN}@github.com/${{ github.repository }}.git" "HEAD:${{ github.head_ref }}"
```
stephenfuqua, line 67: @JBrenesSimpat this line was flagged as a [security risk](https://github.com/Ed-Fi-Alliance-OSS/Data-Management-Service/security/code-scanning/1409), I think by the Scorecard process - which is interesting, as I didn't remember the scorecard action covering something like this!

In this line and line 40, why did you need to specify the `HEAD` explicitly instead of just pushing to the default branch?

I believe it would be safe and correct to leave line 40, then change line 67 to just push to `HEAD`.

PR #1049, opened 2026-06-24
Overnight mutation testing

bradbanister:
This PR improved test coverage and quality via an agent mutation testing loop I've been experimenting with and ran overnight. Separately, I fixed an annoying flaw in DMS container health checks.


eng/docker-compose/local-config.yml
```diff
@@ -49,7 +49,7 @@ services:
       - dms
     restart: unless-stopped
     healthcheck:
-      test: curl -s http://localhost:8081/health >/dev/null || exit 1
+      test: ["CMD-SHELL", "wget -q --spider http://localhost:${DMS_CONFIG_ASPNETCORE_HTTP_PORTS:-8081}/health || exit 1"]
```
stephenfuqua, line 52: I find it so odd that Alpine images come with `wget` but not `curl`. Was that the problem here? I've used this `wget -q --spider ...` technique somewhere else and found it effective. Probably in Ed-Fi Admin App.

src/dms/backend/EdFi.DataManagementService.Backend.Plans.Tests.Unit/MssqlPlanDialectTests.cs
```diff
@@ -0,0 +1,72 @@
...
+                    d.[DocumentUuid],
+                    d.[ContentVersion],
+                    d.[IdentityVersion],
+                    d.[ContentLastModifiedAt],
+                    d.[IdentityLastModifiedAt]
+                FROM [dms].[Document] d
+                INNER JOIN [#page] k ON d.[DocumentId] = k.[DocumentId]
+                ORDER BY d.[DocumentId];
+
+                """
```
stephenfuqua, line 69: This seems quite sensitive to spacing. Any concerns about that?

simpat-adam (approved):

eng/docker-compose/local-config.yml
bradbanister: Yes, exactly

src/dms/backend/EdFi.DataManagementService.Backend.Plans.Tests.Unit/MssqlPlanDialectTests.cs
bradbanister: Not really. If the spacing changes, agents are quick to identify and fix that kind of minor difference with tests.

PR #1056, opened 2026-06-26
[DMS-1224] Return 415 for unsupported write Content-Type to match ODS/API

samuellugo-ship-it:
## What

Reject baseline data-resource write requests (POST/PUT) that carry an explicit, unsupported `Content-Type` with HTTP `415 Unsupported Media Type` and an `application/problem+json` body, before the request body is parsed.

## Why

DMS previously accepted write requests based on the raw body without validating the request `Content-Type`. A `POST /ed-fi/schools` with `Content-Type: text/plain` and a JSON body would succeed, diverging from legacy ODS/API and the live Ed-Fi v7.3.2 API, which return `415`. This is the `ods-parity` behavior tracked by DMS-1224.

## Changes

**Production (core)**
- Add `ValidateContentTypeMiddleware` (file-scoped namespace) wired into the upsert and update pipelines immediately before `ParseBodyMiddleware`, so an unsupported write `Content-Type` short-circuits with 415 before any JSON parsing.
  - Supported: missing `Content-Type` (not rejected), `application/json` (incl. parameters such as `charset`), and `text/json`.
  - Ed-Fi profile media types (`application/vnd.ed-fi.*`) are deferred to `ProfileResolutionMiddleware`, so existing profile write validation (including its own 415s) is unchanged.
  - Media type is classified with structured `System.Net.Http.Headers.MediaTypeHeaderValue` parsing; an explicit value that is unparseable or resolves to any other media type is rejected.
- Add `FailureResponse.ForUnsupportedMediaType`, producing the DMS-standard ProblemDetails: `status 415`, `title "Unsupported Media Type"`, `type urn:ed-fi:api:unsupported-media-type`, message `The value specified in the 'Content-Type' header is not supported by this host.` in both `detail` and the `errors` array (DMS uses an `errors: []` array plus `validationErrors: {}` rather than the ODS singular `message`).

**Tests**
- Add focused unit tests `ValidateContentTypeMiddlewareTests` (415 for `text/plain` / `application/xml` / unparseable values with pipeline short-circuit; pass-through for `application/json`, `application/json; charset=utf-8`, `text/json`, profile writable, and missing `Content-Type`).
- Add E2E scenario "20 Unsupported write Content-Type is rejected with 415" to `OwaspCriticalPaths.feature`, asserting the 415 status, the response `Content-Type: application/problem+json` on the wire, and the full ProblemDetails body.

**E2E test harness**
- Send data-resource write bodies as raw UTF-8 bytes (`DataByte`) instead of `Data` (string). With a JSON `Content-Type`, Playwright's .NET client re-serializes a non-JSON string body via `JsonSerializer.Serialize`, which broke malformed-JSON validation scenarios; raw bytes reach DMS unchanged.
- Split the header helpers by request shape: `GetHeaders()` / `GetHeadersWithIfMatch()` remain auth-only and are used by GET/DELETE; new `GetWriteHeaders()` / `GetWriteHeadersWithIfMatch()` add `Content-Type: application/json` and are used only by POST/PUT body steps (Playwright defaults bodies with no Content-Type to `application/octet-stream`, which DMS now rejects).
- Content-Type by setup path: in `StepDefinitions`, descriptor setup POSTs use Playwright `DataObject` (which self-serializes to JSON) and stay on the auth-only `GetHeaders()` helper. The profile-feature setup helper (`ProfileStepDefinitions.GivenTheSystemHasTheseEntities`) sets an explicit `application/json` header that it reuses for both its `DataObject` descriptor POSTs and its `DataByte` entity POSTs.

## Testing

- Core builds clean (0 warnings, IDE0055 enforced); focused unit suite passing.
- Relational E2E against a stack built from this branch: `OwaspCriticalPaths` 22/22 (incl. the new 415 scenario), `Resources "Create"` validations 30/30, `Create a Descriptor` validations 21/21, `CorrelationId` 1/1.


samuellugo-ship-it:
## Summary of work done

**Behavior change:** Baseline data-resource write requests (POST/PUT) with an explicit, unsupported `Content-Type` now return HTTP `415` with an `application/problem+json` body, rejected before the body is parsed — matching ODS/API and the live Ed-Fi v7.3.2 API.

### Implementation
- **`ValidateContentTypeMiddleware`** (new, core) sits immediately before `ParseBodyMiddleware` in both the upsert and update pipelines, so `text/plain` + a JSON body returns 415 rather than 400 or 201.
  - Allows missing `Content-Type` (not an explicit value), `application/json` (incl. `; charset=utf-8`), and `text/json`.
  - Defers `application/vnd.ed-fi.*` profile media types to `ProfileResolutionMiddleware` — existing profile write validation (and its own 415s) is untouched.
  - Uses structured `MediaTypeHeaderValue` parsing; unparseable or otherwise-unsupported explicit values are rejected.
- **`FailureResponse.ForUnsupportedMediaType`** returns the DMS-standard ProblemDetails: `status 415`, `title "Unsupported Media Type"`, `type urn:ed-fi:api:unsupported-media-type`, message in both `detail` and `errors[]`. Note DMS uses `errors: []` + `validationErrors: {}` rather than the ODS singular `message`.

### Tests
- 12 new `ValidateContentTypeMiddlewareTests` (415 for `text/plain`/`application/xml`/unparseable with pipeline short-circuit; pass-through for json/+charset/text-json/profile/missing). Focused unit suite: 80/80 passing. Both projects build clean (0 warnings, IDE0055 enforced); csharpier applied.
- New OWASP E2E scenario asserting `text/plain` → 415 + the full ProblemDetails body.
- Shared E2E write helpers updated to send explicit `Content-Type: application/json`, because Playwright defaults string bodies to `application/octet-stream`, which DMS now rejects.

### Verification
Ran the full `OwaspCriticalPaths` feature against a freshly built relational stack: **22/22 passing**, including the new 415 scenario and the unregressed 02/10 (expect 400) and data-setup `Background` scenarios. (The first run produced a false 201 because `build-dms.ps1 E2ETest` reuses a stale `ed-fi-api-local` image; a teardown forces a rebuild from source.)

### Scope notes
- Only POST upsert and PUT update (data resource) pipelines are affected; GET/Accept (406), DELETE, token, change-query, and management/config pipelines are unchanged.
- The pre-existing profile-specific 415 (`...content type in the 'Content-Type' header is not supported...`) is intentionally left as-is; the two 415s coexist for distinct reasons.

samuellugo-ship-it:
## CI fix: malformed-JSON E2E scenarios (relational shards 1 & 2)

The first CI run failed two relational E2E shards. This was a real regression introduced by this branch's test-harness change, now fixed in `dfafc396`.

### Symptom
- `Run PostgreSQL Relational E2E (Shard 1)` — `Resources "Create" Operation validations` › "Post an invalid document (missing a comma)"
- `Run PostgreSQL Relational E2E (Shard 2)` — `Create a Descriptor` › "Post a new descriptor with invalid JSON (trailing comma)"

Both assert the exact `data-validation-failed` ProblemDetails body (including the JSON parse error like `'"' is invalid after a value... LineNumber: 5 | BytePositionInLine: 2`). `AreEqual(expected, actual)` returned `false`. Shards 3 and 4 (including the new 415 scenario) passed.

### Root cause
To make the new 415 check coexist with the suite, the shared E2E write helpers were updated to send `Content-Type: application/json` (Playwright defaults string bodies to `application/octet-stream`, which DMS now rejects with 415). However, Playwright's .NET client (1.46.0), when the `Content-Type` is JSON and the string body is **not** valid JSON, re-serializes it:

```
jsonData = IsJsonParsable(s) ? s : JsonSerializer.Serialize(s);
```

So an invalid JSON string was wrapped/escaped as a JSON string literal, and DMS never received the raw malformed bytes — the malformed-JSON validation scenarios got a different response body, failing their exact-body assertions. (Scenarios that only assert status, like OWASP 10, still passed, which is why this surfaced only on exact-body assertions.)

### Fix (`dfafc396`)
Send write bodies as raw UTF-8 bytes instead of a string:

```diff
- new() { Data = body, Headers = ... }
+ new() { DataByte = System.Text.Encoding.UTF8.GetBytes(body), Headers = ... }
```

Bytes bypass the client's string/JSON handling, so the raw body reaches DMS unchanged (malformed stays malformed) while the `Content-Type` header remains `application/json`. This is:
- **Behavior-preserving** for valid JSON (bytes equal the prior body).
- **Parity-preserving** for production — the middleware still rejects `application/octet-stream` and `text/plain`; no production code changed for this fix.

### Verification
Re-ran both previously-failing features against a relational stack built from this branch:
- `Resources "Create"` validations: 30/30 passed
- `Create a Descriptor` validations: 21/21 passed

(Earlier full OWASP feature run: 22/22, including the new 415 scenario.)

> Note for local repro: `build-dms.ps1 E2ETest` does not rebuild the `ed-fi-api-local` compose image, so a `teardown-local-dms.ps1` is needed first to pick up current code — otherwise the stack runs a stale image.

samuellugo-ship-it:
Thanks — both addressed in `5343e83e`.

**[Major] Assert the wire `Content-Type` for the 415 path.** The OWASP scenario now asserts the response media type in addition to status and body:

```gherkin
Then it should respond with 415
 And the response headers include
     """
     { "Content-Type": "application/problem+json" }
     """
 And the response body is
     """ ... """
```

`the response headers include` does a case-insensitive `Contains` check, which correctly matches the wire value `application/problem+json; charset=utf-8`. This complements the core unit test (which asserts `FrontendResponse.ContentType`) by covering the ASP.NET endpoint response header end-to-end. Verified: OWASP feature 22/22.

**[Minor] Scope the JSON Content-Type to write requests only.** Reverted `GetHeaders()` and `GetHeadersWithIfMatch()` to auth-only (used by GET/DELETE), and added dedicated `GetWriteHeaders()` / `GetWriteHeadersWithIfMatch()` (auth + `Content-Type: application/json`) used only by the POST/PUT body steps. GET/DELETE requests no longer carry a Content-Type. Descriptor setup POSTs use Playwright `DataObject` (which serializes to JSON itself) and stay on the auth-only helper. Verified the custom-header write path (CorrelationId feature 1/1) and reads/deletes via the OWASP run.

samuellugo-ship-it:
### Push: `5343e83e` — review fixes

This push contains one commit addressing the two review comments:

- **E2E (Major):** scenario 20 now asserts the response `Content-Type: application/problem+json` on the wire (via `the response headers include`, case-insensitive `Contains`, matching `application/problem+json; charset=utf-8`), not just status and body.
- **E2E (Minor):** header helpers split by request shape — `GetHeaders()`/`GetHeadersWithIfMatch()` are auth-only (GET/DELETE); new `GetWriteHeaders()`/`GetWriteHeadersWithIfMatch()` add `application/json` and are used only by POST/PUT body steps. GET/DELETE no longer send a Content-Type.

No production code changed in this push (test-harness + feature only).

Local verification against the relational stack: OWASP feature 22/22 (incl. scenario 20's new assertion); CorrelationId 1/1 (custom-header write path).

samuellugo-ship-it:
### Push: `dff01c88` — review fixes (both Minor)

- **File-scoped namespace:** `ValidateContentTypeMiddleware.cs` converted from a block-scoped to a file-scoped namespace, per the repo's modern .NET style. (The file had followed the legacy block style of the sibling `ParseBodyMiddleware.cs`.) No behavior change; Core builds clean (0 warnings, IDE0055) and the 15 `ValidateContentTypeMiddlewareTests` pass.
- **Stale PR description:** the PR body has been updated to match the current implementation — `GetHeaders()`/`GetHeadersWithIfMatch()` are auth-only (GET/DELETE), and the JSON `Content-Type` is added by the new `GetWriteHeaders()`/`GetWriteHeadersWithIfMatch()` used only by POST/PUT body steps. The body also now reflects the `DataByte` raw-body change and the 415 wire `Content-Type` assertion.

No production logic changed in this push (namespace syntax only).

samuellugo-ship-it:
Updated the PR description to correct the descriptor-setup Content-Type wording: in `StepDefinitions` the descriptor `DataObject` POSTs stay on the auth-only `GetHeaders()`, while `ProfileStepDefinitions.GivenTheSystemHasTheseEntities` (lines 295-299) sets an explicit `application/json` header reused for both its `DataObject` descriptor POSTs (308-310) and `DataByte` entity POSTs. Description-only; no code change. Thanks for the catch.

samuellugo-ship-it:
Pushed `fd31dc1` addressing the review note on `ValidateContentTypeMiddleware.cs:67`.

**What changed (production code — Core middleware):**
- `IsSupportedWriteContentType` previously used `string.IsNullOrWhiteSpace(contentType)`, which exempted both a *missing* header and an *explicit* empty/whitespace value. The contract only exempts a missing header.
- Narrowed the exemption to `contentType is null`. An explicit empty or whitespace `Content-Type` now falls through to `MediaTypeHeaderValue.TryParse`, which fails on blank input → **415 returned before any body parsing**, matching ODS/API behavior.
- Updated the doc comment to say "absent (null)" instead of the looser "absent (not an explicit value)".

**Tests:**
- Added `Given_A_Post_Request_With_An_Explicit_Blank_Content_Type`, asserting both `""` and `"   "` halt the pipeline and return 415.
- Existing `Given_A_Post_Request_With_No_Content_Type` (omits the header) still confirms a truly missing header continues.

**Verification:**
- `dotnet build` Core — succeeded, 0 warnings (IDE0055 clean)
- `dotnet test --filter ValidateContentTypeMiddleware` — 17 passed, 0 failed
- `dotnet csharpier check` on both files — clean

samuellugo-ship-it:
Pushed `2b666be` addressing the [Major] review note: explicit blank/whitespace Content-Type was still allowed in production.

**Root cause:** The core middleware correctly rejects an explicit blank Content-Type, but it never received one. `AspNetCoreFrontend.ExtractHeadersFrom` (the lines flagged at `AspNetCoreFrontend.cs:62-70`) dropped any header whose only value was blank/whitespace, so an explicit `Content-Type:` reached core as `null` and hit the missing-header exemption. The earlier unit test passed only because it injected the blank value straight into the headers dictionary, bypassing the adapter.

**What changed (production code — frontend adapter):**
- `ExtractHeadersFrom` now restores a sent Content-Type verbatim when the general filtering would otherwise drop it as blank. Core can now distinguish an explicit blank/whitespace value (→ 415 before body parsing) from a genuinely missing header (exempt).
- Scoped deliberately to Content-Type: all other headers keep their existing drop-when-blank behavior, so Authorization/Tenant/Profile handling is unchanged (verified an empty `Authorization` still yields the same 401).

**Tests:**
- New `AspNetCoreFrontendRequestHeaderTests` (5 tests) exercises `ExtractHeadersFrom` directly: preserves empty and whitespace Content-Type, omits a missing one, keeps a valid one, and confirms other blank headers are still dropped (proving the scoping).
- Existing `ValidateContentTypeMiddlewareTests` continues to assert core rejects blank values with 415. The two layers now prove the full chain: adapter preserves → core rejects.

**Verification:**
- `dotnet build` frontend — succeeded, 0 warnings
- `dotnet test --filter Given_AspNetCoreFrontend_Request_Header_Extraction` — 5 passed, 0 failed
- `dotnet csharpier check` on both files — clean

simpat-adam (approved):

PR #1057, opened 2026-06-26
[DMS-1233] Normalize profile extension names to schema extension keys

samuellugo-ship-it:
## Summary

A profile that authors an extension as `<Extension name="Sample" …>` produced a runtime **500** on relational profile writes (e.g. `PUT /ed-fi/staffs/{id}` with profile `Sample-Staff-Extension-Excluded`). The profile XML preserves the authored casing (`Sample`), but the API schema, relational write plan, and reconstituted documents expose the extension payload under the project endpoint key (`sample`). The write pipeline derived the compiled scope `$._ext.Sample`, which `EffectiveSchemaRequiredMembersBuilder` could not resolve case-sensitively against `jsonSchemaForInsert`, and threw.

This normalizes profile extension names to the schema extension key once, at the profile-load seam, so every downstream profile consumer (scope discovery, navigation, request/stored shaping, read projection, response filtering) agrees on the canonical key.

## Approach

1. **`ProfileExtensionSchemaResolver`** (new) — one shared helper for schema navigation and case-insensitive extension-key resolution (`ExtensionPropertiesAt`, `MemberProperties`, `CollectionItemProperties`, `TryResolveExtensionKey`), used by both the validator and the canonicalizer so they resolve the same key at the same location by construction.
2. **`ProfileExtensionCanonicalizer`** (new) — produces a schema-canonicalized `ProfileDefinition`: rewrites every extension rule name to the canonical schema key, resolved **location-aware** (mirroring the validator), drops rules that do not resolve so no unresolved runtime scope is emitted, and defensively de-dups any sibling rules that resolve to the same key.
3. **`ProfileDataValidator`** — recognizes case-differing extensions, and a single mode-independent pre-pass (`CollectExtensionResolutionFailures`) reports every extension rule that does not resolve at its schema location (so a rule under an `IncludeAll` branch cannot vanish silently). Severity: a non-`IncludeAll` rule (it explicitly selects/deselects members), or any rule under an `IncludeOnly` parent, is an **error**; an `IncludeAll` rule under a tolerant parent is a **warning** (then dropped by canonicalization). Two sibling rules resolving to the **same** schema key are always an error (they would throw in `ExtensionRulesByName`).
4. **`CachedProfileService`** — after validation passes, caches the canonicalized definition.

## Test / acceptance

- Un-ignored `ProfileExtensionFiltering` scenario 09 (Extension Excluded write profile → 204 with persisted `_ext.sample` members).
- Added `Sample-StudentAcademicRecord-Multiple-Extensions-IncludeOnly` to the `DiscoveryAPI` `/metadata/specifications` golden list (it now loads because the validator recognizes the `Sample`→`sample` extension).
- Core unit coverage: case-insensitive resolution (incl. a camelCase `sampleStaff` key, proving the key comes from the schema not lowercasing); location-aware canonicalization at every nesting level (root, object, collection item, and inside a resolved extension''s object/collection); unknown-extension severity by mode (incl. non-`IncludeAll` rule under `IncludeAll` parent → error); resolved same-key duplicate → error; unresolved case-variant names → warnings.

## Verification

- Full Core unit suite passes (see the **Run Unit Tests** CI job for the current result).
- Relational E2E on the current `main` base (through DMS-1229), rebuilt image: `ProfileExtensionFiltering` passes including the un-ignored scenario 09 (→ 204), and `DiscoveryAPI` matches the expected `/metadata/specifications` list. Validator/canonicalizer refinements since are invisible to the seeded profile set, which references no unknown or duplicate extensions.

## Notes

- Scenarios 08/10/11/12 remain 415 — their profiles are dropped by PascalCase **property** errors under `IncludeOnly`, which this change does not touch.

### DIFF
--- a/src/dms/core/EdFi.DataManagementService.Core.Tests.Unit/Profile/CachedProfileServiceTests.cs
+++ b/src/dms/core/EdFi.DataManagementService.Core.Tests.Unit/Profile/CachedProfileServiceTests.cs
@@ -105,15 +105,34 @@ public class CachedProfileServiceTests
                 .Returns(ProfileValidationResult.Success);
         }
 
+        // The service canonicalizes extension names against the schema after validation,
+        // so the provider must expose a non-null ApiSchemaDocuments. These profiles have
+        // no extensions, so canonicalization is a no-op.
+        var schemaProvider = effectiveApiSchemaProvider ?? A.Fake<IEffectiveApiSchemaProvider>();
+        if (effectiveApiSchemaProvider is null)
+        {
+            A.CallTo(() => schemaProvider.Documents).Returns(CreateMinimalApiSchemaDocuments());
+        }
+
         return new CachedProfileService(
             cmsProvider,
             validator,
-            effectiveApiSchemaProvider ?? A.Fake<IEffectiveApiSchemaProvider>(),
+            schemaProvider,
             cache ?? CreateHybridCache(),
             new CacheSettings { ProfileCacheExpirationSeconds = 1800 },
             NullLogger<CachedProfileService>.Instance
         );
     }
+
+    private static ApiSchemaDocuments CreateMinimalApiSchemaDocuments() =>
+        new ApiSchemaBuilder()
+            .WithStartProject()
+            .WithStartResource("Student")
+            .WithEndResource()
+            .WithStartResource("School")
+            .WithEndResource()
+            .WithEndProject()
+            .ToApiSchemaDocuments();
 
     [TestFixture]
     public class Given_No_Profiles_Assigned : CachedProfileServiceTests
--- a/src/dms/core/EdFi.DataManagementService.Core.Tests.Unit/Profile/ProfileDataValidatorTests.cs
+++ b/src/dms/core/EdFi.DataManagementService.Core.Tests.Unit/Profile/ProfileDataValidatorTests.cs
@@ -151,6 +151,74 @@ public class ProfileDataValidatorTests
             return apiSchemaDocuments;
         }
 
+        /// <summary>
+        /// Builds a schema whose insert schema exposes <c>_ext.sample</c> containing a nested
+        /// object (<c>petPreference</c>) and a nested collection (<c>pets</c>), used to exercise
+        /// unknown-extension detection inside an extension's own object/collection children.
+        /// </summary>
+        private static ApiSchemaDocuments CreateSchemaWithExtensionChildren(string resourceName)
+        {
+            var jsonSchemaForInsert = new JsonObject
+            {
+                ["type"] = "object",
+                ["properties"] = new JsonObject
+                {
+                    ["_ext"] = new JsonObject
+                    {
+                        ["type"] = "object",
+                        ["properties"] = new JsonObject
+                        {
+                            ["sample"] = new JsonObject
+                            {
+                                ["type"] = "object",
+                                ["properties"] = new JsonObject
+                                {
+                                    ["petPreference"] = new JsonObject
+                                    {
+                                        ["type"] = "object",
+                                        ["properties"] = new JsonObject
+                                        {
+                                            ["minimumWeight"] = new JsonObject { ["type"] = "number" },
+                                        },
+                                    },
+                                    ["pets"] = new JsonObject
+                                    {
+                                        ["type"] = "array",
+                                        ["items"] = new JsonObject
+                                        {
+                                            ["type"] = "object",
+                                            ["properties"] = new JsonObject
+                                            {
+                                                ["petName"] = new JsonObject { ["type"] = "string" },
+                                            },
+                                        },
+                                    },
+                                },
+                            },
+                        },
+                    },
+                },
+            };
+
+            var apiSchemaDocuments = new ApiSchemaBuilder()
+                .WithStartProject()
+                .WithStartResource(resourceName)
+                .WithEndResource()
+                .WithEndProject()
+                .ToApiSchemaDocuments();
+
+            var resourceNode = apiSchemaDocuments
+                .GetCoreProjectSchema()
+                .FindResourceSchemaNodeByResourceName(new(resourceName));
+
+            if (resourceNode is JsonObject resourceObj)
+            {
+                resourceObj["jsonSchemaForInsert"] = jsonSchemaForInsert;
+            }
+
+            return apiSchemaDocuments;
+        }
+
         private static ApiSchemaDocuments CreateSchemaWithProperties(
             string resourceName,
             params string[] propertyNames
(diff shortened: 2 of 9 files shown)


samuellugo-ship-it:
Pushed the core implementation (commits on `DMS-1233`):

1. **`113c94a8`** — `ProfileDataValidator` resolves extension names case-insensitively against the schema `_ext.properties` keys (preferring an exact ordinal match). A validly-cased-differently extension is no longer flagged; a genuinely unknown extension still fails with the existing severity.
2. **`8deacf8e`** — new `ProfileExtensionCanonicalizer` that rewrites extension rule names to the schema key across Read/Write content types (recursing through objects/collections/extensions) and drops rules with no schema match.
3. **`ab9b6288`** — `CachedProfileService` canonicalizes and caches the canonicalized definition after validation passes.
4. **`4d223ca7`** — un-ignore `ProfileExtensionFiltering` scenario 09; add `Sample-StudentAcademicRecord-Multiple-Extensions-IncludeOnly` to the `DiscoveryAPI` `/metadata/specifications` golden list (it now loads because the validator recognizes the `Sample`→`sample` extension).

**Verification so far:** full Core unit suite green (2712 passed, 1 unrelated skip), including new `ProfileDataValidatorTests` (mixed-case resolution + unknown-extension error/warning) and `ProfileExtensionCanonicalizerTests` (rewrite, drop-unknown, nested-in-collection, read+write, no-op).

**Still pending:** relational E2E confirmation of scenario 09 (→ 204 + persisted `_ext.sample` members) and the updated DiscoveryAPI list (shards 2 and 4). Verifying next.

samuellugo-ship-it:
Relational E2E verified locally (freshly rebuilt `ed-fi-api-local` image from this branch, `.env.e2e.relational`, self-contained IdP):

- **ProfileExtensionFiltering** (`Category=@relational-backend&FullyQualifiedName~ProfileExtensionFiltering`): **12/12 passed, 0 skipped** — scenario 09 now returns 204 and persists the expected `_ext.sample` members; scenarios 08/10/11/12 still return 415; the read scenarios (01–07) still pass.
- **DiscoveryAPI** (`Category=@relational-backend&FullyQualifiedName~DiscoveryAPI`): **4/4 passed** — scenario 03's exact `/metadata/specifications` list matches, confirming `Sample-StudentAcademicRecord-Multiple-Extensions-IncludeOnly` now loads and is listed in the right position.

Combined with the green Core unit suite (2712 passed), all acceptance criteria are covered. Leaving as **draft** — let me know to mark ready for review.

samuellugo-ship-it:
Addressed the blocker in `fd9f4c31`.

**Fix:** `ProfileExtensionCanonicalizer` is now schema-**location-aware**. Instead of a resource-wide union of every `_ext` key, it threads the schema `properties` node through the tree walk and resolves each extension rule against the `_ext.properties` keys **at that rule''s own location** — root, or the specific object/collection/extension the rule sits under — mirroring exactly how `ProfileDataValidator` navigates (`_ext` → `properties` → key). A rule that does not resolve at its own location is dropped.

This closes the failure path you described: a root rule referencing `Sample` when the root has no `_ext.sample` (only a nested collection does) no longer survives, so no unresolved root `$._ext.sample` scope is emitted.

**New test:** `Canonicalize_drops_root_extension_rule_when_extension_only_exists_nested` builds a schema where `_ext.sample` exists only on `staffPets[*]._ext` and asserts a root-level `Sample` rule is dropped.

**No happy-path regression:** for scenarios that resolve at root (scenario 09 on Staff, the StudentAcademicRecord discovery flip), the old union map and the new location-aware code resolve identically (`Sample`→`sample` at root), so the previously-reported relational E2E results (ProfileExtensionFiltering 12/12, DiscoveryAPI 4/4) still hold — the only behavioral change is the nested-only edge case, now covered by the unit test. Full Core unit suite: 2713 passed.

samuellugo-ship-it:
Addressed all three findings in `f14b7633`.

**[Blocker] Unknown IncludeAll extensions no longer vanish silently.** `ValidateNestedMemberSelection` (the IncludeAll-parent path) skipped IncludeAll extension rules entirely, so `<ReadContentType memberSelection="IncludeAll"><Extension name="DoesNotExist" memberSelection="IncludeAll"/></ReadContentType>` got no existence check and was then dropped by canonicalization. It now existence-checks extension rules for every member selection (objects/collections stay as harmless no-ops when unknown, but an unmatched extension is *dropped*, so it needs feedback) → unknown extension is now an **error**. Other parent modes already existence-checked via `ValidateExtensionRule`. New test: `Validate_should_error_for_unknown_extension_under_include_all_parent`.

**[Major] Duplicate keys after canonicalization are rejected with feedback.** A new mode-independent pre-pass rejects sibling extension rules whose names are equal case-insensitively (e.g. `sample` + `Sample`) with a clear **error** at every tree level, before they could collapse to one key and throw in `ExtensionRulesByName`. Two names collapse to the same canonical key **iff** they are case-insensitively equal, so this catches exactly the collisions. New test: `Validate_should_error_for_duplicate_extension_names_differing_only_by_case`.

**[Minor] Canonical key proven to come from the schema, not lowercasing.** Added camelCase-schema-key tests (`sampleStaff`) to both the validator and the canonicalizer: an authored `samplestaff` resolves/rewrites to the schema''s `sampleStaff` (not lowercase). Tests: `Validate_should_resolve_extension_name_to_non_lowercase_schema_key`, `Canonicalize_uses_schema_key_casing_not_lowercase`.

Full Core unit suite: **2717 passed**. These changes only add rejections for authoring mistakes that no seeded profile exhibits, so no profile changes load state — re-running the relational ProfileExtensionFiltering + DiscoveryAPI features now to confirm.

samuellugo-ship-it:
Relational E2E re-verified against a freshly rebuilt image containing the validation changes:

- **ProfileExtensionFiltering**: 12/12 passed — scenario 09 still 204 with persisted `_ext.sample` members; 08/10/11/12 still 415.
- **DiscoveryAPI**: 4/4 passed — the exact `/metadata/specifications` list is unchanged, confirming the new unknown/duplicate-extension rejections do not change any seeded profile''s load state (no regression).

Full Core unit suite remains green (2717 passed). All three review findings are addressed and verified.

samuellugo-ship-it:
Addressed the blocker thoroughly in `c1b061d5` — fixing the whole class of gap, not just the reported spot.

**Root cause:** extension-existence validation was scattered across the member-selection methods, every one of which short-circuits on `IncludeAll` before reaching nested extension rules (`ValidateObjectRuleMembers`, `ValidateCollectionRuleMembers`, `ValidateExtensionRuleMembers`, and the `IncludeAll`-parent branch). Canonicalization, by contrast, walks the entire tree and drops any unmatched extension rule — so an unknown extension under **any** `IncludeAll` object/collection/extension/parent branch disappeared with no feedback.

**Fix:** extension existence is now a single mode-independent pre-pass (`CollectUnknownExtensionFailures`) that walks every extension rule at its own schema location — the *same traversal canonicalization uses* — and reports an error for any that does not resolve. The member-selection path no longer reports existence (`ValidateExtensionRule` only validates the members of a resolved extension), so there is no double-reporting. Wherever canonicalization would drop a rule, the pre-pass now reports it. An unknown extension reference is an **error** regardless of member selection (a tolerant warning would let the dropped rule vanish silently); member-level exclusions stay tolerant.

**Tests added/updated:** nested unknown extension under an `IncludeAll` object → error (the exact reported gap); unknown extension under IncludeOnly/ExcludeOnly/IncludeAll parents → error with the full `_ext` path.

Also removed ticket references from source comments.

**Verified** against a freshly rebuilt image: ProfileExtensionFiltering 12/12 (scenario 09 → 204), DiscoveryAPI 4/4 (exact profile list unchanged → the stricter policy drops no seeded profile). Full Core unit suite: 2718 passed.

samuellugo-ship-it:
Both addressed.

**[Major] Mode-aware severity restored (`82697d6e`).** Unknown extension references now follow the existing missing-reference contract instead of always erroring: severity is derived from the enclosing container''s member selection — `IncludeOnly` → error, `ExcludeOnly`/`IncludeAll` → warning — exactly as missing properties/objects/collections are already treated. A warning plus the canonicalizer dropping the unmatched rule avoids the unresolved runtime scope without turning a previously tolerated `ExcludeOnly` unknown-extension exclusion into a profile-not-supported failure. Feedback is still emitted at every location (including under `IncludeAll` branches, the earlier blocker) via the mode-independent pre-pass — only the severity is governed by member selection. Tests updated accordingly (ExcludeOnly/IncludeAll unknown → warning; IncludeOnly unknown → error; nested-under-IncludeAll → warning).

Since no seeded profile references a non-existent extension, this severity change is invisible to the relational E2E set — the previously reported runs (ProfileExtensionFiltering 12/12, DiscoveryAPI 4/4) remain representative; the only behavioral delta is for profiles that reference an extension that does not exist on the resource.

**[Minor] PR body updated** to reflect the current verified status (Core unit 2718 passed; relational ProfileExtensionFiltering 12/12 and DiscoveryAPI 4/4) instead of "pending".

samuellugo-ship-it:
Both closed in `eb8bf318`.

**[Major] Coverage now matches the "every nesting level" claim.** Added focused tests for the two traversal branches that were untested:
- `Validate_should_report_unknown_extension_under_collection_item` — unknown extension on a collection item (`studentPets[]._ext.DoesNotExist`), exercising the collection path.
- `Validate_should_report_unknown_extension_inside_resolved_extension_object` — unknown extension inside an object nested within a resolved extension (`_ext.sample.petPreference._ext.DoesNotExist`).
- `Validate_should_report_unknown_extension_inside_resolved_extension_collection` — same for a collection within a resolved extension (`_ext.sample.pets[]._ext.DoesNotExist`).

Combined with the existing root (IncludeOnly/ExcludeOnly/IncludeAll) and nested-object cases, every traversal branch of the unknown-extension pre-pass is now covered, and each asserts the full path and the mode-derived severity. ProfileDataValidatorTests: 77 passed; full Core unit suite: 2721 passed.

**[Minor] Stale comment fixed.** `CachedProfileService` no longer says unknown extensions are "rejected during validation"; it now states that IncludeOnly unknown references are errors (profile already dropped), while ExcludeOnly/IncludeAll unknown references are warnings whose rules the canonicalizer drops.

samuellugo-ship-it:
Both addressed in `d6028e5e`.

**[Major] Duplicates are now keyed by resolved schema key, not raw name.** I merged duplicate detection into the schema-aware extension walk. Each rule is resolved to its canonical schema key first:
- An **unresolved** name takes the missing-reference severity (IncludeOnly → error, ExcludeOnly/IncludeAll → warning) and is **never** treated as a duplicate — so two unresolved case-variants like `Foo` and `foo` under ExcludeOnly are two warnings (then dropped by canonicalization), not a rejection.
- Two rules that resolve to the **same** schema key (e.g. `sample` + `Sample`) are a genuine collision — both survive canonicalization and `ExtensionRulesByName` would throw — so that remains an error.

New test `Validate_should_warn_not_error_for_unresolved_case_variant_extensions_under_exclude_only` covers the reported case (two warnings, no duplicate error); the existing same-key duplicate test still asserts the error.

**[Minor] PR body verification updated** to the current Core unit count (2722 passed) and the relational E2E results, with a note that the validation-only refinements since the last E2E run are invisible to the seeded profile set (it references no unknown or duplicate extensions).

samuellugo-ship-it:
Rebased onto latest `main` (now atop `[DMS-1224] Return 415 for unsupported write Content-Type`, #1056) — 10 commits replayed cleanly with no conflicts. Verified the rebase against a freshly rebuilt image: that new `ValidateContentTypeMiddleware` defers all `application/vnd.ed-fi.*` media types to profile resolution, so it does not affect these scenarios.

- Core unit suite: 2739 passed.
- Relational E2E (rebuilt image): `ProfileExtensionFiltering` 12/12 (scenario 09 → 204), `DiscoveryAPI` 4/4.

Marked ready for review.

[DMS-1224]: https://edfi.atlassian.net/browse/DMS-1224?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

samuellugo-ship-it:
Re-rebased onto latest `main` (now atop `Nuget docker files packages` #1058) to clear the BEHIND state — 10 commits replayed cleanly, 0 behind. That base commit only changes two `Nuget.Dockerfile` lines (no C#/runtime/test impact), so no rebuild/re-verification was needed; the prior unit + relational E2E results stand. Branch head is now `c4994250`.

samuellugo-ship-it:
Rebased again onto latest `main` — now atop `[DMS-1229] ODS-compatible profile write handling for out-of-profile submitted data` (#1055), which reworks the profile write path. 10 commits replayed cleanly (no conflicts).

Because DMS-1229 changes write-path runtime behavior, I re-verified against a freshly rebuilt image:
- Core unit suite: 2800 passed.
- Relational E2E: `ProfileExtensionFiltering` 12/12 (scenario 09 → 204 under the new write handling), `DiscoveryAPI` 4/4 (golden list still correct — DMS-1229 added no seeded profiles).

Head is now `77621439`, 0 behind `main`. The earlier "Detect Fresh Build Changes" failure was a git merge-base/plumbing error in that detect job (exit 128) caused by the PR being BEHIND, unrelated to this change; this rebase clears the BEHIND state, so the fresh run runs the detect step against a current merge-base.

[DMS-1229]: https://edfi.atlassian.net/browse/DMS-1229?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

samuellugo-ship-it:
Addressed all four findings in `dc1288ce`.

**[High — severity regression]** Unknown-extension severity no longer keys only on the parent. An explicit **non-IncludeAll** extension rule selects/deselects specific members, so the extension must exist — such a rule that names a missing extension is now an **error regardless of parent**. Concretely, `<Extension name="Smaple" memberSelection="IncludeOnly">` under an `IncludeAll` content type is now rejected (error) instead of warned‑and‑silently‑dropped, preserving the explicit restriction. `IncludeAll` rules under a tolerant parent (ExcludeOnly/IncludeAll) remain warnings; any rule under an `IncludeOnly` parent stays an error. New test: `Validate_should_error_for_unknown_include_only_extension_under_include_all_parent`. (All prior severity tests still hold.)

**[Medium — parallel walks drift]** Extracted the duplicated schema navigation + key resolution into one `ProfileExtensionSchemaResolver` (`ExtensionPropertiesAt`, `MemberProperties`, `CollectionItemProperties`, `TryResolveExtensionKey`); both the validator and the canonicalizer now call it via `using static`. The "same key at the same location" invariant holds by construction rather than copy‑paste discipline.

**[Low — defensive dedup]** `CanonicalizeExtensions` now drops a later sibling rule that resolves to an already-emitted schema key, so it can never emit a colliding `ExtensionRulesByName` entry even if invoked without validation.

**[Low — canonicalizer nesting tests]** Added `Canonicalize_rewrites_extension_nested_in_object` and `Canonicalize_rewrites_extension_inside_resolved_extension_object` (`_ext.sample.obj._ext.other`), mirroring the collection test so the shared resolver is covered at every nesting level.

Core unit suite: **2803 passed**. These are validator/canonicalizer refinements invisible to the seeded profile set (which references no unknown or duplicate extensions), so the prior relational E2E results (ProfileExtensionFiltering 12/12, DiscoveryAPI 4/4 on the DMS‑1229 base) stand; CI runs the full relational shards as backstop.

samuellugo-ship-it:
Fixed in `f7de1825` — updated the stale comment in `ValidateNestedMemberSelection` to reference `CollectExtensionResolutionFailures` (the renamed helper). I also swept the rest of the diff for other pre-rename references in comments (`MemberSchemaProperties`, `ResolveExtensionSchemaNode`, `SeverityForMissingReference`, the old walk method names) and found none. Comment-only change.

samuellugo-ship-it:
Updated the PR body Verification section to the current Core unit count (2803 passed) and refreshed the Approach section to match the final shared-resolver design and unknown-extension severity rule (non-IncludeAll rule, or any rule under an IncludeOnly parent, is an error; IncludeAll under a tolerant parent is a warning).

samuellugo-ship-it:
Good call — removed the hardcoded unit-test total from the Verification section. It went stale on every commit that added a test (hence the repeated findings) and the CI "Run Unit Tests" job is the always-current source of truth, so the body now points there and describes the suites qualitatively instead.

samuellugo-ship-it:
Both addressed in `43f93174`.

- **[Major]** Added `Canonicalize_rewrites_extension_inside_resolved_extension_collection` (`_ext.sample.pets[]._ext.other`), mirroring the object-branch test so both recursion paths the canonicalizer takes inside a resolved extension are now covered — the PR body claim ("object/collection") is backed by tests.
- **[Minor]** Updated the `CollectExtensionResolutionFailures` call-site comment to state the current severity rule (non-IncludeAll rule, or any rule under an IncludeOnly parent → error; IncludeAll under a tolerant parent → warning).

Full Core suite passes; no behavior change (a test + a comment).

samuellugo-ship-it:
Confirmed and fixed in `2b3672fa` — this one was a real regression in the extension pre-pass, thanks for catching it.

**Cause:** the pre-pass walked into an object/collection''s extensions even when the parent node did not resolve in the schema (`MemberProperties`/`CollectionItemProperties` returned null). `CheckExtensions` then saw a null `_ext` and reported every nested extension as missing, escalating a non-`IncludeAll` rule to an error — so an `ExcludeOnly` profile naming a non-existent object that contains an `IncludeOnly` extension was rejected, where it previously loaded (the missing object is reported, as a warning, by member-selection validation, which descends only into a resolved node).

**Fix:** stop descending when the parent object/collection does not resolve, mirroring member-selection validation. Loadability for those out-of-scope malformed profiles is unchanged; the canonicalizer still drops the orphaned extension, so no unresolved runtime scope is emitted (no 500). Existing nested-extension tests still pass because they reference objects/collections that *do* resolve (just without that `_ext` key).

**Tests:** added an `ExcludeOnly` profile with an `IncludeOnly` extension under a missing object, and the same under a missing collection — both now load with a warning and no escalated error.

Full Core suite passes; no seeded profile exhibits this shape, so scenario 09 / DiscoveryAPI are unaffected (CI relational shards run as backstop).

JBrenesSimpat (approved):

PR #1070, opened 2026-07-01
[DMS-1250] Support running DMS from Visual Studio

axelmarquezh:


src/dms/Directory.Build.targets
```diff
@@ -1,4 +1,10 @@
 <Project>
+    <PropertyGroup>
+        <_DmsNetCoreSdkRoot Condition="'$(NETCoreSdkBundledVersionsProps)' != ''">$([System.IO.Path]::GetFullPath('$(NETCoreSdkBundledVersionsProps)/..'))</_DmsNetCoreSdkRoot>
+        <_DmsNetCoreSdkRoot Condition="'$(_DmsNetCoreSdkRoot)' == '' and '$(MSBuildRuntimeType)' == 'Core'">$(MSBuildToolsPath)</_DmsNetCoreSdkRoot>
+        <_DmsNewtonsoftJsonPath>$(_DmsNetCoreSdkRoot)/Sdks/Microsoft.NET.Sdk/tools/net472/Newtonsoft.Json.dll</_DmsNewtonsoftJsonPath>
```
bradbanister: Disappointing to see Newtonsoft back in DMS. Is that the whole reason we don't support Visual Studio? If so, we should get rid of Newtonsoft (again). And add to the AGENTS.md to use System.Text.Json instead.

stephenfuqua: +1

src/dms/Directory.Build.targets
bradbanister: This is the actual problem. This embedded C# code should not be using Newtonsoft.Json

bradbanister (changes requested):
See below

src/dms/Directory.Build.targets
axelmarquezh: This PR removes `Newtonsoft.Json` in favor of `System.Text.Json`. But `RoslynCodeTaskFactory` doesn't resolve `System.Text.Json` consistently between dotnet MSBuild and Visual Studio MSBuild, so we still need equivalent explicit assembly-reference handling.

I also updated AGENTS.md to use `System.Text.Json` and prohibit new `Newtonsoft.Json` dependencies or usages.

IMO we should replace this Inline Task with a .NET 10 MSBuild Task so that these packages are visible/auditable by `dotnet list package --outdated` and `--vulnerable`, see [DMS-1307](https://edfi.atlassian.net/browse/DMS-1307).

src/dms/Directory.Build.targets
axelmarquezh: See my comment above.

JBrenesSimpat (approved):

bradbanister (approved):

PR #1083, opened 2026-07-02
[DMS-1238] Add local Docker deployment support for the MSSQL backend

simpat-jesus:
## Summary

Adds turnkey local Docker deployment for the DMS **MSSQL backend**, bringing the SQL Server experience to parity with PostgreSQL — one command, no environment-file bookkeeping: `./bootstrap-local-dms.ps1 -DatabaseEngine mssql` stands up a **single-engine SQL Server stack** (DMS datastore, Configuration Service, and self-contained OpenIddict identity — building on the CMS SQL Server backend from DMS-1243/#1084), provisions the relational schema, and serves a full CRUD cycle. No PostgreSQL container runs on the mssql path.

## What's included

- **`-DatabaseEngine mssql`: single-engine swap** — `mssql.yml` (extended from #1084's CMS swap-in file with the Developer-edition pin, a memory limit, and a hardened healthcheck) swaps in for `postgresql.yml`: both define the same `db` service that `local-config.yml` health-gates on, so SQL Server hosts the DMS datastore, the CMS, and the OpenIddict identity stores, exactly mirroring `start-local-config.ps1`'s engine selection. Kafka/Debezium CDC correctly omitted on the relational path. The engine flag auto-composes the **`.env.mssql` overlay** (`DMS_DATASTORE=mssql`, `DMS_CONFIG_DATASTORE=mssql`, the `MSSQL_*` keys, and the SQL Server connection strings) onto the base env via the same derived-file machinery as `-DataStandardVersion` — and every entry point and phase command (`start-local-dms.ps1`, `configure-local-data-store.ps1`, `provision-dms-schema.ps1`) accepts `-DatabaseEngine` and composes the same overlay, so the turnkey and the documented manual phase flow both see one consistent engine selection and a split-backend misconfiguration is structurally impossible. All `setup-openiddict.ps1` invocations pass engine-aware database parameters (mirroring the CMS stack script). `provision-dms-schema.ps1` resolves the dialect host-first (`host` ⇒ PostgreSQL before any SQL Server marker, so Npgsql's `User Id` alias can't misroute).
- **DMS image fixes required for MSSQL** — publish `runtimes/` into the runtime image (Microsoft.Data.SqlClient resolves its Linux implementation from `runtimes/unix/`), install `icu-libs` and disable Globalization Invariant Mode, and add the `mssql` provider to `DbHealthCheck`.
- **Schema-consistency fix (engine-agnostic — also changes the PostgreSQL and published bootstrap paths)** — standard-mode `prepare-dms-schema.ps1` now stages the exact `SCHEMA_PACKAGES` set from the effective env file instead of the catalog-pinned core-only default, so the provisioned database's schema hash always matches what the DMS container resolves at startup (previously any multi-package env — including the stock `.env.example` — produced a 503 hash mismatch on every write, for both engines). A `SCHEMA_PACKAGES` set that doesn't list exactly one DS-qualified core package (`EdFi.DataStandard<NN>.ApiSchema`) now fails fast at the prepare phase with an actionable error, rather than silently staging a core-only workspace the runtime would contradict.
- **TPDM enabled for DS 5.2** — TPDM registered as a known embedded-claims extension (claims ship in the CMS base `Claims.json`); core-package detection is data-standard-agnostic.
- **`-DataStandardVersion` on the bootstrap entry points** — `ValidateSet("5.2","6.1")`, default `5.2`. The local entry point always composes the bootstrap overlay (`.env.bootstrap.ds52` core + TPDM / `.env.bootstrap.ds61` core only); the published entry point composes it **only when the parameter is explicitly supplied**, preserving pre-existing custom-base published workflows. These local-bootstrap overlays remain distinct from the shared `.env.ds52`/`.env.ds61` E2E/SDK surfaces.
- **Review hardening** — SeedLoader claim-set check fixed to the actual CMS response shape (`claimSetName`); `-EnableKafkaUI` skips cleanly on MSSQL instead of dying on a missing compose service; `MSSQL_SA_PASSWORD` defaults consistently across all three consumers; PSScriptAnalyzer suppressions retained only at genuinely plaintext boundaries (connection-string factory, `sqlcmd -P`) after adopting main's PSCredential refactor for the CMS data-store helpers. Provisioning fail-fasts when a CMS data store's dialect contradicts the environment's `DMS_DATASTORE` (e.g. a stale store reused via `-NoDataStore`), with the teardown + re-bootstrap recovery spelled out in the error; `-DataStoreDatabaseName` is honored on the MSSQL path; and the previously-orphaned spec files (engine overlay, deployment safety, container packaging) now run in the PR CI lane. The container-internal token endpoints moved from the MSSQL overlay into the base `.env.example` template — the DMS container's own `/oauth/token` proxy runs inside the Docker network regardless of engine, so the localhost template values broke the PostgreSQL turnkey path too; `start-local-dms.ps1` also adopted the shared local-settings resolution, so direct invocations (including teardown) work on a clean checkout with no hand-created `.env`.
- **`.env.example` is an example again** — the shared resolver (all entry points and phase commands) now **seeds `.env` once as a copy of `.env.example`** when `.env` is absent and always consumes `.env`; the tracked example is documentation only and is never read at runtime. First run on a clean checkout still needs no manual step, and user edits get a durable home. The default `SCHEMA_PACKAGES` set in the example was trimmed to **core + TPDM** (matching the DS 5.2 bootstrap overlay); Sample/Homograph remain fully supported — the E2E/SDK surfaces keep them in the shared `.env.ds52`/`.env.ds61` overlays, and any extension ApiSchema package can be added to `.env`. Provisioning log lines now render SQL Server `host,port` targets readably (comma whitelisted in the log sanitizer; control characters stay stripped).
- **Microsoft.OpenApi 2.0.0 → 2.7.5** — clears NU1903 (GHSA-v5pm-xwqc-g5wc) under `TreatWarningsAsErrors`.
- **Version-pin hygiene** — catalog core pin aligned to `1.0.332`; a Pester spec asserts all tracked env files and the catalog agree on a single ApiSchema package version.
- **README** — one-command MSSQL deployment, engine-overlay semantics, and Data Standard selection documented.

## Verification

- Pester: full bootstrap suite green — **580 passed / 0 failed** — including new specs for the engine compose-file swap, engine-aware identity parameters, the single-engine overlay contract, engine-overlay composition (wrapper + direct paths, custom bases, idempotency), per-entry-point data-standard gating, dialect resolution + engine-mismatch guarding, SCHEMA_PACKAGES-driven staging, `.env` seeding from the example, and version consistency.
- **Single-engine MSSQL stack verified live end-to-end** (post-rebase onto #1084): from a fully clean Docker state (stale images removed, no hand-created `.env` — the run exercised the first-run `.env` seeding), a single `./bootstrap-local-dms.ps1 -DatabaseEngine mssql -LoadSeedData` rebuilt the images and ran staging → SQL Server → OpenIddict identity init (database + `dmscs` schema self-provisioned on SQL Server) → CMS healthy (startup deploy of `edfi_configurationservice` on SQL Server) → claims gate → data-store registration → `--dialect mssql` provisioning → healthy DMS → built-in Minimal seed load. Verified in SQL Server: `edfi_configurationservice` + `edfi_datamanagementservice` both on the single `dms-mssql` instance, OpenIddict key/application rows present, row counts 3,349 Documents / 3,302 Descriptors / 3,349 ReferentialIdentities / 474 ResourceKeys; `/health` fully Healthy (DbHealthCheck on mssql); **no PostgreSQL container exists**. Env-less teardown (`start-local-dms.ps1 -DatabaseEngine mssql -d -v`) verified. One transient: the first BulkLoadClient attempt died in a known client-side log4net thread-context race (`Stack empty` in `EdFi.LoadTools.LogContext`) after 1,120 clean 201s; the phase rerun (`load-dms-seed-data.ps1`) completed with exit 0 — not a DMS/SQL Server failure. An automated live-Docker MSSQL CI lane is tracked as a follow-up ticket.
- Data Standard selection was live-verified on the earlier dual-database iteration of this branch (pre-#1084 rebase): DS 5.2 discovery advertised Ed-Fi 5.2.0 + TPDM 1.1.0 with full descriptor CRUD (`201/200/204/404`) and TPDM probes green; DS 6.1 advertised Ed-Fi 6.1.0 only, same cycle green. The DMS write path those runs exercised (relational backend on SQL Server) is unchanged by the single-engine redesign; the DS 6.1 and explicit CRUD probes were not separately re-run on it.
- PostgreSQL regression: clean bootstrap provisions `--dialect pgsql`, startup validation passes with matching hash.
- Both solutions build clean; DMS + CMS Docker images build with no NuGet audit failures.


JBrenesSimpat (approved):

simpat-adam (approved):

src/dms/Dockerfile
stephenfuqua, line 48: @simpat-jesus although the line above is a holdover from prior work, can you resolve it in this pull request?

simpat-jesus: Resolved in 5a9c9b957 — that RUN block was tripping ShellCheck SC2086 (unquoted `$version_args` word-splitting, a holdover from the original version-stamping work). The version-stamp arguments are now accumulated as positional parameters and expanded with a quoted `"$@"`. Also pinned this PR's `icu-libs` addition to `=~76` (Alpine 3.23 ships icu-libs-76.1-r1), so hadolint is now fully clean on this Dockerfile. Verified with a `--build-arg VERSION` build of the publish stage.

simpat-adam (approved):

PR #1087, opened 2026-07-03
[DMS-1227]    Return 413 for oversized DMS request bodies and add OWASP E2E coverage

jleiva-gap:


src/dms/frontend/EdFi.DataManagementService.Frontend.AspNetCore/Program.cs
```diff
@@ -70,16 +71,24 @@
...
 
-        // Configure request size limits for schema upload
-        builder.Services.Configure<Microsoft.AspNetCore.Http.Features.FormOptions>(options =>
+        // Configure request size limits for schema upload using the configured request-body limit.
+        int maxRequestBodySizeMegabytes = int.TryParse(
+            builder.Configuration["AppSettings:MaxRequestBodySizeMegabytes"],
+            out int configuredMaxRequestBodySizeMegabytes
+        )
+            ? configuredMaxRequestBodySizeMegabytes
+            : AppSettings.DefaultMaxRequestBodySizeMegabytes;
```
axelmarquezh, line 80: Minor: src/dms/frontend/EdFi.DataManagementService.Frontend.AspNetCore/Configuration/AppSettings.cs:17 leaves MaxRequestBodySizeMegabytes at 0 when omitted, while src/dms/frontend/EdFi.DataManagementService.Frontend.AspNetCore/Program.cs:75 has a separate fallback default. Omitted config can fail validation instead of using the default. Add a property initializer or centralize through validated options.

Detailed explanation:
```
  MaxRequestBodySizeMegabytes is an int, so if configuration binding does not supply a value, it stays at C#’s default value: 0.

  Current split behavior:

  - src/dms/frontend/EdFi.DataManagementService.Frontend.AspNetCore/Program.cs:75 manually reads AppSettings:MaxRequestBodySizeMegabytes and falls back to 10 if parsing fails or the key is missing.
  - src/dms/frontend/EdFi.DataManagementService.Frontend.AspNetCore/Configuration/AppSettings.cs:17 has no initializer, so bound IOptions<AppSettings> gets 0 when the key is missing.
  - src/dms/frontend/EdFi.DataManagementService.Frontend.AspNetCore/Configuration/AppSettings.cs:65 rejects <= 0.

  So the app can simultaneously have Kestrel configured with the intended default 10 MB, while AppSettingsValidator treats the same omitted setting as invalid. That is the inconsistency.

  Today the default appsettings.json includes:

  "MaxRequestBodySizeMegabytes": 10

  so the normal path is probably fine. The issue appears if a test, deployment, alternate config file, or future cleanup omits the key. Then the manual Kestrel fallback says “use 10,” but options validation says “invalid 0.”

  The small fix is:

  public int MaxRequestBodySizeMegabytes { get; set; } = DefaultMaxRequestBodySizeMegabytes;

  That makes the bound options model and the Program fallback agree. A cleaner follow-up would be to avoid manually parsing config in Program.cs and instead configure host limits from one validated AppSettings source, but the initializer is the least invasive correction.
```

axelmarquezh (changes requested):

src/dms/frontend/EdFi.DataManagementService.Frontend.AspNetCore/Program.cs
jleiva-gap: Done

axelmarquezh (approved):

PR #1106, opened 2026-07-12
[DMS-1196] Focused review hardening: gateway host pinning, credential-revoking reset, portproxy reboot fix

simpat-adam:
## What this is

A focused hardening pass on the `eng/azure-vm/` security-review deployment, proposed as the final delta to merge into `DMS-1196`. It replaces an earlier 12-commit review-fix series on this branch: each of those commits was re-evaluated, the changes that fix real defects were kept, and the exploratory validation machinery was deliberately dropped as out of scope for a review environment (details below, so nothing was lost silently). Net: **15 files, +181/−170**.

With these changes, review passes on the branch come back clean — this is the set we believe clears DMS-1196 for merge.

### Gateway: stop trusting the client's Host header (`compose/nginx/default.conf.template`)
- `Host` and `X-Forwarded-Host` sent to the backends are now pinned to `${PUBLIC_HOST}` instead of echoing `$host`. The apps trust the gateway's `X-Forwarded-*` headers, so a spoofed Host previously flowed into redirect/metadata generation — a textbook host-header-injection vector on an Internet-exposed environment.
- A `default_server` TLS vhost returns **421** for any unmatched SNI/Host, so nothing with an attacker-chosen hostname reaches the applications. The named vhost also accepts `localhost` for the documented VM-loopback bootstrap/health path.
- Rendered through the official image's `envsubst` and verified with `nginx -t`.

### Reset now revokes credentials (`reset.sh`, `down.sh`, `bootstrap.ps1`, docs)
- `reset.sh` passes `keycloak.yml` to `down -v`, dropping the realm together with the CMS/database volumes. Review applications are Keycloak clients: keeping the realm while rebuilding CMS left **old API credentials valid against reassigned data-store IDs**. Prompts, printed rebuild steps, and the runbooks (TROUBLESHOOTING, MANUAL, provision/README) are updated to match, including a health wait-loop before re-bootstrap so the `attempted` marker isn't burned against a not-yet-healthy CMS.
- `bootstrap.ps1 -KeycloakOnly` is removed: rebuilding only the realm's service clients cannot restore the review-application clients whose secrets/UUIDs CMS references, so the flag was a broken-credentials trap. Clean redeploy (`REDEPLOY.md`) is the documented recovery for Keycloak volume loss or image-pin changes.
- `down.sh`/`reset.sh` clear the bootstrap markers **before** the destructive down: an interrupted teardown can no longer leave a stale `complete` marker that makes `setup-env.ps1` skip bootstrap against wiped databases.

### Windows host: port forwarding survived reboots only by luck (`provision/windows/portproxy.ps1`)
The startup task invokes the script via `powershell.exe -File`, which passes `-Ports 80,443` as a **single string** — the `[int[]]` parameter fails to bind and the boot task dies before restoring connectivity, so the environment went dark after every VM reboot. The parameter is now `[string[]]` with comma-token flattening; both the interactive and `-File` invocation forms bind.

### Smaller fixes
- `bootstrap.ps1` validates MT tenant names (charset, 19-char cap from the generated API-client name, case-insensitive distinctness) before any identity/CMS mutation.
- `grandbend.sh` legacy-template guard now requires an `edfi` DDL marker **and** the `EffectiveSchema` marker; the OR-regex accepted a document-store dump that merely mentioned one.
- `http/single-tenant.http` creates school `255901999` instead of `255901001`, which collides with Grand Bend seed data.
- `MANUAL.md` resolves the NSG via the VM's NIC instead of "first NSG in the resource group".
- `windows/README.md` scopes the RDP inbound rule to admin CIDRs by default instead of Any-with-a-footnote.

### Tests: runtime-only (`eng/docker-compose/tests/AzureVmDeployment.Tests.ps1`)
The "deployment invariants" block string-matched script source, which our test policy disallows — it's gone. The lifecycle suite still stubs `docker` and actually executes `down.sh`/`reset.sh`, with new cases for the marker-ordering fix, the Keycloak-inclusive reset, and the non-interactive reset refusal.

## Verification

- Pester: 7/7 green (runtime suite above)
- PSScriptAnalyzer: clean on all changed scripts (also enforced by the pre-commit hook)
- `bash -n` on all changed shell scripts; PowerShell AST parse on all changed `.ps1`
- nginx template rendered via `envsubst` and validated with `nginx -t`
- Not re-deployed to a live Azure VM as part of this pass

### DIFF
--- a/eng/azure-vm/TROUBLESHOOTING.md
+++ b/eng/azure-vm/TROUBLESHOOTING.md
@@ -44,14 +44,19 @@ docker exec -it dms-sec-postgres psql -U postgres -l    # list databases
 | Bulk-load **`invalid_client` / 401** with a fresh key | `DMS-1231`, fixed in `:pre` ≥ 2026-06-24 (#1047) — generated secrets are now Basic-safe. On an older image the secret may contain `+`/`%`; re-mint until `+`/`%`/space-free, or URL-encode it in the Basic header. |
 | Bulk-load **429 / "circuit is now open"** mid-run | Rate limiter + circuit breaker (`FAILURE_RATIO=0.01`) tripped by parallel load. Load descriptors first, raise the rate limit, then resources; or use `seed/clone-data.sh`. |
 
-## Reset (keep the Keycloak realm)
+## Reset deployment state
 
-A `-v` reset empties the data DBs, so — like a first stand-up — bootstrap, schema provisioning,
-and the DMS start all have to run again (`reset.sh` prints these same steps on exit):
+A reset empties the data DBs and the Keycloak realm, so old review credentials are revoked. Like a
+first stand-up, bootstrap, schema provisioning, and the DMS start all have to run again
+(`reset.sh` prints these same steps on exit):
 
 ```bash
 ./reset.sh
-pwsh ./bootstrap/bootstrap.ps1 -SkipKeycloak -BaseUrl https://localhost -Insecure
+# wait for the freshly recreated identity/CMS services before bootstrap writes its attempted marker:
+until curl -skf https://localhost/auth/realms/master >/dev/null \
+  && curl -skf https://localhost/st-config/health >/dev/null \
+  && curl -skf https://localhost/mt-config/health >/dev/null; do sleep 5; done
+pwsh ./bootstrap/bootstrap.ps1 -BaseUrl https://localhost -Insecure
 # provision the relational schema into edfi_st / edfi_mt / edfi_mt_t2
 # (api-schema-tools, or restore the populated template; see ../docs/infrastructure.md), then:
 ./up.sh st-dms mt-dms
--- a/eng/azure-vm/compose/.env.example
+++ b/eng/azure-vm/compose/.env.example
@@ -6,7 +6,9 @@
 # Generate strong secrets, e.g.:  openssl rand -base64 32
 #
 # Client secrets must be 32-128 chars with at least one lowercase, uppercase,
-# digit, and special character (! @ # $ % ^ & * ( ) - _ = + [ ] { } : ; , . ?).
+# digit, and special character (! @ # ^ * ( ) - _ = + [ ] { } : ; , . ?).
+# Never use '$' in any value here: docker compose interpolates $VAR/${VAR} inside .env, so the
+# containers would receive a DIFFERENT value than the deployment scripts read (bootstrap refuses).
 
 # ---------------------------------------------------------------------------
 # Public address (how the review team reaches the gateway)
--- a/eng/azure-vm/compose/bootstrap/bootstrap.ps1
+++ b/eng/azure-vm/compose/bootstrap/bootstrap.ps1
@@ -18,7 +18,6 @@
 #   pwsh ./bootstrap.ps1                 # uses ../.env
 #   pwsh ./bootstrap.ps1 -Insecure       # local self-signed cert
 #   pwsh ./bootstrap.ps1 -SkipKeycloak   # realm/clients already created
-#   pwsh ./bootstrap.ps1 -KeycloakOnly   # rebuild realm/clients after wiping only Keycloak H2
 [CmdletBinding()]
 [Diagnostics.CodeAnalysis.SuppressMessageAttribute('PSReviewUnusedParameter', 'ClaimSetName', Justification = 'Consumed as the default value of the New-ReviewApplication -ClaimSet parameter; the analyzer does not track usage inside nested function parameter defaults.')]
 param(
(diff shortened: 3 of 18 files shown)


simpat-adam:
## Follow-up: nine external-review findings validated and fixed (`ba2ca28d`)

A second review pass surfaced nine findings. Each was verified against the code before fixing — three by live experiment — and all nine had a real core. Fixes are minimal-form, +127/−11 across 8 files:

| Finding | Verdict | Fix |
|---|---|---|
| REDEPLOY stages core-only ApiSchema vs core+TPDM template | **Real (P1).** `.env.template` confirms the populated template is built from the DS 5.2 core+TPDM surface; the bare `prepare-dms-schema.ps1` call stages the core-only default → every template-restored DB fails the EffectiveSchema startup check | Part C now passes `-EnvironmentFile ./.env.template` |
| `down.sh` misses bundled flags (`-vt 0`) | **Real.** Verified live: docker/pflag parses bundled short flags, wrapper's `case` didn't | Any single-dash cluster containing `v` is treated as a volume drop; long flags unaffected |
| REDEPLOY Part A `\|\| true` then `rm -f .env` | **Real** (severity tempered by the vault mandate) | Existence-guarded `down.sh` that fails loudly + both state volumes verified absent before `.env` is deleted |
| `update.sh` auto-applies a Keycloak pin change over H2 | **Real.** Nothing enforced the clean-redeploy requirement documented in `keycloak.yml` | Guard compares configured vs running image *reference* and aborts with redeploy instructions (same-tag repulls don't trip it) |
| Markers cleared before deletion is confirmed | **Real — a regression in `b77d16d3`.** With markers gone and volumes intact after a failed down, a re-run would duplicate live identity/CMS objects (`Add-DataStore`/`Vendor`/`Application` are not idempotent) | Minimal `reset-pending` sentinel: written before the destructive down, removed only on success; `bootstrap.ps1` refuses while it exists. Tests updated to pin both directions |
| Compose interpolates `$` in `.env` values | **Real but narrow.** Verified live (`ab$PLAIN` → `abhello` in the container). Generated secrets can't contain `$`; the `.env.example` charset line invited it | `$` removed from the documented charset; `EnvVal` refuses any value containing `$` |
| `-LoadGrandbend` next-steps say to provision `edfi_st` | **Real doc bug** — contradicts `infrastructure.md` | Conditional next-steps: MT databases only |
| ST claim set hardcoded, `-ClaimSetName` MT-only | **Real, minor.** Hardcode equalled the parameter default, so default behavior unchanged | Hardcode dropped; the parameter now applies to all three apps |
| Detached checkout vs `update.sh` `git pull --ff-only` | **Real, mildest** — update.sh's error already names the workaround | Part B note: `SKIP_GIT=1 ./update.sh` or repeat Part B |

Verification: Pester 10/10 (3 new runtime cases + sentinel assertions), PSScriptAnalyzer clean, `bash -n` clean. Same caveat as before: not re-deployed to a live VM.

simpat-adam:
## Third review round validated and fixed (`508c18f8`) — and a review freeze

Five findings this round, all with a factual core; fixed in +104/−26. Verdicts:

| Finding | Verdict | Fix |
|---|---|---|
| `down.sh` misses deprecated `--volume` and last-value semantics of repeated flags | **Real** (alias confirmed by live test: Compose accepts `--volume` with a deprecation warning). `-v --volumes=false` resolves to *false* in Compose but the wrapper latched *true*, clearing markers while volumes survive | Alias recognized; the volumes boolean is now assigned per occurrence, matching Compose's last-value resolution. Two new runtime tests |
| REDEPLOY Part A: pasted block continues after failure; daemon-down reads as "volume absent" | **Real chain**, low likelihood | `docker info` preflight + explicit `exit 1` on teardown failure |
| `update.sh` Keycloak guard fails open when the container is gone (plain `down` keeps the volume) | **Real — the fairest finding of the round**; a stopped stack is a normal state | Guard fails closed: deployed ref from container, else from a persisted `.bootstrap/keycloak-image` (written after each successful update, cleared when the volume is dropped); volume-present + no-ref refuses |
| `$`-validation happens after services/marker state exists | **Real ordering flaw** | Whole-file scan up front in `bootstrap.ps1` (before any mutation/marker) and in `setup-env.ps1` before the first `docker compose up` |
| Part A doesn't clean gitignored `eng/docker-compose/.bootstrap`; stale staging survives checkout | **Real**, undermines the round-2 template fix | `rm -rf` added to Part A |

Verification: Pester 12/12, PSScriptAnalyzer clean, live tests for the `--volume` alias.

**Review freeze from here.** Three rounds in, the finding severity curve tells the story: round 2's best was *"the documented default path bricks both DMS services"*; this round's best was *"a guard fails open in a secondary flow"*, and the tail was *"an operator might type `-v --volumes=false`"*. Open-ended review of guards produces findings about guards, indefinitely. Further findings on this branch will be triaged in PR discussion — accepted only with a validated, realistic failure chain for this environment's actual lifetime — rather than turned into more commits. The merge decision now rests with human review of this PR.

simpat-jesus (approved):

PR #1110, opened 2026-07-14
[DMS-1208] Deduplicate PostgreSQL child and _ext stamping by affected document

samuellugo-ship-it:
The pinned summary comment on this PR is the single source of truth — it contains the full problem statement, implementation, test evidence, review guidance, and current status. Please read that comment.

### DIFF
--- a/reference/design/backend-redesign/epics/10-update-tracking-change-queries/10-mirror-stamping-triggers.md
+++ b/reference/design/backend-redesign/epics/10-update-tracking-change-queries/10-mirror-stamping-triggers.md
@@ -13,7 +13,7 @@ For representation-changing updates and deletes, the trigger must allocate exact
 
 The trigger's affected-document detection must also ignore updates whose only differences are stamp columns. This prevents mirror updates from causing recursive or redundant stamp activity.
 
-Known PostgreSQL follow-up: this story keeps PostgreSQL child / `_ext` `DocumentStamping` triggers on their current row-level shape. When one PostgreSQL statement changes multiple child or `_ext` rows that share the same root `DocumentId`, the final mirror still equals `dms.Document`, but the trigger can allocate more than one `ContentVersion` for that one affected document. Dedupe-by-document PostgreSQL statement-level stamping is tracked separately in `28-postgresql-statement-level-child-stamping.md`.
+Known PostgreSQL follow-up: this story keeps PostgreSQL child / `_ext` `DocumentStamping` triggers on their row-level shape as delivered here. When one PostgreSQL statement changes multiple child or `_ext` rows that share the same root `DocumentId`, the final mirror still equals `dms.Document`, but the trigger can allocate more than one `ContentVersion` for that one affected document. Dedupe-by-document PostgreSQL statement-level stamping is tracked separately in `28-postgresql-statement-level-child-stamping.md`. (Delivered in DMS-1208: that dedupe-by-affected-root, statement-level shape is now implemented for PostgreSQL child, nested-child, and `_ext` stamping.)
 
 ## Acceptance Criteria
 
--- a/src/dms/backend/EdFi.DataManagementService.Backend.Ddl.Tests.Unit/Fixtures/ddl-emission/expected/ddl-manifest/extension-mapping.json
+++ b/src/dms/backend/EdFi.DataManagementService.Backend.Ddl.Tests.Unit/Fixtures/ddl-emission/expected/ddl-manifest/extension-mapping.json
@@ -9,8 +9,8 @@
     },
     {
       "dialect": "pgsql",
-      "normalized_sql_sha256": "dc02e45cb14e52d80a19aeab86973552d296212f375e5fe448607bd8f1ff488c",
-      "statement_count": 68
+      "normalized_sql_sha256": "f76e315fe4dcc65caa2d911d732c0e363eed07c2ef0d720ba3246c569be5c4e7",
+      "statement_count": 89
     }
   ]
 }
--- a/src/dms/backend/EdFi.DataManagementService.Backend.Ddl.Tests.Unit/Fixtures/ddl-emission/expected/ddl-manifest/nested-collections.json
+++ b/src/dms/backend/EdFi.DataManagementService.Backend.Ddl.Tests.Unit/Fixtures/ddl-emission/expected/ddl-manifest/nested-collections.json
@@ -9,8 +9,8 @@
     },
     {
       "dialect": "pgsql",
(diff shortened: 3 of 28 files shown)


samuellugo-ship-it:
# DMS-1208 — source of truth (ready for review; review findings addressed, rebased on `main`)

_This is the single running summary for the PR and tells the whole story; please read this rather than expecting a per-push comment thread._

## Problem
PostgreSQL rendered **every** `DocumentStamping` trigger as one combined `BEFORE INSERT OR UPDATE OR DELETE … FOR EACH ROW` trigger. For non-root tables (child, nested-child, root `_ext`, collection-aligned `_ext`) that stamped `dms.Document` **once per affected source row**, so a single statement changing several child/`_ext` rows that share one owning root allocated multiple `dms.ChangeVersionSequence` values and over-advanced the change-version watermark. SQL Server already avoided this via a statement-level, de-duplicated `affectedDocs` workset. This was the PostgreSQL follow-up explicitly deferred from DMS-1173.

## Solution
Non-root PostgreSQL `DocumentStamping` triggers (those whose `MirrorStampTargetTable` is the owning resource root, not the source table) now render as **three event-specific statement-level triggers + functions** (INSERT / UPDATE / DELETE) using transition tables (`REFERENCING … FOR EACH STATEMENT`). PostgreSQL forbids transition tables on a multi-event trigger, so the combined trigger is split per event. Each function:
- builds a **deduplicated affected-root workset** — `SELECT DISTINCT` from the single transition table (INSERT/DELETE); a `UNION` of the new-image and old-image arms, correlated by the table's stable primary key and filtered by a null-safe stored-column value diff, for UPDATE (so a locator change stamps both the old and new owning root once);
- stamps `dms.Document` **once per distinct owning `DocumentId`** and mirrors the returned `ContentVersion`/`ContentLastModifiedAt` into the resource root (no second sequence allocation);
- keeps the **root-existence guard** so cascaded child/`_ext` deletes cannot advance the watermark past the root tombstone.

The legacy combined trigger is dropped for idempotent replacement. **Root and descriptor stamping stay row-level and byte-for-byte unchanged. The SQL Server renderer is untouched.** All rendering lives in `RelationalModelDdlEmitter.EmitPgsqlNonRootStatementLevelStamping`.

## Commits
1. **Renderer + unit tests + regenerated PostgreSQL DDL goldens** — the statement-level rendering; reworked PostgreSQL `DocumentStamping` unit tests (replacing the old `FOR EACH ROW` expectation) and extended the fixture with nested-child + collection-aligned `_ext` tables for four-kind coverage; regenerated `pgsql.sql` and the `pgsql` fields of `ddl.manifest.json`.
2. **PostgreSQL cardinality integration tests** — `PostgresqlChildStampingCardinalityTests`.
3. **SQL Server parity integration tests** — `MssqlChildStampingCardinalityTests` (1:1 mirror of the PostgreSQL suite).
4. **Regenerated the SchemaTools PostgreSQL provisioned-schema golden** — `provisioned-schema.pgsql.manifest.json`. This golden introspects the live applied schema; splitting the non-root stamping triggers changed only its `triggers` and `functions` sections (each +760 = 380 non-root tables × 2). All other sections (tables, columns, constraints, indexes, views, sequences, seed data) are byte-identical.
5. **Review follow-up (INSERT/DELETE cardinality coverage + doc note)** — extended the cardinality suites with a multi-row single-statement child **INSERT** dedup test and two **DELETE** tests (multi-row child delete dedup, and a cascading root delete that holds the visible watermark at the root tombstone's `ChangeVersion`); SQL Server parity twins for each. Also noted in the DMS-1173 story doc that its deferred PostgreSQL statement-level follow-up is delivered here. **Tests + one doc line only — no renderer / DDL / golden change.**

## Review follow-up
Addressed the review team's findings, folded into commit 5 and rebased onto latest `main`:
1. **(Medium) No behavioral coverage of the DELETE statement-level path** — the deduplicated DELETE trigger and its tombstone-ordering root-existence guard were exercised only at golden-SQL text level. Added a multi-row child DELETE dedup test and a cascading-root-delete test proving `dms.Document.ContentVersion` stays pinned at the root tombstone's `ChangeVersion` while the cascaded child/nested-child/`_ext` `AFTER … FOR EACH STATEMENT` stamps are skipped by the guard. Confirmed **non-vacuous** with a standalone probe: without the guard the child cascade fires with a populated `OLD TABLE` and advances the watermark past the tombstone.
2. **(Low) Multi-row single-statement INSERT dedup not exercised** — the seed helpers insert one child row per statement, so the `ins` trigger's `SELECT DISTINCT` was never fed a multi-row image. Added an `INSERT … VALUES (…),(…),(…)` child test asserting exactly one `ContentVersion` per owning root. (Child-only suffices — child and `_ext` share the identical `SELECT DISTINCT newtab.<root>` shape.)
3. **(Low) Prior-story doc wording** — these `epics/NN-*.md` are per-story specs, so `10-mirror-stamping-triggers.md` is DMS-1173's historical spec (its forward-reference to this story's `28-…md` is correct, and the living `update-tracking.md` is already accurate). Removed the one present-tense "current row-level shape" phrasing and added a "delivered in DMS-1208" note; left the DMS-1173 acceptance-criteria / out-of-scope statements intact.

## Verification
- **DDL unit + golden suite:** green; churn is rendering-only (see below).
- **Churn is rendering-only:** the only DDL that changed is `pgsql.sql` + the `pgsql` fields of `ddl.manifest.json`, plus the two live-introspection sections of the provisioned-schema PG manifest. `mssql.sql`, the derived-model manifests, the effective-schema manifests, and the MSSQL provisioned-schema golden are byte-identical — the signal that no semantics or SQL Server output moved. Commit 5 adds tests + one doc line only, so it moves no DDL/golden at all.
- **PostgreSQL 16.x (real DB):** the four-kind focused fixture provisions and the write path drives the new triggers; the cardinality suite is green — multi-row child and collection-aligned `_ext` **inserts, updates, and deletes** each advance `dms.GetMaxChangeVersion()` by exactly one with `dms.Document` == root mirror; plus no-op (0), locator-change (old + new root), multi-root (one distinct version per root), and a cascading root delete that holds the visible watermark at the root tombstone's `ChangeVersion`.
- **SQL Server 2022 (real DB):** the same parity assertions green, proving both engines advance `ContentVersion` cardinality identically for the same logical change.
- **CI:** green through the last pre-rebase run; re-running after the rebase onto `main` — see the PR checks.

## How to review
Start at `EmitPgsqlNonRootStatementLevelStamping` in `RelationalModelDdlEmitter.cs`; the emitted shape is visible in the regenerated `pgsql.sql` goldens (e.g. `IntegrationFixtures/profile-nested-and-root-extension-children`). The two `*ChildStampingCardinalityTests` files document the cross-engine behavior (insert / update / delete cardinality + tombstone ordering).

JBrenesSimpat (approved):

PR #1113, opened 2026-07-14
[DMS-1218] Make CMS API error responses conform to the Ed-Fi Error Response Knowledge Base

samuellugo-ship-it:
## Summary

Brings CMS Configuration Service API error responses into conformance with the [Ed-Fi Error Response Knowledge Base](https://edfi.atlassian.net/wiki/spaces/rc/pages/58589185/Error+Response+Knowledge+Base) (RFC 9457 Problem Details), **subject to the documented exceptions below**. Some endpoint branches previously returned bare framework responses, ad hoc `{ error, message }` JSON, or a non-standard validation `type`; query-parameter and framework binding/deserialization failures were also not classified to the correct taxonomy.

Jira: [DMS-1218](https://edfi.atlassian.net/browse/DMS-1218)

## What changed

- CMS non-success error responses return `application/problem+json` with the full Ed-Fi contract — `type`, `title`, `detail`, `status`, `correlationId`, `validationErrors` (`{}` when empty), and `errors` (`[]` when empty) — with the body `status` matching the HTTP status (subject to the OAuth/OIDC protocol exception below).
- Type URIs: data validation → `urn:ed-fi:api:bad-request:data`; query-parameter validation → `urn:ed-fi:api:bad-request:parameter`; not-found → `urn:ed-fi:api:not-found`; authorization → `urn:ed-fi:api:security:authorization`; non-unique-identity conflict → `urn:ed-fi:api:conflict:non-unique-identity`; internal errors → `urn:ed-fi:api:internal-server-error`. Framework request errors use Knowledge-Base types only: 415 → `urn:ed-fi:api:unsupported-media-type`; 400/413/other → the generic `urn:ed-fi:api:bad-request` with the actual status preserved.
- Query-parameter validation on **the 12 CMS collection endpoints already using paging validators** now returns the `bad-request:parameter` contract with value-free messages (the rejected value is never echoed), covering both semantically-invalid values (e.g. `limit=0`) and unbindable values (e.g. `offset=abc`).
- Framework binding/deserialization classification is made consistent across environments via `RouteHandlerOptions.ThrowOnBadRequest`: a malformed JSON body → sanitized `bad-request:data` (best-effort `$` JSON path, fixed message, no parser text); a query-binding failure → `bad-request:parameter`; invalid route binding and empty body → the generic `bad-request`; 404/405/413/415 keep their existing status-code-pages contracts.
- OAuth/OIDC protocol endpoints (`/connect/token`, `/connect/introspect`, `/connect/revoke`) keep the standard `{ error, error_description }` shape (documented protocol exception). Token-endpoint client authentication follows RFC 6749 §§2.3/5.2 and token-success follows RFC 6749 §5.1. `/connect/register` uses the Ed-Fi contract.
- Provider, database, and exception details are never surfaced to clients — they are logged server-side only.
- The OpenAPI document at `/metadata/specifications` publishes the shared `ProblemDetails` schema (all seven members required) and references the reusable error responses.

## Documented exceptions

- **OAuth/OIDC error shape** — the three `/connect/*` protocol endpoints return `{ error, error_description }` (RFC 6749 §5.2 / RFC 7662 / RFC 7009), not the Ed-Fi Problem Details contract.
- **ClaimSet concurrency** — update/delete/copy multi-user conflicts return the generic `urn:ed-fi:api:conflict` (409). This is an internal shared claims-hierarchy `LastModifiedDate` race, not a failed client precondition; 412 is not used because the Knowledge Base reserves `optimistic-lock-failed` for the ETag/If-Match feature, which CMS does not implement.
- **Keycloak token revocation** — `/connect/revoke` returns `503 temporarily_unavailable` under Keycloak because CMS does not yet proxy revocation to Keycloak. This is a documented **supported-provider limitation**, not RFC 7009-compliant temporary unavailability. `/connect/revoke` also does not yet perform RFC 7009 §2.1 client authentication or token-ownership verification (self-contained provider included). Tracked in follow-up [DMS-1327](https://edfi.atlassian.net/browse/DMS-1327), which is a closure condition of this ticket.
- **Malformed-body deserializer signals** — empty chunked bodies, invalid UTF-8, and declared-charset mismatches surface as a `JsonException` with no stable structured discriminator, so they share the same sanitized `bad-request:data` contract (content-length empty bodies remain the generic `bad-request`).

## Scope

Production scope is CMS (`src/config`) plus **one DMS E2E test-helper compatibility change** — `src/dms/tests/EdFi.DataManagementService.Tests.E2E/Authorization/ProfileAwareAuthorizationProvider.cs` — so the helper tolerates a pre-existing profile reported as `409 urn:ed-fi:api:conflict:non-unique-identity` (no DMS production change). CMS success responses and DMS API error behavior are out of scope.

## Decisions & follow-up

Implementation decisions are recorded on DMS-1218 (comments `96051`, `96084`) and in the "Documented exceptions" section above. Exact per-operation OpenAPI error responses remain an architect-accepted heuristic limit, deferred to [DMS-1293](https://edfi.atlassian.net/browse/DMS-1293). RFC 7009 revocation client authentication, token-ownership verification, and Keycloak support are tracked in [DMS-1327](https://edfi.atlassian.net/browse/DMS-1327).

[DMS-1218]: https://edfi.atlassian.net/browse/DMS-1218?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DMS-1327]: https://edfi.atlassian.net/browse/DMS-1327?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


samuellugo-ship-it:
**Decisions** are recorded on the Jira ticket ([DMS-1218](https://edfi.atlassian.net/browse/DMS-1218) comments `96051` and `96084`) so they are not re-flagged. Exact per-operation OpenAPI error documentation is deferred to [DMS-1293](https://edfi.atlassian.net/browse/DMS-1293).


[DMS-1218]: https://edfi.atlassian.net/browse/DMS-1218?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DMS-1293]: https://edfi.atlassian.net/browse/DMS-1293?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

samuellugo-ship-it:
**Re: [Medium] `Program.cs:88` applies Ed-Fi status-code pages to OAuth endpoints — keeping the current behavior (no change)**

Raising this to explain why the current response is intended rather than a defect. `GET /connect/token` (and other wrong-method requests to the token-family endpoints) returns `405 Method Not Allowed` with `Allow: POST` and an Ed-Fi Problem Details body (`urn:ed-fi:api:method-not-allowed`) — not an OAuth `{ error, error_description }` — and that is correct:

1. **The OAuth handlers are never reached.** `/connect/token`, `/connect/introspect`, and `/connect/revoke` are mapped POST-only (`IdentityModule.MapEndpoints`). A GET is rejected by routing as a 405 before any OAuth handler runs, so it is not an OAuth protocol error at all.
2. **`UseStatusCodePages` only fills otherwise-empty responses.** It writes only the body, preserving the 405 status and the `Allow: POST` header. Real OAuth protocol errors — a POST the handler processes and rejects — already write a JSON body via `OAuthErrorResults`, so they are never overwritten and keep their `{ error, error_description }` shape.
3. **RFCs.** RFC 6749 §3.2 requires clients to use POST for token requests. §5.2 defines token-endpoint OAuth errors as request-*processing* failures (HTTP 400, or 401 for `invalid_client`); it does not require a routing-level 405 (wrong method) to be expressed as OAuth JSON. RFC 9110 §15.5.6 requires a 405 to carry `Allow`.
4. **The documented OAuth exception is operation-based, not path-based.** Per the decision recorded on DMS-1218 (comment `96051`), the `{ error, error_description }` exception covers errors emitted by the OAuth protocol *handlers*, not every framework error that happens to occur at those paths. A wrong-method 405 never invokes a handler.

Changing this would either misrepresent a 405 as a 400 (and drop `Allow`) or invent a non-standard "405 + OAuth JSON" hybrid that no RFC requires. The behavior is locked by `StatusCodePagesPipelineTests.It_returns_405_with_allow_header_and_complete_body`. Happy to revisit if there is a normative requirement or product decision establishing the exception as path-based.

PR #1134, opened 2026-07-29
[DMS-1062] Implement View-based Authorization Strategy for GET-many

jleiva-gap:
## What

Implements the custom view-based (auth view) authorization strategy for GET-many on regular resources and descriptors, and orders the AND/OR authorization terminals per the CMS-configured execution order.

`OwnershipBased` is **not** implemented here. It remains known-but-not-enabled (501) for every operation, GET-many included. DMS-1060 owns the complete strategy, including the tenant-qualified CMS `ApplicationContext` token source and write-side `CreatedByOwnershipTokenId` stamping.

## Why

Custom view-based authorization was unimplemented for GET-many, so resources configured with a `{BasisResource}With...` strategy failed closed instead of filtering the page. Per `reference/design/backend-redesign/design-docs/auth.md`, Namespace-based and custom view-based are AND strategies that execute in the order configured in the Admin DB, ownership-based executes last within that group, and relationship OR strategies run afterward. Getting that order right determines which failure the client sees, so the planner has to select the terminal by configured index rather than by outcome type.

## Changes

Custom view authorization planning and SQL

- Add `CustomViewAuthorizationPlanner` and `CustomViewAuthorizationContracts` to resolve a configured `{BasisResource}With...` strategy to its basis resource, the `auth` view, and the DocumentId join path to the subject root table.
- Extend `RelationshipAuthorizationStrategyClassifier` to recognize the custom-view naming convention, resolve the preferred basis resource across projects, and report `UnknownCustomViewBasisResource` when it cannot be resolved.
- Emit custom view joins in `PageDocumentIdSqlCompiler` for both dialects, keyed on the basis resource DocumentId rather than natural keys.
- Add `CustomViewAuthorizationValidator`, which probes each planned auth view before short-circuiting a request, so a missing view or a view without a DocumentId column surfaces as a controlled 500 rather than an uncaught DB error. Failures raise `CustomViewAuthorizationValidationException`, mapped to a system-error response in `CoreExceptionLoggingMiddleware`.
- Both engines accept a regular view or a materialized view as the auth object and require a `bigint` `DocumentId`. The checks read the engine catalogs (`pg_catalog.pg_class` / `pg_catalog.pg_attribute`, `sys.views` / `sys.columns`) rather than `information_schema`, whose column metadata omits materialized views. Neither engine emits an extra root-to-basis path probe: a resolved-path error surfaces from the page query itself, identically on both.
- The PostgreSQL probe wraps its catalog guards in a `DO` block that embeds the schema, view, and column names as string literals, so the dollar-quote delimiter is derived from those literals (`$dmscv$` escalating to `$dmscvx$`, …) rather than fixed at `$$`. A view or strategy name containing `$$` would otherwise close the block early and leave the rest of the body parsed as top-level SQL. The bind probe stays row-free (`LIMIT 0` / `TOP (0)`).

OwnershipBased stays out of scope

- `OwnershipBased` is classified known-but-not-enabled for every operation and produces the existing fail-closed 501. No GET-many path builds ownership checks or an ownership parameterization, and `RelationalAuthorizationContext.Create` has no ownership token source, so every production request carries an empty ownership list. The classifier has no ownership-promotion mode to re-enable.
- A custom view configured ahead of an `OwnershipBased` terminal is still validated before the 501 is returned, so an earlier custom-view configuration failure keeps its own response.
- No ownership execution scaffolding ships either. `OwnershipTokenIdParameterization` and its factory/validator, `OwnershipTokenIdLimitExceededException`, `OwnershipTokenIdSqlHelper`, `OwnershipTokenIdParameterValues`, the page-SQL ownership filter, the ownership parameter-budget argument, the MSSQL ownership cap constant, and the `RelationalAuthorizationContext` ownership input are all absent — deleted, or reverted to `main` where the file predates this branch. DMS-1060 introduces them alongside the token source and stamping they depend on.

Authorization terminal ordering

- `RelationalAuthorizationPlanner` selects the GET-many terminal by `RawConfiguredIndex` instead of by outcome type: a Namespace terminal configured ahead of a classifier failure (unknown custom-view basis, or an unrecognized strategy) now wins, while a classifier failure configured before `NamespaceBased` still wins. Both Namespace terminals participate — no configured prefixes (403) and no usable root column (500). A failure whose configured position is unknown still fails closed, and non-GET-many operations and requests without a Namespace terminal are unchanged.
- Add `CustomViewAuthorizationTerminalOrdering` so only custom views configured before a terminal are validated, and later ones are skipped. The same rule applies to every terminal family, not just the namespace ones: a classifier security-configuration failure and a custom-view planning failure each suppress the custom views configured after them, so a later missing or non-conforming `auth.{StrategyName}` cannot mask an earlier configuration error. `CustomViewAuthorizationPlanOutcome.SecurityConfiguration` carries its successfully planned checks for the same reason — the earlier views still have to be validated before the failure is reported.
- Apply the same ordering to the descriptor GET-many path in `DescriptorReadHandler`, which shares the planner with regular resources. Its 501 terminal validates a resolved custom view first, matching the relational query path, and its classifier-failure terminal applies the same configured-index filter.
- Descriptor GET-many reports the specific no-custom-view-join-path error rather than the generic unknown-strategy message; the sentence is shared with the regular-resource path so the two cannot drift. Unknown-basis and invalid-strategy wording is unchanged, and diagnostics keep the specific `RelationshipAuthorization.{FailureKind}` discriminator.
- Terminals that are genuinely last in the AND chain still validate every resolved custom view: `OwnershipBased` executes after all of them, so its 501 is reported only after each one has run.

Wiring

- Thread the custom view authorization spec through `RelationalDocumentStoreRepository` and `DescriptorReadHandler`.

Tests

- Unit coverage for the planner, classifier, page SQL compiler, custom view validator, and parameter budget, including regression tests that pin each terminal ordering in both directions — earlier configuration failures are not masked by later custom views, and later custom views are not validated once an earlier terminal wins — for both the regular-resource and descriptor paths.
- PostgreSQL and MSSQL integration coverage for custom view and relational query authorization, including a real-database descriptor GET-many custom-view filter and its missing-view failure, and a real-database auth view whose name contains `$$`.
- E2E scenarios in `CustomViewAuthorization.feature`: custom view filters GET-many, empty custom view returns an empty success, and missing or DocumentId-less views return a system error. The step definitions are database-engine-aware, and one successful filtering scenario is tagged `@MssqlRepresentative`.
- Malformed-`AUTH1` (`ns1|0`) coverage at the SQL Server backend and API boundaries, distinct from the existing unmappable-index (`ns1|9|m`) cases.

## Deferred

- Custom-view validation runs on the read-side `IRelationalCommandExecutor`, so a GET-many no longer opens a write transaction to probe the auth views. It does not share a connection with the page query: the concrete PostgreSQL and SQL Server executors open and dispose a pooled read connection per `ExecuteReaderAsync` call, so validation costs a separate read connection and round trip. It remains a separate statement from the page query: the standalone validator enforces conditions the page query cannot reliably detect (table-versus-view, incompatible `DocumentId` types on empty or disjoint views). Folding it into the page query would need an execution-lifecycle design across the session, hydrator, and executor APIs. Note the cost this accepts: **every** custom view-based GET-many issues a validation round trip before the page query, on each request. There is deliberately no cache or TTL memo — custom views are CMS-configured and can change without a deployment, so any caching strategy is a team decision about staleness tolerance, not something to settle inside this story.
- Transitive (multi-step) custom-view paths compile to an uncorrelated subquery that re-scans the resource root table — `r.DocumentId IN (SELECT t0.DocumentId FROM <root> t0 JOIN ... WHERE ... IN (SELECT cv.DocumentId ...))` — instead of a correlated `EXISTS`. Left as-is on purpose: this is hot-path page SQL for every transitive custom view, and at small row counts the planner frequently flattens it to the same plan, so a rewrite needs a realistic-row-count measurement before it is worth the regression surface. Direct (single-step) paths are unaffected — they emit a plain column-`IN`-view predicate.


samuellugo-ship-it:
Tokenless `OwnershipBased` is no longer partially enabled: the GET-many reclassification gate now requires non-empty `OwnershipTokenIds`, so it stays `KnownButNotEnabled` (501) instead of returning an empty `QuerySuccess`. DMS-1060 still owns stamping and the CMS application-context plumbing (blocked by DMS-1372/DMS-1373). Ownership support here is provisional, only to validate custom-view composition against real ownership context. Also adds the MSSQL 1,999/2,000 ownership-token cap, real SQL Server coverage for non-empty ownership binding, and engine-aware custom-view E2E steps with one `@MssqlRepresentative` scenario.

The unrelated malformed-AUTH1 coverage previously in `b7a4ae80` was dropped from this series and is preserved locally for the correct ticket.

Note: one `Backend.Tests.Unit` run reported a single failure that I could not identify; four subsequent full runs were 2308/2308, so it is unreproduced rather than diagnosed.

samuellugo-ship-it:
Restored the malformed-`AUTH1` (`ns1|0`) coverage that a mid-cycle review had asked to drop as unrelated; a later review identified its absence as a coverage gap. Test-only cherry-pick of `b7a4ae80`, 216 additions across three files, no production change, and the existing `ns1|9|m` `PayloadMappingFailed` fixtures untouched.

**Review-loop note — please do not remove this coverage again** without explicitly resolving that contradictory guidance. The two payloads exercise different branches and both are needed:

| Payload | Diagnostic |
|---|---|
| `ns1\|9\|m` | `NamespaceAuthorization.Auth1.PayloadMappingFailed` (parses, index unmappable) |
| `ns1\|0` | `NamespaceAuthorization.Auth1.InvalidPayload` (third section omitted, parse fails) |

samuellugo-ship-it:
Wrapped custom-view page-query provider errors on both GET-many paths (they previously escaped as unhandled `DbException` and lost the `urn:ed-fi:api:system` contract), plus live-DB intersection tests proving AND semantics on both engines.

samuellugo-ship-it:
Heads-up for deployment: SQL Server custom-view validation now compares catalog names with a binary collation, so an `auth.{StrategyName}` view or `DocumentId` column that differs only by case is rejected with the system 500 instead of silently passing on a case-insensitive collation.

JBrenesSimpat (approved):

JBrenesSimpat (approved):

PR #1160, opened 2026-08-09
[DMS-1315] Cache Reads with Relational Fallback

bradbanister:


src/dms/backend/EdFi.DataManagementService.Backend/RelationalReadTargetLookupService.cs
```diff
@@ -22,8 +22,12 @@ public abstract record RelationalReadTargetLookupResult
 {
     private RelationalReadTargetLookupResult() { }
 
-    public sealed record ExistingDocument(long DocumentId, DocumentUuid DocumentUuid, long ContentVersion)
-        : RelationalReadTargetLookupResult;
+    public sealed record ExistingDocument(
+        long DocumentId,
+        DocumentUuid DocumentUuid,
+        long ContentVersion,
+        DateTimeOffset ContentLastModifiedAt = default
```
simpat-adam: **[Correctness]** `= default` makes a broken candidate silently constructible.

This value becomes `DocumentCacheReadAccelerationCandidate.ContentLastModifiedAt`, which the classifier compares to `dms.DocumentCache.LastModifiedAt`. A mismatch is not a miss — `DocumentCacheReadLookupContracts.cs:814-825` classifies it `DeterministicInvariantFailure`, which escalates to a target-fatal health diagnostic. So any site that omits the argument gets `0001-01-01`, never hits cache, and poisons target health on every read of that resource.

The one production caller already throws when the column is null. **Change requested:** make the parameter required and update the test construction sites.

src/dms/tests/EdFi.DataManagementService.Tests.Integration/Scenarios/DocumentCacheReadAccelerationScenario.cs
```diff
@@ -0,0 +1,971 @@
...
+            "cache-missing-001",
+            "Relational Missing"
+        );
+        var missingStudentDocumentUuid = Guid.Parse(missingStudentLocationPath.Split('/')[^1]);
+        DocumentMetadata missingStudentMetadata = await ReadDocumentMetadataAsync(
+            harness,
+            missingStudentDocumentUuid
+        );
+        await DeleteCacheRowsAsync(harness, missingStudentMetadata.DocumentId);
+        (await CountCacheRowsAsync(harness, missingStudentMetadata.DocumentId))
```
simpat-adam, line 132: **[Acceptance criteria]** Direct fill is never proven to write a cache row outside unit fakes.

The design names "direct-fill writer invocation, direct-fill timeout" in the provider-integration acceptance list. This count runs *before* the GET on line 135 — and that GET is precisely the request that should direct-fill — but the count is never re-read. Elsewhere the scenario only asserts direct-fill *skips*; the provider lookup tests cover lookup SQL only; the coordinator unit suite uses a fake materializer/writer.

**Change requested:** after the miss-driven GET, assert a `dms.DocumentCache` row now exists with the expected `ContentVersion`/`StreamEtag`, and add a `DirectFillTimeout` case.

src/dms/backend/EdFi.DataManagementService.Backend/DocumentCacheReadResponseShaper.cs
```diff
@@ -0,0 +1,351 @@
...
+                        hit.Candidate,
+                        hit.DocumentJson
+                    )
+                );
+            }
+
+            return DocumentCacheReadLookupResult<QueryResult>.Hit(
+                new QueryResult.QuerySuccess(
+                    edfiDocs,
+                    authorizedCandidatePage.TotalCount is null
```
simpat-adam: **[Correctness]** Cache hit and relational fallback use different rules for the `Total-Count` header.

Here the total count is emitted when `TotalCount is not null`; the fallback and empty-page paths use `Paging.IncludesTotalCount` / `PaginationParameters.TotalCount` (`RelationalDocumentStoreRepository.cs:1163`, `DescriptorReadHandler.cs:454`). These agree only because `DescriptorQueryPageKeysetPlanner.cs:74` sets `IncludeTotalCountSql: paginationParameters.TotalCount` — an incidental planner property, not an invariant of the candidate-page contract. The design forbids any public difference between a hit and its fallback.

**Change requested:** carry the request's total-count decision on `DocumentCacheReadAccelerationCandidatePage` and have both paths read it.

src/dms/backend/EdFi.DataManagementService.Backend/DocumentCacheReadAccelerationCoordinator.cs
```diff
@@ -0,0 +1,1490 @@
...
+            return complete.Result;
+        }
+
+        var candidateSelection =
+            selectionResult as DocumentCacheReadAccelerationQuerySelectionResult.CandidatePage
+            ?? throw new InvalidOperationException(
+                $"Unsupported query candidate selection result '{selectionResult.GetType().Name}'."
+            );
+
+        if (candidateSelection.AuthorizedCandidatePage.IsEmpty)
```
simpat-adam: **[Dead code]** This empty-page branch is unreachable, and it encodes a third total-count rule.

Both production callers return `Complete` for empty pages before a `CandidatePage` is ever constructed (`RelationalDocumentStoreRepository.cs:1158`, `DescriptorReadHandler.cs:450`). The dead branch decides total count with `TotalCount is null`, which is neither of the two rules the live callers use.

**Change requested:** delete it, or make it the single implementation and drop the caller-side duplicates.

src/dms/backend/EdFi.DataManagementService.Backend/DocumentCacheReadAccelerationCoordinator.cs
```diff
@@ -0,0 +1,1490 @@
...
+                        targetContext,
+                        exception,
+                        DocumentCacheReadTelemetryLabel.UnexpectedException
+                    );
+                }
+            }
+        }
+        finally
+        {
+            _readTelemetry.RecordDirectFillDuration(
```
simpat-adam, line 1360: **[Telemetry fidelity]** The direct-fill duration histogram is labeled with the last candidate's outcome.

`directFillDurationOutcome` is reassigned inside the per-candidate loop, so this single sample carries whatever the final iteration left. A GET-many page where nine fills succeed and the tenth fails records the whole elapsed time under `Failed`.

**Change requested:** record per-candidate durations, or use a distinct aggregate label that does not impersonate a per-document outcome.

src/dms/backend/EdFi.DataManagementService.Backend/DocumentCacheReadAccelerationCoordinator.cs
```diff
@@ -0,0 +1,1490 @@
...
+                or DocumentCacheWriterResult.AlreadyCurrentNoWork;
+
+    private static string SelectDirectFillCancellationReason(
+        CancellationToken requestCancellationToken,
+        CancellationTokenSource directFillTimeout
+    )
+    {
+        if (requestCancellationToken.IsCancellationRequested)
+        {
+            return "CallerCanceled";
```
simpat-adam: **[Consistency]** These string literals bypass the label constants.

`DocumentCacheReadTelemetryLabel` defines `CallerCanceled`, `TimedOut`, and `Canceled` (`DocumentCacheReadTelemetry.cs:48-50`), and every other outcome in this file goes through the constants. A rename would split the metric series with no compile error.

**Change requested:** use the constants.

src/dms/backend/EdFi.DataManagementService.Backend/DocumentCacheReadAccelerationCoordinator.cs
```diff
@@ -0,0 +1,1490 @@
...
+
+    public bool IsAdapterAcquisitionFailure { get; }
+
+    public static DocumentCacheReadLookupResult<TResult> Hit(TResult result)
+    {
+        ArgumentNullException.ThrowIfNull(result);
+
+        return new DocumentCacheReadLookupResult<TResult>(
+            result,
+            DocumentCacheReadAccelerationFallbackReason.CacheLookupMiss,
```
simpat-adam: **[Consistency]** `Hit()` stamps `FallbackReason = CacheLookupMiss` on a hit.

It is never read while `CachedResult` is non-null, but it is a misleading state in a type that otherwise guards its invariants hard (see the `ArgumentException` at line 146).

**Change requested:** add a `None` fallback reason for the hit case.

src/dms/backend/EdFi.DataManagementService.Backend.Postgresql/PostgresqlReferenceResolverServiceCollectionExtensions.cs
```diff
@@ -80,7 +80,13 @@ public static IServiceCollection AddPostgresqlReferenceResolver(this IServiceCol
...
             PostgresqlSessionDocumentHydrator
         >();
-
+        services.Replace(
+            ServiceDescriptor.Scoped<
+                IDocumentCacheReadLookupAdapter,
+                PostgresqlDocumentCacheReadLookupAdapter
+            >()
+        );
+        services.AddDocumentCacheReadAccelerationCoordinator();
```
simpat-adam, line 89: **[Overhead]** Registered unconditionally, so the disabled feature still emits two counter series per read.

This replaces the passthrough coordinator regardless of configuration (same at `MssqlReferenceResolverServiceCollectionExtensions.cs:83`). With `ReadAcceleration:Enabled` false — the default — every external GET and GET-many emits `RecordFallback(ReadAccelerationDisabled)` **and** `RecordDirectFill(SkippedReadAccelerationDisabled)`, six tags each. Two permanent time series and two `Counter.Add` calls per read for a feature nobody turned on.

**Change requested:** keep `PassthroughDocumentCacheReadAccelerationCoordinator` registered when the option is off, or gate the disabled-path counters.

src/dms/backend/EdFi.DataManagementService.Backend/DescriptorReadHandler.cs
```diff
@@ -110,101 +223,525 @@ public async Task<GetResult> HandleGetByIdAsync(
...
+        catch (InvalidOperationException ex)
+        {
+            return new QueryResult.UnknownFailure(ex.Message);
+        }
+        catch (ArgumentException ex)
+        {
+            return new QueryResult.UnknownFailure(ex.Message);
+        }
+
+        var descriptorRowsByDocumentId = descriptorRows.ToDictionary(
```
simpat-adam: **[Defensive consistency]** This `ToDictionary` sits outside the try/catch that converts reader failures into `QueryResult.UnknownFailure`, so a duplicate `DocumentId` would throw `ArgumentException` past the ProblemDetails boundary. It cannot happen with the PK today, but `SelectedDescriptorQueryRowsStillMatch` defends the identical case with `TryAdd` at line 649.

Also, lines 617-622 do `.Where(ContainsKey).Select(id => dict[id])` — a double lookup where `TryGetValue` reads better.

**Change requested:** match the `TryAdd` pattern used below, and collapse the double lookup.

simpat-adam (commented):
## Code Review — [DMS-1315] Cache Reads with Relational Fallback

*Reviewed by Claude Code (claude-opus-5) against `reference/design/backend-redesign/epics/18-document-cache/05-cache-backed-read-path.md` at head `13dab80d`.*

The design contract is implemented closely — decision table, fence semantics, the `Tracking`-read / `Rebuilding`-write eligibility split, all-or-nothing page rule, cancellation classification, and response-shaping order all match the Resolved Runtime Contract and both Q&A rounds. Served-ETag inputs were checked against every other `ServedEtagContext` site; hit and fallback compose identically. The single-statement consistency requirement is safe because RCSI is a hard provider prerequisite and the lookup fences on it.

Inline comments below are the actionable items, recorded here for tracking. Ranked: the `= default` parameter (correctness), the direct-fill evidence gap (acceptance criteria), then consistency/dead-code/telemetry items.

Note: unit and integration CI jobs were still pending at review time; no test results are claimed here.

src/dms/backend/EdFi.DataManagementService.Backend/DocumentCacheReadAccelerationCoordinator.cs
```diff
@@ -0,0 +1,1487 @@
...
+                DocumentCacheReadTelemetryLabel.SkippedTargetIneligible
+            );
+            return;
+        }
+
+        TimeSpan timeout = targetContext.EffectiveSettings.DirectFillTimeout;
+        using CancellationTokenSource directFillTimeout = CancellationTokenSource.CreateLinkedTokenSource(
+            cancellationToken
+        );
+        directFillTimeout.CancelAfter(timeout);
```
simpat-adam: **[Correctness]** Direct-fill setup runs outside the guard that makes direct fill best effort, so it can destroy an already-computed successful response.

The design is explicit that direct-fill failures "must not replace, delay beyond the direct-fill budget, or retry the already-computed relational response" (design lines 189-191). Everything from line 1180 to 1197 sits *before* the `try` at line 1199, so anything that throws here escapes `TryDirectFillAsync`, escapes `GetByIdAsync`/`QueryAsync`, and reaches `CoreExceptionLoggingMiddleware` — turning a 200 that relational fallback already computed into a 500.

This is reachable through configuration. `CancellationTokenSource.CancelAfter(TimeSpan)` throws `ArgumentOutOfRangeException` when `delay.TotalMilliseconds > int.MaxValue` (~24.85 days), and `DocumentCacheOptionsValidator.ValidateReadAcceleration` (`DocumentCacheOptions.cs:275-279`) only rejects non-positive values — there is no upper bound. So `"DataManagement:DocumentCache:ReadAcceleration:DirectFillTimeout": "30.00:00:00"` binds, validates, starts the host, and then fails every cache-miss read with a 500 after the correct body was already in hand. `new DataStoreId(...)` and `new DocumentCacheProjectionTargetKey(...)` on lines 1186-1190 are exposed the same way.

**Change requested:** move the timeout/context setup inside the `try`, and add an upper-bound check to `ValidateReadAcceleration` so a nonsensical `DirectFillTimeout` fails at startup rather than per request.

src/dms/backend/EdFi.DataManagementService.Backend/DocumentCacheProjectionObservation.cs
```diff
@@ -783,11 +818,23 @@ void MarkTargetContextNoncurrent(
...
+}
+
 public sealed class DocumentCacheProjectionObservationStore
     : IDocumentCacheProjectionObservationProvider,
         IDocumentCacheProjectionObservationSink,
-        IDocumentCacheProjectionCurrentTargetHealthSink
+        IDocumentCacheProjectionCurrentTargetHealthSink,
+        IDocumentCacheProjectionTargetDiagnosticSink
 {
+    private static readonly int PendingTargetDiagnosticLimit = new DocumentCacheProjectorOptions().PageSize;
```
simpat-adam: **[Consistency]** This bound reads like configuration but is a hard-coded 100 that ignores the operator's `Projector:PageSize`.

`new DocumentCacheProjectorOptions().PageSize` constructs a throwaway default instance, so this is always the compiled-in default (`DocumentCacheOptions.cs:48`) no matter what `DataManagement:DocumentCache:Projector:PageSize` is set to. Every other diagnostic bound in this file uses `EffectiveProjectorPageSize` taken from the snapshot — see line 454 and the `EnsureMatchingPageSize` calls at 442-450 — so an operator who raises `PageSize` gets a pending buffer that silently stays at 100 while the merged buffer grows.

The cap itself is a good fix; pending diagnostics for a context that never becomes current would otherwise grow without limit. **Change requested:** either take the bound from the same effective page size the rest of the store uses, or make it an explicit named constant so it does not read as a config-derived value.

src/dms/backend/EdFi.DataManagementService.Backend/DocumentCacheReadResponseShaper.cs
```diff
@@ -0,0 +1,351 @@
...
+    IReadableProfileProjector readableProfileProjector,
+    IServedEtagComposer servedEtagComposer,
+    IOptions<ResourceLinksOptions> linksOptions,
+    ILogger<DocumentCacheReadResponseShaper>? logger = null
+) : IDocumentCacheReadResponseShaper
+{
+    private const string IdPropertyName = "id";
+    private const string EtagPropertyName = "_etag";
+    private const string LastModifiedDatePropertyName = "_lastModifiedDate";
+    private const string LastModifiedDateFormat = "yyyy-MM-ddTHH:mm:ss'Z'";
```
simpat-adam: **[Coupling]** This format string is a fourth copy, and it is now the only one whose divergence is target-fatal.

The identical literal lives in `DocumentCacheMaterializer.cs:25` (the writer that produces `DocumentCache.DocumentJson`), `RelationalReadMaterializer.cs:125`, and `DescriptorDocumentMaterializer.cs:14`. The duplication predates this PR, but the consequence here is new: `ValidateCachedDocument` (line 287) compares the stored `_lastModifiedDate` against `FormatLastModifiedDate(candidate.ContentLastModifiedAt)` with `StringComparison.Ordinal`, and a mismatch raises `DocumentJsonLastModifiedDateMismatch`, which `TryShape` maps to `DeterministicInvariantFailure` and a target-fatal health diagnostic — not a cache miss. So if anyone ever changes the writer's format without changing this one, every read of every cached document poisons target health, and the goldens will not catch it because both sides currently agree.

**Change requested:** hoist the format to one shared constant that both the materializer and this validator reference, so the writer and the reader cannot drift.

simpat-adam (commented):
## Re-review — [DMS-1315] Cache Reads with Relational Fallback

*Reviewed by Claude Code (claude-opus-5) against `reference/design/backend-redesign/epics/18-document-cache/05-cache-backed-read-path.md` at head `1d147c4a9`. Previous review was at `13dab80d`.*

CI is green at this head: 90 checks pass, 46 skipped, 0 failing, 0 pending.

### Status of the nine findings from the previous review

**Fixed (7):**

| Finding | Resolution |
|---|---|
| `= default` on `ContentLastModifiedAt` | Now validated in `RelationalReadTargetLookupResult.ExistingDocument` — `ValidateContentLastModifiedAt` throws on `default`, so a broken candidate is no longer constructible. Stronger than what I asked for. |
| Direct-fill never proven to write a row | `DocumentCacheReadAccelerationScenario.cs:152-158` now reads the `dms.DocumentCache` row back and asserts its contents; a `DirectFillTimeout` case was added at lines 190-221 asserting `Attempted`/`TimedOut` and that the relational body survives. |
| `Total-Count` divergence between hit and fallback | `DocumentCacheReadAccelerationCandidatePage.IncludesTotalCount` now carries the request decision, and the shaper reads it (`DocumentCacheReadResponseShaper.cs:134`). |
| Direct-fill cancellation string literals | Replaced by `SelectDirectFillCancellationReason` over `DocumentCacheReadTelemetryLabel` constants. |
| `Hit()` stamping `FallbackReason = CacheLookupMiss` | `DocumentCacheReadAccelerationFallbackReason.None` added and asserted by the invariant guards at lines 141/149. |
| Disabled feature still emitting two counter series per read | `AddDocumentCacheReadAccelerationCoordinator` now returns `PassthroughDocumentCacheReadAccelerationCoordinator` when `ReadAcceleration:Enabled` is false, and the disabled-path counters were deleted outright. |
| `ToDictionary` outside the failure boundary in `DescriptorReadHandler` | Both sites now use `TryAdd` + `TryGetValue`. |

**Still open (2):**

1. **`DocumentCacheReadAccelerationCoordinator.cs:557` is still unreachable.** The `fix-dead-code` commit aligned the branch's total-count rule with the callers but did not remove it. Both production callers still return `Complete` before a `CandidatePage` is constructed — `RelationalDocumentStoreRepository.cs:1159` (`candidatePage.IsEmpty`) and `DescriptorReadHandler.cs:452` (`rowsPage.Page.Rows.Count == 0`), and `CreateDescriptorReadAccelerationCandidatePage` maps rows 1:1 so a non-empty rows page can never yield an empty candidate page. The empty-page rule is now written three times and agrees three times, which is the fragile part. **Change requested:** delete the coordinator branch, or make it the single implementation and drop the two caller-side short-circuits.

2. **`DocumentCacheReadAccelerationCoordinator.cs:1320` still labels the direct-fill duration histogram with the last candidate's outcome.** `directFillDurationOutcome` is reassigned inside the per-candidate loop (lines 1236, 1267, 1284, 1296, 1308) and the single sample in the `finally` reads whatever the final iteration left. A GET-many page where nine fills succeed and the tenth fails records the whole elapsed time under `Failed`. The new integration assertion at `DocumentCacheReadAccelerationScenario.cs:219` passes only because that case has one candidate. **Change requested:** record per-candidate durations, or use a distinct aggregate label that does not impersonate a per-document outcome.

### New in this review

The post-review commits are otherwise good work. `fix-scope` correctly deletes `WriteDirectFillCandidateWithoutWork` and `SelectDirectFillRepairAction`, bringing the writer back in line with the design's "direct fill is not a request-path cache repair guarantee" rule (design lines 176-180). `fix-cancel-and-dispose` correctly excludes `ObjectDisposedException` from `IsResultShapeFailure` — it derives from `InvalidOperationException`, so request-abort disposal was previously being classified as a target-fatal invariant failure — and the three middleware rethrows keep caller cancellation off the 500 path. `fix-page-fallback` restores `CustomViewAuthorizationValidationException` wrapping on both selected-page hydration paths, which the candidate/hydrate split had dropped.

Three inline comments below: one correctness item on the direct-fill guard boundary, and two consistency items.

simpat-adam (approved):

simpat-adam (approved):

simpat-adam (approved):

simpat-adam (approved):

PR #1164, opened 2026-08-11
[DMS-1385] Candidate planning and provider cursor SQL

simpat-adam:
## Summary

Extends the shared page-`DocumentId` candidate contract with an explicit candidate mode and compiles
seek-based cursor page selection for PostgreSQL and SQL Server. Traditional page-selection SQL is
unchanged, byte for byte.

This story supplies compiled plans only. Cursor execution, hydration, and the `Next-Page-Token`
header belong to DMS-1386; partition window SQL and the `/partitions` pipeline belong to DMS-1387.
`RelationalDocumentStoreRepository.QueryDocuments` still returns `QueryFailureNotImplemented` for
non-traditional paging, by design.

## Contract

`PageCandidateMode` is a discriminated choice of `Traditional`, `Cursor`, and `UnpagedCandidates`.
`OffsetParameterName`, `LimitParameterName`, `IncludeTotalCountSql`, and `OrderingMode` moved off
`PageDocumentIdQuerySpec` onto `Traditional`, which makes three invalid combinations
**unrepresentable** rather than rejected at runtime:

- a cursor page carrying a total count,
- a cursor page ordered by `ContentVersion` — a token anchored on the highest selected `DocumentId`
  is unsafe when that is not the page's ordering key, which is why ContentVersion-anchored cursors
  are deferred to DMS-1394,
- an unpaged candidate relation carrying a page size.

Five `QuerySqlParameterRole` members are added, appended after `Limit` so no existing ordinal shifts:
`CursorInclusiveMinimum`, `CursorInclusiveMaximum`, `PageSize`, `PartitionCount`, and
`MinimumPartitionSize`.

## Provider SQL

PostgreSQL:

```sql
SELECT r."DocumentId"
FROM "edfi"."StudentSchoolAssociation" r
WHERE
    (r."SchoolYear" >= @schoolYear)
    AND (r."Student_DocumentId" IS NOT NULL AND r."StudentUniqueId_Unified" = @studentUniqueId)
    AND (r."DocumentId" >= @cursorMin)
    AND (r."DocumentId" <= @cursorMax)
ORDER BY r."DocumentId" ASC
LIMIT @pageSize
;
```

SQL Server uses `TOP` rather than `OFFSET 0`, so the no-offset invariant is literal rather than
merely semantic:

```sql
SELECT TOP (@pageSize) r.[DocumentId]
FROM [edfi].[StudentSchoolAssociation] r
WHERE
    ...
    AND (r.[DocumentId] >= @cursorMin)
    AND (r.[DocumentId] <= @cursorMax)
ORDER BY r.[DocumentId] ASC
;
```

Cursor bounds are appended **after** every filter and authorization predicate. That placement is what
keeps the authorization fragment byte-identical across modes, which in turn makes the shared-candidate
guarantee checkable rather than asserted.

The unpaged candidate relation emits **no `ORDER BY`**. This is required, not cosmetic: its consumer
wraps it in a common table expression and applies its own row numbering, and SQL Server rejects
`ORDER BY` in a CTE that has no `TOP` or `OFFSET`. Both providers execute the wrapped form in the
integration probes.

## Planners

Both page keyset planners now consume the typed `CollectionPaging` choice from DMS-1383 and share one
planning core, so filters, change-version predicates, unified-alias rewrites, authorization, and
empty-page short-circuits are literally the same code across all three modes. A dedicated
`TryPlanCandidates` / `PlanCandidates` returns `CandidateQueryPlan` — a non-hydration result, because
the relation partition planning consumes is never hydrated.

Filter-name allocation reserves only the **active mode's** own names. Reserving another mode's names
would suffix a filter parameter over a collision the query does not actually have, which would move
the SQL of a mode with no stake in that name. Cross-mode parity is semantic — same field, predicate,
and bound value — not identical SQL parameter tokens.

`ChangeQueries/TrackedChangeQueryPlanner` is untouched: it builds its own SQL and never uses this
spec, so `/deletes` and `/keyChanges` cannot acquire cursor behavior by construction.

## Partition roles

`PartitionCount` and `MinimumPartitionSize` are defined, and their names (`number`,
`minimumPartitionSize`) are reserved and collision-validated now, so a resource filter cannot shadow
the partition window's parameters later. They are deliberately **not** placed in the plan's parameter
inventory: no emitted SQL binds them yet, and an inventory entry with no placeholder would fail
runtime parameter binding.

## Plan contract

The normalized plan DTO gains an optional `candidate_mode` discriminator, **omitted entirely for
traditional plans** so their canonical JSON and hashes stay byte-identical. Without it, an unpaged
candidate plan is indistinguishable from a traditional plan that lost its paging roles. The mode is
supplied by the caller that compiled the plan rather than inferred from the role inventory, because
inferring would defeat exactly that check.

Role-inventory validation is positional per declared mode, so mixtures, duplicates, wrong order, and
missing roles are all rejected.

## Testing

**Traditional textual gate.** The eight existing traditional golden SQL files and
`plan-contracts.manifest.json` are unchanged across both commits. Six new fixtures cover cursor and
unpaged SQL for both dialects.

**Compiler and planner units.** SQL shape, negative offset/row-number/count/`DISTINCT` assertions,
parameter inventory order, collision and distinctness validation including the reserved partition
names, and cross-mode identity of the shared candidate region. Planner tests prove traditional,
cursor, and unpaged planning yield identical filters, authorization, and filter parameter values for
the same request, and that traditional allocation stays unsuffixed for all five newly introduced
names while cursor and unpaged still disambiguate genuine collisions.

**Real-provider execution evidence.** New PostgreSQL and SQL Server 2025 probes execute the compiled
candidate SQL across every authorization shape the compiler emits — no further restrictions,
relationship EdOrg, two OR-combined relationship strategies, a self person subject, namespace,
single-step and multi-step custom view, and namespace combined with custom view — in all three modes.
The underlying authorization rows are seeded to duplicate, so a join-based plan would multiply
candidate rows and corrupt the row numbering and count partition boundaries derive from. Each
accessible `DocumentId` is asserted to appear exactly once, with no `DISTINCT` in the SQL.

`OwnershipBased` is absent on purpose: it is known but not enabled for GET-many and fails closed in
the authorization planner, so no candidate SQL is ever compiled for it. That path is already covered
by existing planner tests.

Cursor window, inverted range, page sizes 0 / 1 / maximum, and extreme `Int64` bounds are proven by
execution. Page size zero returning no rows is direct evidence that correctness does not depend on
the hydration batch's zero-limit short-circuit.

### Results

| Check | Result |
| --- | --- |
| `dotnet build src/dms/EdFi.DataManagementService.sln` | 0 errors |
| `Backend.Plans.Tests.Unit` | 1833 passed, 0 failed |
| `Backend.Tests.Unit` | 2744 passed, 0 failed |
| PostgreSQL 16 candidate integration | 9 passed |
| SQL Server 2025 (17.0.4065.4) candidate + shard guardrail | 12 passed |
| Golden fixtures | 14 passed |
| CSharpier | clean |

## Documentation

Includes the authorized design corrections removing DMS-1391 as a hard dependency of DMS-1385, so
repository truth matches the updated Jira descriptions. DMS-1391 still provides the traditional-paging
baseline the final performance gate compares against; it simply does not gate this story.

## Notes for reviewers

- Branch is based on `b67432fd4`; `main` has advanced three commits since, and a merge dry-run is
  clean. One of those, DMS-1282, also touches pagination validation, but at the Core request-validation
  boundary rather than the backend candidate compiler this PR changes.
- Known follow-up (non-blocking): `PostgresqlCandidateUniquenessTests` teardown drops its three probe
  tables but leaves seeded authorization rows in the shared configured database. A leased database, or
  cleaning only uniquely owned seed rows, would isolate it fully.
- Handoff to DMS-1386: `HydrationBatchBuilder.HasZeroLimit` keys only off the `Limit` role, so a
  `pageSize = 0` cursor page will not take the empty-keyset short-circuit once cursor execution lands.
  Correctness is unaffected — `LIMIT 0` and `TOP (0)` return no rows, proven on both providers — so
  this is a missed optimization for that story rather than a defect here.


src/dms/backend/EdFi.DataManagementService.Backend/DescriptorQueryPageKeysetPlanner.cs
```diff
@@ -27,11 +25,57 @@ public PageKeysetSpec.Query Plan(
...
+
+        return new PageKeysetSpec.Query(plannedCandidates.Plan, plannedCandidates.ParameterValues);
+    }
+
+    /// <summary>
+    /// Plans the unpaged, unordered descriptor candidate relation over the same
+    /// <c>dms.Descriptor</c> root, <c>ResourceKeyId</c> discriminator, filters, change-version window,
+    /// and authorization the paged modes use.
+    /// </summary>
+    public CandidateQueryPlan PlanCandidates(
```
bradbanister, line 64: **Medium:** `PlanCandidates` has no empty-result contract. Descriptor preprocessing can return `EmptyPage`, but the private planner throws for anything other than `Continue`; GET-many currently guards this while the future partition caller would have to duplicate it. Please mirror regular resources with `TryPlanCandidates(... out CandidateQueryPlan?, out string?)` and test invalid descriptor-id/empty preprocessing.

src/dms/backend/EdFi.DataManagementService.Backend/RelationalQueryPageKeysetPlanner.cs
```diff
@@ -74,29 +74,64 @@ public bool TryPlan(
...
         return plannedQuery is not null;
     }
 
-    private PageKeysetSpec.Query? PlanOrEmptyPage(
+    /// <summary>
+    /// Plans the unpaged, unordered candidate relation over the same filters, change-version window,
+    /// and authorization the paged modes use. Partition boundary planning consumes this relation and
+    /// applies its own row numbering, so the result is not a hydration keyset.
+    /// </summary>
+    public bool TryPlanCandidates(
```
bradbanister, line 99: **Medium:** This shares planning only after callers provide preprocessed query elements and a change-version range. DMS-1385 also owns shared Core validation/planning, but GET-many still performs it inside `ValidateQueryMiddleware` and `PartitionRequestValidator` defers filters/change versions. Please extract a common Core path, or explicitly move that requirement to DMS-1387, so equivalent raw requests cannot drift.

src/dms/backend/EdFi.DataManagementService.Backend.Plans.Tests.Unit/NormalizedPlanContractCodec.cs
```diff
@@ -136,11 +136,25 @@ public static ResourceReadPlanDto Encode(ExternalPlans.ResourceReadPlan plan)
...
+    /// <param name="plan">The compiled plan.</param>
+    /// <param name="candidateMode">
+    /// The mode the plan was compiled in, supplied by the caller that compiled it. Null declares
+    /// traditional paging. The mode is not inferred from the role inventory: inferring it would let a
+    /// traditional plan that lost its paging roles encode as a valid unpaged candidate plan, which is
+    /// the defect the discriminator exists to catch.
+    /// </param>
+    public static PageDocumentIdSqlPlanDto Encode(
+        ExternalPlans.PageDocumentIdSqlPlan plan,
+        PageCandidateModeDto? candidateMode = null
```
bradbanister, line 153: **Medium:** Optional `candidateMode = null` lets a cursor plan encode as traditional and fail only on decode. Decode then discards the discriminator, so encode/decode/re-encode loses `candidate_mode`; it also accepts total-count SQL for nontraditional modes. Make the mode explicit and preserved, reject counts for cursor/unpaged, and use a named legacy-traditional encoder if JSON omission must remain.

src/dms/backend/EdFi.DataManagementService.Backend.Postgresql.Tests.Integration/PostgresqlCandidateUniquenessTests.cs
```diff
@@ -0,0 +1,938 @@
...
+                "single-step custom view authorization",
+                CandidateProbeAuthorizationSpecs.SingleStepCustomView(
+                    _rootTable,
+                    _documentIdColumn,
+                    _customViewTable
+                )
+            ),
+        ];
+
+    /// <summary>
```
bradbanister, line 686: **Medium:** This correctly establishes descriptor custom-view checks as supported, but DMS-1386, DMS-1387, and DMS-1389 still say descriptors support only no-further/namespace. Update those stories; otherwise downstream cursor/partition implementation and its authorization matrix can omit a supported path.

src/dms/backend/EdFi.DataManagementService.Backend.Plans/PageDocumentIdQueryModels.cs
```diff
@@ -166,6 +166,104 @@ public enum PageOrderingMode
...
+    /// <param name="MinimumPartitionSizeParameterName">
+    /// The reserved minimum partition size parameter name.
+    /// </param>
+    /// <remarks>
+    /// Emits no <c>ORDER BY</c>: the consumer wraps this relation in a common table expression and
+    /// applies its own row numbering, and SQL Server rejects <c>ORDER BY</c> in a CTE that has no
+    /// <c>TOP</c> or <c>OFFSET</c>. The two parameter names are reserved and collision-validated so a
+    /// resource filter cannot shadow them, but no role is emitted until partition-window SQL binds them.
+    /// </remarks>
+    public sealed record UnpagedCandidates(
```
bradbanister, line 262: **Low:** These customizable partition parameter names never reach `CandidateQueryPlan`: the compiler marks them unbound and filters their roles, while the planner returns no mode or reserved names. Either carry them into DMS-1387 or make `UnpagedCandidates` a marker using canonical constants; the latter also removes dead `IsBound`/role plumbing and duplicate owned-name arrays.

src/dms/backend/EdFi.DataManagementService.Backend.Tests.Common/CandidateProbeAuthorizationSpecs.cs
```diff
@@ -0,0 +1,238 @@
...
+
+/// <summary>
+/// Builds one authorization specification per shape the page candidate compiler emits, for the
+/// candidate uniqueness probes.
+/// </summary>
+/// <remarks>
+/// The auth objects come from the production contracts rather than from hardcoded names, so the probe
+/// seeds and queries exactly the tables and columns the compiled SQL references.
+/// </remarks>
+internal static class CandidateProbeAuthorizationSpecs
```
bradbanister, line 19: **Low:** This file and `CandidateProbePlannerInputs` are identical to their MSSQL copies except for namespace: 442 helper lines are duplicated in each provider project. Both projects already reference shared common projects; move these helpers there to prevent the cross-provider parity data from drifting.

src/dms/backend/EdFi.DataManagementService.Backend.Postgresql.Tests.Integration/PostgresqlCandidateUniquenessTests.cs
```diff
@@ -0,0 +1,938 @@
...
+        await using var connection = await _dataSource.OpenConnectionAsync();
+
+        foreach (var statement in BuildSchemaStatements())
+        {
+            await using var command = new NpgsqlCommand(statement, connection);
+            await command.ExecuteNonQueryAsync();
+        }
+    }
+
+    [OneTimeTearDown]
```
bradbanister, line 150: **Low:** `OneTimeTearDown` blindly disposes `_dataSource` and `_database`. If setup fails before either assignment, cleanup can mask the root cause with a secondary null failure. Guard each field, following the more defensive MSSQL sibling.

src/dms/backend/EdFi.DataManagementService.Backend.Plans/PageDocumentIdSqlCompiler.cs
```diff
@@ -1537,6 +1738,23 @@ private static string GetTargetAlias(QueryPredicateTarget target)
     /// <param name="Operator">The comparison operator.</param>
     /// <param name="ParameterName">The bare SQL parameter name that supplies the value.</param>
     /// <param name="ScalarKind">Optional scalar-kind metadata for provider-specific comparison behavior.</param>
+    /// <summary>
```
bradbanister: **Low:** The `RewrittenPredicate` XML docs are interleaved with `ModeParameterName` here, so unrelated `<param>` tags attach to the wrong record and the intended record loses its documentation. Split/move the comment; this could also become a warnings-as-errors failure if XML validation is tightened.

src/dms/backend/EdFi.DataManagementService.Backend.Plans.Tests.Unit/ChangeVersionFilterAuthoritativeSqlShapeTests.cs
```diff
@@ -28,11 +28,8 @@ public class Given_ChangeVersionFilters_Over_Authoritative_MappingSets
         "../Fixtures/authoritative/sample/inputs/sample-api-schema-authoritative.json";
 
     private static readonly ChangeVersionRange _changeVersionRange = new(100L, 200L);
-    private static readonly PaginationParameters _paginationParameters = new(
-        Limit: 25,
-        Offset: 0,
-        TotalCount: true,
-        MaximumPageSize: 500
+    private static readonly CollectionPaging _paginationParameters = new CollectionPaging.Traditional(
```
bradbanister, line 30: **Low:** While changing this fixture, update the class summary near the top: it says descriptor filters use `dms.Document.ContentVersion`, but the assertions correctly verify `dms.Descriptor.ContentVersion`.

bradbanister:
Static review summary: **4 medium and 5 low findings**. The medium items are descriptor candidate planning throwing instead of representing an `EmptyPage`, the story-required Core filter/change-version sharing not yet being extracted, the normalized candidate-mode codec being able to mislabel/lose the mode and admit total-count SQL, and sibling stories omitting supported descriptor custom-view authorization. Low items cover the non-observable unpaged parameter-name customization, duplicated provider test helpers, PostgreSQL teardown masking setup failures, misplaced XML documentation, and a stale descriptor change-version test summary. Provider cursor SQL, authorization placement, and traditional SQL preservation otherwise looked correct. No local tests or builds were run per request.

src/dms/backend/EdFi.DataManagementService.Backend/DescriptorQueryPageKeysetPlanner.cs
simpat-adam: Pushing back on this one.

The throw here already mirrors the regular-resource planner: `PlanOrEmptyPage` throws `ArgumentException` on any non-`Continue` preprocessing result too (`RelationalQueryPageKeysetPlanner.cs:141-147`), so `TryPlanCandidates` does not turn an `EmptyPage` preprocessing result into `false` either. The two planners already agree on this input.

The relational `false`/null path exists for a different reason: relational value conversion happens *inside* planning (`TryConvertRawValue`, reached from `PlanRootColumnPredicate`), because only the planner knows the root column's scalar type. Descriptor preprocessing converts values up front and returns `RelationalQueryPreprocessingOutcome.EmptyPage` itself (`DescriptorQueryRequestPreprocessor.cs:58`) - the invalid descriptor id case you cite is exactly that. So a descriptor `TryPlanCandidates` could never return `false`; it would be a `bool` that is always `true`.

DMS-1387 has to handle preprocessing `EmptyPage` before planning for both families regardless, because it calls the preprocessor itself - the same guard `DescriptorReadHandler.cs:274` owns today. That is not a duplicated guard, it is the one guard each caller owns.

Happy to reconsider if you can point at a planning-time empty-candidate case on the descriptor path that I have missed.

src/dms/backend/EdFi.DataManagementService.Backend/RelationalQueryPageKeysetPlanner.cs
simpat-adam: Agreed that the story text was misleading, and I took the second option you offered: the Core-side requirement is now formally assigned to DMS-1387 (commit `1ef4c1ff3`).

What this story delivers is the backend half - both planners expose one candidate entry point, so any consumer supplying the same preprocessed filters, change-version window and authorization receives identical predicates, parameter names and bound values.

The Core half has no second consumer yet. There is no `/partitions` request pipeline to share with, and `PartitionRequestValidator` deliberately reserves neither resource-property filters nor change-version parameters for exactly that reason. Extracting `ValidateQueryMiddleware`'s filter parsing now would produce a seam with one caller and no second path to prove equivalence against, which is the gap you are describing rather than a fix for it.

`02-shared-candidate-planning.md` now states which layer is shared here, and `06-partition-pipeline-and-sql.md` carries the matching scope line: change-version parameters already have a standalone validator that pipeline can call, while resource-filter parsing must be extracted from `ValidateQueryMiddleware` rather than duplicated. Say the word if you would rather the extraction land in this story instead.

src/dms/backend/EdFi.DataManagementService.Backend.Plans.Tests.Unit/NormalizedPlanContractCodec.cs
simpat-adam: Fixed the two substantive parts in commit `329988345`.

- `Decode` now rejects `TotalCountSql` on any nontraditional plan. You are right that count validation was independent of the discriminator. The compiler only emits count SQL for `Traditional { IncludeTotalCountSql: true }` (`PageDocumentIdSqlCompiler.cs:100`), so the shape is unreachable from the compiler, but the codec exists to reject malformed contracts rather than only the ones today's compiler can produce. Two new tests cover cursor and unpaged.
- `Encode` now validates the declared mode against the plan's own role inventory instead of leaving that to `Decode`. That is the real hazard behind "only rejected later during decoding": canonical-JSON and hash callers never decode, so a plan declared as a mode it was not compiled in could be hashed into a golden with nothing complaining. This did break the pre-existing hash order-sensitivity test, which deliberately permuted a compiled plan's paging roles; it now permutes the encoded DTO, matching how the duplicate-name permutation assertions in the same fixture already build their inputs.

I did not return the mode from `Decode` or make the argument required. `PageDocumentIdSqlPlan` has no mode field - the mode is a compile input - so a decoded plan has nothing to re-declare, and `PageCandidateModeDto` deliberately has no `Traditional` member so traditional plans omit the discriminator and keep their pre-cursor hashes byte-identical. I did rename the two round-trip tests to say what they actually prove, with a note on why re-encoding is not part of the contract.

If you want mode-specific encoders regardless, that is a mechanical change across the page-plan call sites and I would rather do it as its own commit than fold it into this one.

src/dms/backend/EdFi.DataManagementService.Backend.Postgresql.Tests.Integration/PostgresqlCandidateUniquenessTests.cs
simpat-adam: Confirmed and fixed in commit `1ef4c1ff3`.

`BuildDescriptorQueryAuthorizationSpec` passes `CustomViewChecks` through with `Strategies: []`, so custom-view checks are compiled into the descriptor candidate relation today. Only relationship and ownership are genuinely unsupported for descriptors, because those need claim-EducationOrganization binding, which custom-view checks do not use at all.

Corrected the claim in all three siblings - `04-regular-resource-cursor-execution.md`, `06-partition-pipeline-and-sql.md`, and `08a-authorization-matrix.md` (both its scope line and its acceptance line) - each now naming relationship/ownership as the real exclusion instead of implying custom views are unsupported.

Flagging the consequence rather than burying it: this widens descriptor test scope for DMS-1386, DMS-1387 and DMS-1389. That is a scope change for those stories, not something this branch absorbs.

src/dms/backend/EdFi.DataManagementService.Backend.Plans/PageDocumentIdQueryModels.cs
simpat-adam: Pushing back on "effectively dead".

The supplied names are not unreachable. `BuildModeParameterNames` feeds them to `ValidateModeParameterNames` and `ValidateFilterParameterNamesDoNotCollideWithModeParameters`, so a custom name is honored exactly where it matters: a resource filter cannot shadow it. What is suppressed is inventory *emission*, through `IsBound: false`, and that is the documented invariant - an inventory entry with no placeholder in the emitted SQL would fail runtime binding.

Making unpaged mode a marker would not remove the `IsBound` plumbing either. The compiler still has to reserve the two canonical names while not emitting them, so the bound-versus-reserved distinction survives the change under a different name. And `PageCandidateModePlanning`'s array is the filter-name allocator's reservation list at a different layer, already spelled from `PageCandidateParameterNames` rather than repeating the values.

On carrying the mode and reserved names on `CandidateQueryPlan` for DMS-1387: that consumer constructs the mode itself through `PageCandidateModePlanning.ForUnpagedCandidates()`, so it already knows the names without the plan carrying them. I would rather add the field when DMS-1387 shows it needs one than add it speculatively now.

src/dms/backend/EdFi.DataManagementService.Backend.Tests.Common/CandidateProbeAuthorizationSpecs.cs
simpat-adam: Done in commit `4bfcf671d`. Both helpers now live in `Backend.Tests.Common`, which already references Backend, Backend.Plans, Backend.External and Core, and already grants `InternalsVisibleTo` to both provider integration projects - so they stay `internal`, neither provider project needed a new reference, and no production project's `InternalsVisibleTo` list changed. 442 duplicated lines removed.

Re-ran both probe fixtures against the moved helpers: 12 passed on PostgreSQL 16.3 and 12 passed on SQL Server 2025 (RTM-CU7).

src/dms/backend/EdFi.DataManagementService.Backend.Postgresql.Tests.Integration/PostgresqlCandidateUniquenessTests.cs
simpat-adam: Done in commit `4bfcf671d`. Both fields are guarded now, following the SQL Server sibling, with a comment recording that setup can fail between the two assignments and that an unguarded dispose would throw over the real failure.

src/dms/backend/EdFi.DataManagementService.Backend.Plans/PageDocumentIdSqlCompiler.cs
simpat-adam: Done in commit `1ef4c1ff3`. `RewrittenPredicate`'s summary and `<param>` tags are back on `RewrittenPredicate`, and `ModeParameterName` follows with its own summary and parameters.

src/dms/backend/EdFi.DataManagementService.Backend.Plans.Tests.Unit/ChangeVersionFilterAuthoritativeSqlShapeTests.cs
simpat-adam: Done in commit `1ef4c1ff3`. The summary now says descriptor change-version filters land on the `dms.Descriptor` root's mirrored `ContentVersion`, scoped by the project-qualified `ResourceKeyId` predicate - which is what the assertions verify. Also dropped the Jira key from that sentence, per the repo convention against issue keys in source comments.

simpat-adam:
Responded to all nine findings: six fixed, three pushed back with the code paths that support the pushback. Individual replies are in each thread; head is now `1ef4c1ff3`, rebased onto `main` @ `46c472196`.

**Fixed**

| # | Finding | Commit |
| --- | --- | --- |
| 3 | Plan-contract codec: `Decode` rejects count SQL for nontraditional modes, and `Encode` now validates the declared mode against the plan's roles instead of deferring to `Decode` | `329988345` |
| 6 | Duplicated provider probe helpers moved to `Backend.Tests.Common` (442 lines de-duplicated) | `4bfcf671d` |
| 7 | PostgreSQL one-time teardown guards both fields | `4bfcf671d` |
| 4 | Descriptor custom-view support corrected in `04`, `06` and `08a` | `1ef4c1ff3` |
| 2 | Core-side validation sharing formally assigned to DMS-1387, with this story's scope narrowed to the backend half it actually delivers | `1ef4c1ff3` |
| 8, 9 | Interleaved XML docs separated; change-version fixture summary corrected to the `dms.Descriptor` root | `1ef4c1ff3` |

Finding 3 is fixed in substance but not in the suggested shape: the mode is a compile input that `PageDocumentIdSqlPlan` does not carry, so `Decode` has nothing to return it with, and the optional-null default is what keeps every pre-cursor traditional hash byte-identical. Details in that thread.

**Pushed back**

- **1** - the descriptor throw already mirrors `RelationalQueryPageKeysetPlanner`, which throws on the same non-`Continue` input. The relational false path exists only because relational value conversion happens inside planning, while descriptor preprocessing converts up front and returns `EmptyPage` itself, so a descriptor `TryPlanCandidates` could never return `false`.
- **5** - the unpaged parameter names are not dead: they feed collision reservation, and only inventory emission is suppressed. `IsBound` survives any marker-style rewrite because reservation and emission still differ.

**Two follow-ups worth flagging**

- Correcting finding 4 widens descriptor test scope for DMS-1386, DMS-1387 and DMS-1389. That needs an owner decision on those stories, not a change here.
- Finding 2's resolution puts the `ValidateQueryMiddleware` filter-parsing extraction on DMS-1387 explicitly, so it cannot be duplicated there by accident.

**Verification**

- `EdFi.DataManagementService.sln` builds clean, 0 warnings with warnings-as-errors on.
- Plans unit 1835 passed; Backend unit 2745 passed.
- Candidate probe fixtures: 12 passed on PostgreSQL 16.3, 12 passed on SQL Server 2025 (RTM-CU7).
- The rebase was clean; `main` had advanced by one CDC-docs-only commit, so every code file is byte-identical to the tree those runs covered.

bradbanister (approved):

PR #1183, opened 2026-08-18
[DMS-1316] DocumentCache Health, Readiness, and Telemetry

bradbanister:


src/dms/backend/EdFi.DataManagementService.Backend/DocumentCacheStatusService.cs
```diff
@@ -0,0 +1,1284 @@
...
+                    providerObservationCancellationToken
+                )
+                .ConfigureAwait(false);
+
+            if (result.Outcome == DocumentCacheStatusCurrentSourceObservationOutcome.Cancelled)
+            {
+                callerCancellationToken.ThrowIfCancellationRequested();
+            }
+
+            ProviderObservationCancellationSource cancellationSource = GetFirstCancellationSource();
```
simpat-adam, line 370: [P1] A provider returning success after the observation or endpoint timeout is still treated as successful because the captured cancellation source is ignored for successful results. A non-cooperative or racing provider can therefore exceed the endpoint budget and report healthy data instead of `statusObservationTimeout` or `statusEndpointTimeout`. Please check the cancellation source before accepting a successful provider result.

src/dms/backend/EdFi.DataManagementService.Backend/DocumentCacheStatusService.cs
```diff
@@ -0,0 +1,1284 @@
...
+            return new();
+        }
+
+        return new(
+            targetHealth
+                .PoisonTraversal.SuppressedDocumentIds.Select(
+                    documentId => new DocumentCacheStatusPoisonTraversalDiagnosticEvent(
+                        documentId,
+                        targetHealth.ObservedAt,
+                        DocumentCacheStatusPoisonTraversalDiagnosticCategory.SkippedUntilRetry,
```
simpat-adam: [P2] These poison-traversal entries are fabricated from the latest suppressed-ID snapshot: every event receives the current target timestamp and `skippedUntilRetry`. Events are not retained across snapshots, and the required `retryScheduled` and `pageCapacityExhausted` categories can never be emitted. Please retain the actual bounded traversal events and their original timestamps/categories.

src/dms/tests/EdFi.DataManagementService.Tests.Integration/Tests/DocumentCache/Given_DocumentCacheStatusEndpointAuthorization.cs
```diff
@@ -0,0 +1,433 @@
...
+            builder.ConfigureServices(services =>
+            {
+                AddEssentialMocks(services);
+                services.Replace(
+                    ServiceDescriptor.Singleton<IJwtValidationService>(
+                        jwtValidationService
+                            ?? new ScriptedJwtValidationService(new Dictionary<string, ClaimsPrincipal?>())
+                    )
+                );
+                services.Replace(
```
simpat-adam, line 85: [P2] The only API integration coverage replaces `IDocumentCacheStatusService` with a scripted fake, so it never exercises production provider observations, multi-target isolation, or diagnostic sanitization through the endpoint. DMS-1316 explicitly requires API integration coverage for those behaviors; please add a test that runs the real status composition pipeline.

src/dms/backend/EdFi.DataManagementService.Backend.Postgresql.Tests.Integration/PostgresqlDocumentCacheStatusCurrentSourceObserverTests.cs
```diff
@@ -0,0 +1,432 @@
...
+        );
+
+        result.Outcome.Should().Be(DocumentCacheStatusCurrentSourceObservationOutcome.Failed);
+        result.DurableObservedAt.Should().BeNull();
+        result.LifecycleState.Should().BeNull();
+        result.QueuePresence.Should().BeNull();
+    }
+
+    [Test]
+    public async Task It_uses_the_ordered_single_row_projection_work_index_and_not_source_or_cache_scans()
```
simpat-adam, line 265: [P2] This test uses only two documents and verifies plan shape; PostgreSQL also forces sequential scans off later in the fixture. It does not provide the required scale evidence that health/caught-up/oldest-work polling stays bounded as document cardinality grows. Please add representative high-cardinality coverage for both providers.

reference/design/backend-redesign/epics/18-document-cache/09-documentcache-administration-cli.md
```diff
@@ -0,0 +1,111 @@
+---
+jira: DMS-1428
```
simpat-adam, line 2: [P2] This PR introduces a separate DMS-1428 administration-CLI story and wires it into the epic/dependency documents. That planning artifact is outside the DMS-1316 health/readiness/telemetry implementation scope and should be moved to its own change.

.gitignore
```diff
@@ -70,6 +70,10 @@ eng/docker-compose/.env.perf-hostpg
 # Local data-load credentials
 /dataload-creds.json
 
+# Local task/progress scratch files
```
simpat-adam: [P2] These root-level `tasks.json` and `progress.txt` ignore rules already exist later in this file and are unrelated to DMS-1316. Please remove this duplicate scope-hygiene churn.

src/dms/core/EdFi.DataManagementService.Core/Middleware/JwtRoleAuthenticationMiddleware.cs
```diff
@@ -90,7 +89,7 @@ public async Task Execute(RequestInfo requestInfo, Func<Task> next)
         // Check for required role if configured
         if (!string.IsNullOrEmpty(_options.ClientRole))
         {
-            var hasRequiredRole = principal.HasClaim(ClaimTypes.Role, _options.ClientRole);
+            var hasRequiredRole = principal.IsInRole(_options.ClientRole);
```
simpat-adam: **Blocker.** `IsInRole` resolves against `ClaimsIdentity.RoleClaimType`; the `HasClaim(ClaimTypes.Role, ...)` it replaces did not. `JwtValidationService` sets `RoleClaimType` from `JwtAuthentication:RoleClaimType`, whose shipped default is `"role"` (`src/dms/frontend/.../appsettings.json`) — but the Configuration Service issues the role claim as `http://schemas.microsoft.com/ws/2008/06/identity/claims/role` (`src/config/frontend/.../appsettings.json`, consumed by `JwtTokenGenerator`; same URI on the Keycloak path in `KeycloakClientRepository`). The old explicit lookup matched regardless of `RoleClaimType` and papered over that mismatch. `IsInRole` does not: running DMS + CMS from source with default appsettings now returns 403 on every authenticated data-API request.

`MapInboundClaims = false` in `JwtValidationService` compounds it — `DefaultInboundClaimTypeMap` mapped both `role` and `roles` to `ClaimTypes.Role`, so IdPs emitting the plural `roles` claim (Entra app roles) also stop matching.

Neither gate catches this: every `eng/docker-compose/.env*` sets `DMS_JWT_ROLE_CLAIM_TYPE` to the long URI, and the unit tests build the principal with `new ClaimsIdentity(claims, "Test")`, whose default `RoleClaimType` is already `ClaimTypes.Role`. Suggest keeping the explicit claim-type lookup (or accepting both the configured type and `ClaimTypes.Role`), and adding a test that constructs the identity with the shipped default `RoleClaimType`.

src/dms/backend/EdFi.DataManagementService.Backend/DocumentCacheEnqueueTelemetry.cs
```diff
@@ -0,0 +1,681 @@
...
+        )
+        {
+            category = DocumentCacheEnqueueFailureCategory.StateMissingOrInvalid;
+            return true;
+        }
+
+        bool hasEnqueueArtifactEvidence = HasEnqueueArtifactEvidence(classificationText);
+
+        if (
+            hasEnqueueArtifactEvidence && providerCommandTimeoutClassifier.IsProviderCommandTimeout(exception)
```
simpat-adam: This branch and `ProviderUnavailable` (line 597) both AND `hasEnqueueArtifactEvidence`, which is derived only from `exception.Message` (line 654). No provider produces a message that is both a timeout/connection signal *and* a name of the enqueue artifact — the repo's own classifiers match on codes precisely because the text carries no object name (`PostgresqlDocumentCacheProviderCommandTimeoutClassifier` → SQLSTATE `57014`; the MSSQL one → `Number == -2`, text `"Execution Timeout Expired."`). A real enqueue timeout or connection loss therefore falls through to `return false` at line 629: no metric, no retained status event.

The covering tests fabricate composite messages (`DocumentCacheEnqueueTelemetryTests.cs:612,625,677,690`) that no provider emits, which is why this passes. Suggest dropping the artifact-evidence conjunct for the code-based classifiers, and asserting against a message shaped like a real provider exception.

src/dms/core/EdFi.DataManagementService.Core/Configuration/DocumentCacheOptions.cs
```diff
@@ -174,19 +227,57 @@ private static bool CanBeSentAsTenantHeader(string value)
...
+        bool explicitStatusSection = _configuration
+            .AsEnumerable()
+            .Any(setting => string.Equals(setting.Key, StatusSectionName, StringComparison.Ordinal));
+
+        if (!explicitStatusSection)
+        {
+            return;
+        }
+
+        string failure = statusSection.Value is null
```
simpat-adam: `"Status": {}` in appsettings refuses host start. The JSON configuration provider surfaces an empty object as a null-valued key with no children, so `GetChildren().Any()` is false, `explicitStatusSection` is true, and this reports *"Status section must not be null."* With `ValidateOnStart` (`WebApplicationBuilderExtensions`) the host then refuses to boot on a legitimate defaults-only override. The `null` and scalar cases are pinned by tests; `{}` is not.

src/dms/backend/EdFi.DataManagementService.Backend/DocumentCacheEnqueueTelemetry.cs
```diff
@@ -0,0 +1,681 @@
...
+
+internal static class DocumentCacheEnqueueFailureClassifier
+{
+    private const string StateMissingMessage =
+        "dms.DocumentCacheState singleton row is missing or unreadable for projection enqueue.";
+    private const string UnsupportedLifecycleValueMessage =
+        "dms.DocumentCacheState.ProjectionLifecycleState has unsupported value";
+    private const string DocumentProjectionWorkName = "DocumentProjectionWork";
+    private const string EnqueueFunctionPrefix = "TF_Document_EnqueueProjection";
+    private const string EnqueueTriggerName = "TR_Document_EnqueueProjectionWork";
```
simpat-adam: This is the SQL Server trigger name only — PostgreSQL generates `TR_Document_EnqueueProjectionInsert` / `TR_Document_EnqueueProjectionUpdate` (confirmed in the DDL goldens). A PG failure message naming the trigger satisfies neither `HasEnqueueArtifactEvidence` nor `EnqueueTriggerUnavailable`. The sibling function constant is already a prefix; suggest the same here (`TR_Document_EnqueueProjection`). The one trigger-name test uses the MSSQL name, so the gap is invisible.

src/dms/backend/EdFi.DataManagementService.Backend/DocumentCacheEnqueueTelemetry.cs
```diff
@@ -0,0 +1,681 @@
...
+        description: "DocumentCache canonical write enqueue failures."
+    );
+
+    public void RecordSuccess(DocumentCacheEnqueueTelemetryContext context)
+    {
+        ArgumentNullException.ThrowIfNull(context);
+
+        _successCounter.Add(1, ToTags(context, category: null));
+
+        _logger.LogInformation(
```
simpat-adam: Information level fires once per *successful* write. `RecordSuccess` runs on every resource/descriptor write once lifecycle is not `Disabled`, and the default minimum level is Information — so this is request-rate-proportional log volume. Every sibling telemetry class logs its per-operation event at Debug (`DocumentCacheReadTelemetry`, `DocumentCacheProjectionTelemetry`), and there are no other `LogInformation` calls in the write-path telemetry files. Suggest Debug.

src/dms/backend/EdFi.DataManagementService.Backend/DocumentCacheEnqueueTelemetry.cs
```diff
@@ -0,0 +1,681 @@
...
+        )
+            ? targetKey
+            : null;
+    }
+
+    private static DocumentCacheTargetObservation? TryGetCurrentTarget(
+        IDocumentCacheTargetRegistry? targetRegistry,
+        DocumentCacheTargetKey targetKey
+    ) =>
+        targetRegistry?.CurrentSnapshot.Targets.SingleOrDefault(target => target.TargetKey.Equals(targetKey));
```
simpat-adam: `SingleOrDefault` throws on a duplicate target key, where the registry's own accessor for this same lookup uses `FirstOrDefault` (`DocumentCacheTargetRegistry.CurrentSnapshot.GetTarget`). It matters because the telemetry call runs *after* `CommitAsync`, inside `DescriptorWriteHandler`'s `catch { TryRollbackAsync; throw; }` — a throw here would report a committed write as a failure and attempt rollback on a committed transaction. Unreachable through the production registry, but reachable via the test-double registries, and telemetry should not be able to throw at all. Suggest reusing `GetTarget(targetKey)`.

src/dms/backend/EdFi.DataManagementService.Backend/DocumentCacheStatusTelemetry.cs
```diff
@@ -0,0 +1,259 @@
...
+            new("provider", providerToken.Value),
+            new("target", DocumentCacheTelemetryTargetLabel.FromTargetKey(targetKey)),
+            new("outcome", DocumentCacheTelemetryLabel.LowerCamel(outcome)),
+            new("reason", DocumentCacheTelemetryLabel.LowerCamel(reason)),
+        ];
+
+        double durationSeconds = ClampToNonNegativeSeconds(duration);
+        _providerObservationDuration.Record(durationSeconds, durationTags);
+
+        if (oldestWorkAgeSeconds is not null && lifecycleState is not null)
```
simpat-adam: The `&& lifecycleState is not null` conjunct is unsatisfiable for the `stateMissingOrInvalid` outcome, so its oldest-work age is never recorded. `DocumentCacheStatusCurrentSourceObservation` forbids that outcome from carrying a lifecycle but explicitly allows it to carry oldest-work facts — and a test pins exactly that shape (`DocumentCacheStatusServiceTests`, age 300). The design doc's recording condition ("observes work and returns an oldest-work age") is met; only the `lifecycle` tag is genuinely absent. Suggest recording the age whenever it is present and omitting/defaulting the lifecycle tag. The `without_oldest_work_age` test nulls both inputs, so it cannot distinguish the two guards.

src/dms/backend/EdFi.DataManagementService.Backend/DocumentCacheStatusService.cs
```diff
@@ -0,0 +1,1284 @@
...
+            targetObservation.ResolutionState switch
+            {
+                DocumentCacheTargetResolutionState.Resolved => DocumentCacheStatusResolutionStatus.Resolved,
+                DocumentCacheTargetResolutionState.Unresolved =>
+                    DocumentCacheStatusResolutionStatus.Unresolved,
+                _ => DocumentCacheStatusResolutionStatus.Unknown,
+            },
+            targetObservation.ResolutionState == DocumentCacheTargetResolutionState.Resolved
+                ? DocumentCacheStatusResolutionReason.None
+                : ToResolutionReason(latestDiagnostic?.Category),
```
simpat-adam, line 961: `ToResolutionReason` has no `null` case and falls through to `_ => TargetNotFound`, so a target still in its seeded boot state reports `resolution.reason: "targetNotFound"`. `DocumentCacheTargetRegistry` seeds every configured target as `Configured(...)` with empty `Diagnostics`, and the status endpoint never refreshes — so any request before the first successful refresh is indistinguishable from a target genuinely absent from CMS, which also maps to `TargetNotFound`. `ResolutionReason.None` is the honest value here. No test drives a `Configured` target through `GetStatusAsync`.

src/dms/backend/EdFi.DataManagementService.Backend.Mssql.Tests.Integration/MssqlDocumentCacheStatusCurrentSourceObserverTests.cs
```diff
@@ -0,0 +1,485 @@
...
+            UPDATE STATISTICS [dms].[DocumentProjectionWork];
+            """
+        );
+
+        string plan = await ReadStatisticsXmlPlanAsync(
+            MssqlDocumentCacheStatusCurrentSourceObserver.StatusObservationSql
+        );
+
+        plan.Should().Contain("IX_DocumentProjectionWork_FirstEnqueuedAt_DocumentId");
+        plan.Should().NotContain("[dms].[Document]");
```
simpat-adam: This guard cannot fail. Showplan XML names objects as separate attributes (`Schema="[dms]" Table="[Document]"`), never the concatenated `[dms].[Document]` — the only place that string can appear is the `StatementText` echo, so this asserts on statement text rather than on the plan. A plan that reaches `dms.Document` other than by naming it literally would pass. Same for the `[dms].[DocumentCache]` assertion. Suggest matching the `Schema=` / `Table=` attribute pair, or parsing the XML.

src/dms/backend/EdFi.DataManagementService.Backend.Postgresql.Tests.Integration/PostgresqlDocumentCacheStatusCurrentSourceObserverTests.cs
```diff
@@ -0,0 +1,432 @@
...
+            oldestDocumentId,
+            requiredContentVersion: 10,
+            FirstEnqueuedAt,
+            FirstEnqueuedAt.AddSeconds(5)
+        );
+
+        string plan = await ExplainStatusObservationSqlAsync();
+
+        plan.Should().Contain("IX_DocumentProjectionWork_FirstEnqueuedAt_DocumentId");
+        plan.Should().NotContain(" on \"Document\" ");
```
simpat-adam: The trailing space makes this guard conditional on an alias. Plan lines are joined with `Environment.NewLine`, so a node that renders the relation last on its line (`... on "Document"`, no alias) is followed by the newline and never matches this substring. Suggest dropping the trailing space, or matching the relation name on a word boundary.

simpat-adam (commented):
Review of the document-cache status/telemetry work in this PR. One blocker on the JWT role check, plus several classification paths that are unreachable and a few test guards that cannot fail. Details inline.

simpat-adam (approved):

PR #1192, opened 2026-08-21
[DMS-1389] Cursor and partition authorization matrix

simpat-adam:
## What

Adds the cursor/partition authorization matrix: integration coverage proving that for the same principal, filters, and seeded data, a full cursor walk and the union of the `/partitions` ranges resolve **exactly** the accessible candidate set, and that a client cannot widen that set by editing the range inside a `pageToken`.

Ten rows per provider, run against real PostgreSQL and real SQL Server 2025 — 20 tests, test-only, no production code touched.

| Row | Carrier | Strategy | What it pins |
|---|---|---|---|
| R-NF | regular | no-further | Both surfaces agree when nothing filters |
| R-REL | regular | `RelationshipsWithEdOrgsOnly` | EdOrg hierarchy reach |
| R-NS | regular | `NamespaceBased` | Prefix filtering |
| R-VIEW | regular | custom view | A real `auth` view over a resource column |
| R-MULTI | regular | relationship, two EdOrg claims | One row per document is a property of the *plan*, not the data |
| D-NF | descriptor | no-further | |
| D-NS | descriptor | `NamespaceBased` | |
| D-VIEW | descriptor | custom view | Descriptor custom views reach both surfaces |
| E-REG / E-DESC | both | authorization admits nothing | Empty partition array, no `Next-Page-Token` |

## Why

The two surfaces compile the *same* authorized candidate relation but reach it differently: pages select a bounded range from it, while boundaries rank and count the whole of it. A defect that authorized one and not the other would leave every single-surface test green. So every row compares the two against an expected set derived from the **seed**, never from either surface under test.

Three design choices carry most of the evidential weight:

**The host maximum page size is lowered to 2.** The mandatory minimum partition size is `MaximumPageSize * 5`, so at the deployed 500 no collection seeded over HTTP could ever be cut into more than one partition, and every union and disjointness assertion would pass without crossing a single boundary. At 2 the minimum is ten candidate rows, which the 21-document accessible sets clear, so three requested partitions really produce three.

**The inaccessible controls are interleaved, not clustered.** Every fourth document is the control, so boundary rows fall in different places for the authorized and unauthorized sets. Boundaries computed *before* authorization cannot coincidentally agree with a correctly authorized walk.

**`totalCount` is the duplicate detector.** It is compiled from the same candidate relation in traditional mode, so it is a cross-check rather than an independent oracle — and it is the most sensitive one available for a relation that started returning a document twice. Walk results are asserted as a *list* with `OnlyHaveUniqueItems` before any set conversion, so a repeat fails rather than being normalized away.

## Changes

Four new files under `src/dms/tests/EdFi.DataManagementService.Tests.Integration/`:

- **`Scenarios/CursorPartitionAuthorizationMatrixScenario.cs`** — the assertions. `RunMatrixAsync` is shared by every row: total-count cross-check, full cursor walk, partition ascent/abutment/unbounded-tail, start-id accessibility, pairwise disjointness, union equivalence, forged-range probes, and filter agreement across both surfaces.
- **`Scenarios/CursorPartitionAuthorizationMatrixSupport.cs`** — seeding, principals, and the database-side auth views. Kept beside the assertions so one file states what is proven and the other what it is proven over.
- **`Tests/Postgresql/…`** and **`Tests/Mssql/…`** — the provider bindings. Each row's class adds only its principal and strategy; the shared base supplies the fixture, the real authorization middleware (`BypassAuthorization => false`), and the lowered page size.

### Negative cases

- **Widened forged token.** A maximally widened range names every identifier the store could hold, so authorization alone keeps the inaccessible documents out. The test asserts the forged token is byte-identical to what widening a *real* emitted partition token produces, then walks it to its terminal empty response — one clean page would not be enough, since an interleaved inaccessible document could wait on a later page.
- **Match-nothing ranges.** A range narrowed onto a single row the caller may not read, an inverted range, and both extreme `Int64` ranges each return nothing with no continuation.
- **Partition starting ids** are asserted to be actual accessible candidate ids and never inaccessible ones — a start taken from an inaccessible row would itself disclose it.

### Authorization is exercised through the real path

The custom-view rows provision a genuine `auth` view on the leased database exactly as an administrator would, because the view is a deployment artifact DMS never creates. Both select over a real resource column rather than a hard-coded identifier list, and the root table's schema is *resolved* from the catalog rather than assumed, so a schema-naming change fails with a message naming the table instead of silently creating a view over nothing.

For the regular-resource view every field another strategy could authorize on is seeded so that it would admit the whole collection — each namespace sits under the caller's authorized prefix, every reference points at the authorized school — so the view is demonstrably the only thing excluding the denied documents.

### Coverage boundaries, recorded rather than implied

- `OwnershipBased` is excluded for every operation including GET-many: it is recognized but not enabled, so the request fails closed before a candidate relation exists.
- **Relationship strategies are the descriptor exclusion**, because descriptors expose no education-organization or person securable elements to range over.
- Descriptor **custom views are in scope**. `PrepareDescriptorQueryAsync` and `PrepareDescriptorPartitionAsync` both call the same `BuildDescriptorQueryAuthorizationSpec`, so descriptor custom-view checks are compiled into the candidate relation behind *both* the page walk and the partition boundaries — precisely the agreement this matrix exists to prove. Omitting it would have left that compiled path uncovered on both providers.

## Verification

- CSharpier clean. Integration project builds **0 warnings / 0 errors**.
- Focused matrix: **20 passed, 0 failed, 0 skipped** against live PostgreSQL and SQL Server 2025, with both admin connections passed explicitly and both credential sets read from the running containers rather than assumed. Zero skips is the evidence the PostgreSQL lane actually executed rather than silently skipping on an absent connection string.
- **Non-vacuity, not just green.** The `D-VIEW` row was mutation-checked: widening its auth view to `WHERE 1 = 1` fails with `Expected totalCount to be 21 … but found 28`. The mutation was reverted and the suite re-run green, so the assertions are provably bound to the auth view rather than passing incidentally.
- Rebased onto current `main` and **re-verified there at 20/0/0**. This mattered: `main` gained DMS-1331, which anchors relationship-authorization paging subqueries on root-row columns and rewrites the very subquery shape the `R-REL` and `R-MULTI` rows read through. The rebase was conflict-free, but that proves nothing about files a branch *adds*, so it was audited separately — `main` touched neither `ApiIntegrationTestBase` nor `Core/Paging`, and the passing run on the new base covers the behavioral half.

## Notes for review

- Test-only. No production code, public contract, or OpenAPI surface changes.
- The custom-view rows create their view without a defensive drop, relying on a per-row leased database, consistent with the surrounding fixtures. It would only matter if these were ever pointed at a shared database.
- Full-project integration runs are left to CI; local full-suite PostgreSQL runs are known to exhaust shared memory at stock container settings.

### DIFF
--- a/src/dms/tests/EdFi.DataManagementService.Tests.Integration/Scenarios/CursorPartitionAuthorizationMatrixSupport.cs
+++ b/src/dms/tests/EdFi.DataManagementService.Tests.Integration/Scenarios/CursorPartitionAuthorizationMatrixSupport.cs
@@ -1,0 +1,483 @@
+// SPDX-License-Identifier: Apache-2.0
+// Licensed to the Ed-Fi Alliance under one or more agreements.
+// The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+// See the LICENSE and NOTICES files in the project root for more information.
+
+using System.Data.Common;
+using System.Globalization;
+using System.Net;
+using System.Text.Json.Nodes;
+using EdFi.DataManagementService.Backend.Tests.Common;
+using EdFi.DataManagementService.Core.External.Security;
+using EdFi.DataManagementService.Core.Security;
+using EdFi.DataManagementService.Tests.Integration.Fixtures;
+using FluentAssertions;
+using static EdFi.DataManagementService.Tests.Integration.Scenarios.PeopleRelationshipGetManyScenarioHelpers;
+
+namespace EdFi.DataManagementService.Tests.Integration.Scenarios;
+
+/// <summary>
+/// Seeding, principal configuration, and database-side support for the cursor/partition authorization
+/// matrix. Kept beside the assertions rather than inside them so the scenario file states what is proven
+/// and this file states what it is proven over.
+/// </summary>
+internal static class CursorPartitionAuthorizationMatrixSupport
+{
+    // The shared authorization fixture already publishes the principal, the two schools, and the prefix
+    // pair its DDL and provider suites are built around; restating them here would let this matrix drift
+    // from the fixture it seeds.
+    public const long ClaimEducationOrganizationId =
+        RelationshipAuthorizationCrudTestSupport.ClaimEducationOrganizationId;
+
+    /// <summary>
+    /// A second education organization claim held alongside the first. Both reach the authorized school
+    /// through the hierarchy table, so every accessible candidate matches more than one authorization row.
+    /// </summary>
+    public const long SecondClaimEducationOrganizationId = ClaimEducationOrganizationId + 1;
+
+    public const long AuthorizedSchoolId = RelationshipAuthorizationCrudTestSupport.AuthorizedSchoolId;
+    public const long UnauthorizedSchoolId = RelationshipAuthorizationCrudTestSupport.UnauthorizedSchoolId;
+
+    public const string AuthorizedNamespacePrefix =
+        RelationshipAuthorizationCrudTestSupport.AuthorizedNamespacePrefix;
+    public const string UnauthorizedNamespacePrefix =
+        RelationshipAuthorizationCrudTestSupport.UnauthorizedNamespacePrefix;
+
+    /// <summary>
+    /// A prefix the caller holds that no seeded value starts with. Distinct from holding no prefixes at
+    /// all, which is the no-prefixes-configured 403 preflight terminal rather than an authorized-but-empty
+    /// candidate set.
+    /// </summary>
+    public const string UnmatchedNamespacePrefix = "uri://nothing.example/";
+
+    /// <summary>
+    /// Obeys the "{BasisResource}With..." convention the strategy classifier requires, so the basis
+    /// resource resolves to the seeded carrier and the check filters its root DocumentId by membership in
+    /// the matching auth view.
+    /// </summary>
+    public const string CustomViewStrategyName =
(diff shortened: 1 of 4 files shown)


src/dms/tests/EdFi.DataManagementService.Tests.Integration/Scenarios/CursorPartitionAuthorizationMatrixSupport.cs
```diff
@@ -0,0 +1,536 @@
...
+    public static IClaimSetProvider CreateRelationshipReadClaimSetProvider(FixtureContext fixture) =>
+        CreateClaimSetProvider(
+            fixture,
+            [
+                new RelationshipReadResource(
+                    AuthzProjectName,
+                    RelationshipAuthorizationCrudTestSupport.NamespaceResourceName
+                ),
+            ],
+            AuthorizationStrategyNameConstants.RelationshipsWithEdOrgsOnly
```
bradbanister, line 206: **Medium:** This is the only regular-resource relationship strategy configured by the matrix, so the story's "every supported authorization strategy" requirement is not covered. In particular, the inverted EdOrg path and the people/student strategy families use distinct candidate-plan branches and could regress while this suite remains green. Add an inverted EdOrg fixture with reverse hierarchy data and representative eligible person/student carriers for the remaining supported relationship strategies.

bradbanister:
Static review: one medium-severity coverage gap in the authorization matrix.

bradbanister (approved):

PR #1204, opened 2026-08-26
[DMS-1278] Accept the raw claims document on POST /management/upload-claims

samuellugo-ship-it:
## Problem

`GET /management/current-claims` returns the raw claims document:

```json
{ "claimSets": [...], "claimsHierarchy": [...] }
```

but `POST /management/upload-claims` bound a `UploadClaimsRequest(JsonNode Claims)` wrapper, so it
required the document nested under a `claims` property. Downloading the current claims and uploading
the file unchanged — the workflow the two endpoints exist to support — returned HTTP 400
`"Claims JSON is required"`.

The wrapper lived at exactly one place, the minimal-API handler parameter. Everything below it already
consumed the raw document: `IClaimsUploadService.UploadClaimsAsync(JsonNode)`, `ClaimsUploadService`'s
`claimSets`/`claimsHierarchy` structure check, and `JsonSchemaForClaims.json`, whose root requires
those two properties with `additionalProperties: false`.

## Change

Bind the upload request body as `JsonNode` and declare that in the OpenAPI request metadata, making the
two endpoints share one canonical contract. The GET response can now be posted back verbatim.

`JsonNode` rather than `JsonObject` keeps a non-object body on the service's existing
`"Claims JSON must be an object"` path, and the parameter stays non-nullable so an absent or literal
`null` body keeps the framework's missing-body 400.

A body missing both canonical properties — an empty object, or the previously required wrapper — is
reported by the existing structure-failure mapping as missing `claimSets` and `claimsHierarchy`, so a
caller on the old shape is told exactly what the body must contain. No new code path, and no
sniffing branch that would leave two live request shapes.

Authorization, the `DangerouslyEnableUnrestrictedClaimsLoading` gate, `X-Reload-Id`, reload behavior,
and the deny-by-default failure mapping are unchanged.

The in-repo callers that sent the wrapper move to the raw shape: the CMS E2E claims scenarios, the DMS
E2E CMS claim-set upload helper, and the REST Client sync demo.

## Tests

- Focused unit coverage for the round trip (the GET response body posted back byte for byte, validated
  by the real `ClaimsValidator` so the document is also proven schema-valid), for rejection of the
  wrapper shape, and for the empty-object, non-object, malformed, absent, and literal-`null` bodies.
- A new CMS E2E scenario uploads the live `current-claims` response unchanged and verifies the claims
  are unchanged afterward. Tagged `@MssqlRepresentative` so the SQL Server lane also exercises
  re-uploading a whole document over persisted claim sets.
- The DMS E2E `ClaimSetSynchronization` scenarios cover the changed upload helper.

## Documentation

`docs/CLAIMS-LOADING-GUIDE.md` already showed the canonical body, so it needed no correction, but it
never said the `current-claims` response *is* that document. It now does, with a download-then-upload
example. Two adjacent inaccuracies found while editing that section are also fixed: the upload and
reload sections claimed a reload ID *header* (both return it in the response body; only
`GET /management/current-claims` sets `X-Reload-Id`), and the claims examples used a `resources`
collection with claim-level `actions`, neither of which the enforced schema allows — they now use
nested `claims` with `claimSets`/`authorizationStrategyOverrides`, validated against
`JsonSchemaForClaims.json`.

## Notes to consider

- **Public documentation is a separate repo.** The ticket's acceptance criteria name
  `odsApi_versioned_docs/version-8.0/how-to-guides/how-to-create-and-manage-api-security-metadata` and
  `.../platform-dev-guide/utilities/cms-hierarchy-tool.md`. Those files are not in this repository; they
  live in `ed-fi-alliance-oss/ed-fi-alliance-oss.github.io`. Both published pages still instruct the
  wrapper shape, and both need two corrections: the body is the document itself, and the hierarchy
  examples use nested `claims` rather than `resources`. Reviewed replacement text is available for the
  docs team on request — this PR does not touch that repository.
- **Breaking change to an undocumented shape.** The wrapper is no longer accepted. The endpoint is
  gated behind `DangerouslyEnableUnrestrictedClaimsLoading`, and this repository's own guide already
  documented the raw shape, so the raw-only contract is the one the ticket asks for rather than a
  dual-shape compatibility path. Anyone running the updated REST Client demo or DMS E2E helper against
  an older published configuration-service image will get a 400 until that image carries this change.
- `UploadClaimsRequest` is intentionally left in place; it is public backend surface and its removal is
  not needed for this fix.

### DIFF
--- a/docs/CLAIMS-LOADING-GUIDE.md
+++ b/docs/CLAIMS-LOADING-GUIDE.md
@@ -316,13 +316,14 @@ curl -X POST -H "Content-Type: application/json" \
 ```
 
 **Important Notes:**
-- The uploaded structure completely replaces the current claims
+- The uploaded document replaces the claims hierarchy and every non-system-reserved claim set;
+  system-reserved claim sets are preserved whether or not the document repeats them
 - Must include both `claimSets` and `claimsHierarchy` sections
 - The request body is the claims document itself. It is not nested under a `claims` property; a body
   shaped that way is rejected with HTTP 400 naming the missing `claimSets` and `claimsHierarchy`
   properties.
 - Subject to JSON Schema validation
-- Returns a reload ID header for tracking
+- Returns the new reload ID in the response body: `{ "success": true, "reloadId": "<guid>" }`
 
 ### Reload Claims from Fragments
 
@@ -333,10 +334,14 @@ curl -X POST -H "Authorization: Bearer $TOKEN" \
   http://localhost:8081/management/reload-claims
 ```
 
-Response includes:
-- X-Reload-Id header with unique identifier
-- Status of reload operation
-- Any validation errors encountered
+On success the response body carries the status and the new reload ID:
+
+```json
+{ "success": true, "reloadId": "<guid>" }
+```
+
+A failed reload is returned as an Ed-Fi problem-details error response, not as an error list on a
+success body.
 
 ### Get Current Claims
 
@@ -347,8 +352,10 @@ curl -H "Authorization: Bearer $TOKEN" \
   http://localhost:8081/management/current-claims
 ```
 
-Returns the full Claims.json structure currently in use. This is exactly the document the upload
-endpoint accepts, so the response can be uploaded again without modification.
+Returns the full Claims.json structure currently in use, along with an `X-Reload-Id` response header
+identifying the loaded claims. This is the only endpoint that sets that header, and the body is
+exactly the document the upload endpoint accepts, so the response can be uploaded again without
+modification.
 
 ### Round-Trip: Download and Re-Upload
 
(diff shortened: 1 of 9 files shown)


samuellugo-ship-it:
The obsolete top-level `claims` wrapper in the public v8 docs is fixed in Ed-Fi-Alliance-OSS/ed-fi-alliance-oss.github.io#630. The how-to guide now posts the downloaded document directly instead of rebuilding it with `@{ claims = ... }`, and the CmsHierarchy page's upload example drops the outer wrapper while keeping the hierarchy array in the two-section document as `claimsHierarchy`. Both revised examples validate against `JsonSchemaForClaims.json`.

Note on that repo's CI: `npm run lint` already fails on its `main` with pre-existing markdownlint errors across older versioned docs — `0` in the two changed files and `0` anywhere under `version-8`. `npm run build` passes, including the broken-link check.

simpat-adam (approved):

PR #1211, opened 2026-08-28
[DMS-1373] Load and Cache API-client Ownership Tokens from CMS in DMS

jleiva-gap:
Extends DMS application-context retrieval so the creator and read/modify ownership tokens
maintained by CMS reach the relational authorization pipeline.

`ApplicationContext` now carries a nullable `CreatorOwnershipTokenId` and a non-null read-only
`OwnershipTokenIds` collection read from `GET /v3/apiClients/{clientId}`. These stay separate from
the JWT-derived `ClientAuthorizations` — no ownership claim or identity-provider mapper is added —
and are propagated into `RelationalAuthorizationContext` alongside the existing client
authorizations. The DMS-1060 ownership SQL strategy is not implemented here.

Application-context lookup returns typed `Success`, `NotFound`, and `Unavailable` results, with a
malformed successful CMS response classified as unavailable. `NotFound` maps to an invalid-token
401; `Unavailable` maps to 503. A request-scoped holder memoizes the first outcome, so an
authenticated resource request resolves the context at most once. Context is required for every
POST, and for GET, PUT, and DELETE whose selected authorization strategies include
`OwnershipBased`; profile resolution independently requires it because it needs an `ApplicationId`
before strategies are known.

The existing application-context cache is kept and made tenant-aware: the tenant participates in
the cache key, and the CMS provider sends a `Tenant` header on the individual request when a tenant
is present, omitting it for single-tenant requests. `docs/CACHING-STRATEGY.md` is updated to the
resulting runtime contract, including the breadth of the profile-resolution dependency.

## Note to consider

The ownership-token contract drift raised in review was resolved in one direction: DMS exposes
ownership token ids as sorted-distinct, normalized at the `ConfigurationServiceApplicationProvider`
success boundary, rather than rejecting duplicate or unsorted CMS values. This keeps DMS decoupled
from CMS query ordering while giving downstream authorization a deterministic list. Presence,
short-range, and max-count validation are unchanged, and a genuinely invalid response still maps to
`Unavailable`.

### DIFF
--- a/src/dms/core/EdFi.DataManagementService.Core/Configuration/ConfigurationServiceApplicationProvider.cs
+++ b/src/dms/core/EdFi.DataManagementService.Core/Configuration/ConfigurationServiceApplicationProvider.cs
@@ -189,7 +189,21 @@ public class ConfigurationServiceApplicationProvider(
             && applicationContext.OwnershipTokenIds.Count <= MaximumOwnershipTokenCount
             && applicationContext.OwnershipTokenIds.Distinct().Count()
                 == applicationContext.OwnershipTokenIds.Count
-            && applicationContext.OwnershipTokenIds.All(IsValidOwnershipTokenId);
+            && applicationContext.OwnershipTokenIds.All(IsValidOwnershipTokenId)
+            && IsAscending(applicationContext.OwnershipTokenIds);
+    }
+
+    private static bool IsAscending(IReadOnlyList<short> ownershipTokenIds)
+    {
+        for (int i = 1; i < ownershipTokenIds.Count; i++)
+        {
+            if (ownershipTokenIds[i] <= ownershipTokenIds[i - 1])
+            {
+                return false;
+            }
+        }
+
+        return true;
     }
 
     private static bool IsValidOwnershipTokenId(short tokenId) =>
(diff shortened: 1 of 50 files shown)


samuellugo-ship-it (approved):

src/dms/core/EdFi.DataManagementService.Core/Configuration/ConfigurationServiceApplicationProvider.cs
```diff
@@ -20,72 +20,100 @@ public class ConfigurationServiceApplicationProvider(
...
 
             string responseBody = await response.Content.ReadAsStringAsync();
-            ApplicationContext? applicationContext = JsonSerializer.Deserialize<ApplicationContext>(
-                responseBody,
-                _jsonOptions
-            );
+            ApplicationContext? applicationContext = DeserializeApplicationContext(responseBody);
 
-            if (applicationContext == null)
+            if (
```
vimayya, line 113: Reviewed the last two commits only (`c5547354c`, `5c39baecc`).
Both look good. caching-doc update is correct. The `ApplicationId <= 0` guard is worth having. 
Non-blocking note: 
 **Rejection logging.** The `if` these guards join now has six distinct causes sharing
   one message, `"Failed to deserialize application context"` and a genuine parse failure
   doesn't even reach it, since `JsonException` is caught separately below. Mostly
   pre-existing, but the clientId-mismatch clause is the one I'd split out: it means CMS
   returned a context for a *different* client than we asked for, which deserves its own
   log line rather than being reported as malformed JSON.

vimayya (approved):

PR #1212, opened 2026-08-28
[DMS-1367] Route Eligible DMS Reads to Snapshots and Read Replicas

simpat-adam:
## Summary

- **DMS-1367**: routes an eligible DMS read to a snapshot or read replica database when the parent
  data store has one configured, and keeps every other request on the primary. One physical database
  is chosen once per request, before any database is opened, and every database operation in that
  request — fingerprint read, resource-key read, authorization SQL, repository query, hydration —
  uses it. A read replica is used automatically on a read-only, replica-eligible pipeline; a snapshot
  is used only when the request asks for one with `Use-Snapshot: true`. Implements
  `reference/design/backend-redesign/epics/10-update-tracking-change-queries/39-snapshot-read-replica-runtime-routing.md`.

- **Derivative configuration model.** `DataStore` becomes a record with an explicit constructor and a
  frozen `ImmutableDictionary` of derivative connection strings; the name, order, type, and default of
  all seven existing members are unchanged, so every existing construction site compiles untouched.
  `ConfigurationServiceDataStoreProvider` deserializes the `dataStoreDerivatives` collection CMS
  already returns and decrypts each derivative inside its own fault boundary: the type is recognized
  ordinally and exactly first, a blank stored value is treated as not configured *before* decryption
  is attempted, and only `InvalidOperationException` — the decryption service's documented failure
  contract — is caught, so an unrelated defect is not reinterpreted as absent configuration. No
  ciphertext, plaintext, key, or connection string reaches a log, and the ordinary not-configured path
  logs nothing so it stays distinguishable from the undecryptable one. A value that decrypts to a
  non-blank but provider-invalid string stays *configured* and fails later, at acquisition.

- **Endpoint validation now runs before database validation in every pipeline (behavior change for
  every deployment, derivative or not).** `ApiService` gains `GetEndpointValidationSteps` and
  `GetDatabaseValidationSteps` and all eleven pipeline factories are recomposed from them. Target
  selection has to sit between the two phases, so the reorder is a prerequisite rather than part of
  the routing change. Where a hoisted validation and a later database validation would both fail, the
  request now receives the hoisted verdict instead of the 503 the old order returned: an unroutable
  request answers the endpoint 404, and an invalid mutation route shape answers the route-semantics
  405.

- **Two-phase, write-once request selection.** `IDataStoreSelection` records the parent (tenant, route
  context, client authorization, logging identity) and, separately, the effective target. Reading
  either before assignment throws naming the missing phase, and a second assignment throws. There is
  deliberately **no** accessor that falls back from target to parent, so a pipeline that reaches the
  database without selecting fails loudly instead of quietly reading the primary.
  `EffectiveTargetSelector` is a pure function over the pipeline's `DerivativeRoutingPolicy`, the
  configured derivatives, and the `Use-Snapshot` header: presence and blank checks only, never a
  parse. The snapshot axis is evaluated first, so an explicit snapshot request outranks a configured
  replica; a snapshot that is not configured is rejected with no fallback; a snapshot on a mutation is
  rejected; every other path selects Primary explicitly. `SelectEffectiveDataStoreTargetMiddleware` is
  composed into all nine database-reading pipelines immediately before the database phase (and before
  route-semantics validation on mutations); the two method-not-allowed pipelines get no selection step
  because they open no database. A composition test derives the set of database-reading factories by
  reflection and asserts it equals the set the policy tests cover, so a newly added pipeline cannot go
  unpoliced.

- **One acquisition boundary per provider, and a leased data-source/pool lifetime.**
  `MssqlConnectionAcquisition` is the single place a SQL Server connection string is parsed: a primary
  passes through byte for byte with no builder constructed, so primary pool identity and any
  operator-supplied pool setting are untouched, while a derivative is rebuilt with
  `PoolBlockingPeriod.NeverBlock` so SqlClient does not replay a failed login on the next request. All
  five SQL Server seams share that identity through `AcquireLeaseAsync`. On PostgreSQL,
  `NpgsqlDataSourceCache` has a leased-only surface — `GetOrCreate` is removed rather than shimmed, so
  "no untracked or immortal data source" is structural. Both providers keep one state lock that never
  builds, opens, parses, clears, or disposes while held, so one slow database cannot stall another's
  acquisitions.

- **Configured ownership is published as one ordered global snapshot.** `ConfiguredTargetOwner`,
  `DataStoreOwnershipSnapshot`, and `IDataStoreOwnershipReconciler` are Core-owned;
  `ConfigurationServiceDataStoreProvider` replaces the tenant entry, increments the version, projects
  the owner union across every loaded tenant, and invokes reconcilers in a fixed order — all in one
  critical section, with the CMS token exchange, fetch, and `DataStore` construction outside it. Every
  failure path throws before the lock, so a failed load replaces no entry, consumes no version, and
  invokes no reconciler. Ownership is expressed only in configured terms and never parsed during
  reconciliation, so a derivative no provider could open is published like any other owner. Both
  provider caches reconcile against it: non-increasing versions are rejected, retirement is assigned
  from current ownership in *both* directions (an owner removed and re-added before its last lease
  ends keeps its live data source), and the whole candidate state is computed into locals before
  publication so a failed enumeration leaves version and owner set untouched and a same-version retry
  applies in full. SqlClient pool clearing is identity-scoped rather than object-scoped, so a retired
  unleased identity gets a generation-scoped tombstone published under the lock, the clear runs
  outside it, and completion is signalled from a `finally`.

- **Derivative validation verdicts get a lifetime of their own.** Both validation caches are keyed by
  (policy class, configured connection string) rather than by the string alone — the policy class is
  what keeps a primary and a derivative apart when their configured text is identical. Primary keeps
  every verdict it has today for the process lifetime. A derivative success expires at the new bounded
  TTL, every derivative fault is evicted, and a verdict the reading request finds unusable is dropped
  immediately, because a replica or snapshot can be rebuilt or repointed underneath a running service.
  Reads are synchronous, returning the value task plus an invalidation token that exists on every path
  including the fault paths; the token is a no-op for Primary and an *exact-entry* removal for a
  derivative, so a late verdict from a superseded entry removes nothing. The four validation failure
  responses no longer tell every caller to restart the service: that remediation is true only for a
  primary, and the derivative wording states that the result was not retained and the next request
  revalidates. Problem type, title, status, content type, and envelope are identical either way.

- **New setting `CacheSettings__DerivativeValidationCacheExpirationSeconds`** (default 600, range
  1–3600, env `DMS_DERIVATIVE_VALIDATION_CACHE_EXPIRATION_SECONDS`). A non-positive value falls back
  to 600 with a warning that says outright that expiration is *not* being disabled — deliberately
  inverting `DataStoreCacheExpirationSeconds` in the same section, because a verdict about a database
  an operator can rebuild without telling DMS must never be immortal. Above 3600 clamps to 3600. The
  effective value is further bounded by the data store cache TTL where one exists, so a verdict never
  outlives the connection string it was reached for.

- **The parent's DocumentCache is kept out of a derivative read.** The cache is materialized from and
  keyed by the parent database, and a derivative carries its own `dms.DocumentCache`.
  `TryGetSelectedTargetKey` now returns false first — before the target registry is consulted or a key
  is constructed — whenever an effective target is set and its kind is not Primary, reporting the new
  `DerivativeTargetSelected` fallback reason and `SkippedDerivativeTarget` telemetry label. Both entry
  points go through it, so lookup, canonical `ContentVersion` comparison, and direct fill are all
  unreachable for a derivative.

- **The first database touch of a request is inside the lease.** PostgreSQL fingerprint and
  resource-key validation opened a connection straight from the effective target's connection string,
  putting the very first read of a derivative outside the cache and therefore outside effective-string
  identity, ownership recompute, shared-owner retirement, and in-flight lease protection. Both readers
  now hold a lease across the read, matching the SQL Server pair.

- **Operator surface and docs.** `appsettings.json` and both compose files carry the new setting;
  `CONFIGURATION.md` and `CACHING-STRATEGY.md` document the default, range, clamp-versus-fallback
  behavior, the data store TTL interaction, and the convention inversion. `CACHING-STRATEGY.md` also
  corrects a pre-existing claim that the validation caches are scoped per data store: the key carries
  no data store id and no tenant, so two data stores configured with the same connection string share
  one entry, which is the point of keying on what is being validated rather than who asked.

### Notes for reviewers

- **Deliberately deferred to DMS-1368.** A snapshot requested on a mutation is rejected at selection
  with the *existing generic* method-not-allowed body, no `Allow` header, and the default content
  type; a snapshot that is not configured gets the existing not-found body. What this story pins is
  the status code, the point in the pipeline the decision is made, and the absence of database access.
  The snapshot-specific problem details and `Allow` set belong to DMS-1368.
- **Story 39's E2E unit is not in this branch.** Snapshot/replica databases in the local Docker stack
  plus CMS derivative registration and the six Story 39 E2E scenarios are not delivered. Coverage here
  is unit-level plus API-level integration against three real leased databases per dialect.
- **A test-support helper was fixed here, and it strengthens ~19 pre-existing assertions at once.**
  `PostgresqlTargetReachability.MakeUnreachableAsync` used `ALTER DATABASE ... CONNECTION LIMIT 0`,
  which a superuser is exempt from — and both the local suites and CI connect as `postgres`. The
  targets were never unreachable; a request failed only when it happened to reuse a pooled session the
  accompanying `pg_terminate_backend` had killed. It now uses `ALLOW_CONNECTIONS false`, which binds
  superusers too (a direct open is refused with `55000`). All 33 PostgreSQL derivative tests pass
  against targets that are genuinely unreachable for the first time.
- `LeasedNpgsqlConnection.DisposeOwnedAsync`'s XML summary overstates what `try`/`finally` guarantees
  if the `finally` itself throws. Known, left as-is.
- Inserting `DerivativeTargetSelected` mid-enum shifts later ordinals of that internal enum. Every
  repository consumer compares members or names, so there is no behavioral or compatibility
  consequence.
- `/deletes` and `/keyChanges` are proven positively on the `AuthoritativeDs52` fixtures. On the
  routing fixture's ApiSchema, Student maps no tracked-change identity, so both answer 500 whichever
  target serves them; that predates this work and is not something routing can fix.
- `main` has advanced 7 commits since this branch's base; it has not been rebased.

## Test plan

- [ ] `dotnet build ./src/dms/EdFi.DataManagementService.sln` and `dotnet csharpier check` on the
      changed files
- [ ] Unit — configuration and ownership: `Core.Tests.Unit/Configuration`
      (`DataStoreDerivativeTypeTests`, `DataStoreTests`,
      `ConfigurationServiceDataStoreProviderDerivativeTests` including the blank-stored and
      decrypts-to-blank boundaries with a recording decryption service, `DataStoreSelectionTests` for
      the write-once phases, `DataStoreOwnershipPublicationTests` for the gated ordering matrix,
      `DerivativeValidationCacheExpiration{,Binding}Tests`, `ValidationCachePolicyTests`,
      `ValidationEntryCacheTests`, `ValidationCacheRemediationTests`,
      `ValidationCacheMiddlewareOwnershipTests`)
- [ ] Unit — selection and header: `Core.Tests.Unit` `EffectiveTargetSelectorTests`,
      `SelectEffectiveDataStoreTargetMiddlewareTests`, `UseSnapshotHeaderTests`, plus
      `PipelineOrderingTests` for the phase order, per-factory phase composition, and the
      reflection-derived "every database-reading pipeline is policed" assertion
- [ ] Unit — provider lifetimes: `Backend.Tests.Unit` `NpgsqlDataSourceCacheInterleavingTests`,
      `NpgsqlDataSourceProviderLeaseTests`, `PostgresqlEffectiveTargetValidationLeasingTests`,
      `MssqlConnectionAcquisitionTests`, `MssqlOwnershipReconciliationTests`,
      `MssqlClearingTombstoneTests`, `MssqlSeamAcquisitionIdentityTests`,
      `MssqlAcquisitionCompositionTests`, both `*OwnershipReconcilerRegistrationTests`,
      `DocumentCacheAdministrativeMutexLeaseOwnershipTests`. All barrier-driven; none reads elapsed
      time.
- [ ] Unit — DocumentCache bypass: `Core.Tests.Unit/DocumentCache` — GET-by-id and query for both
      derivative kinds with read acceleration enabled and a resolving target, a Primary selection
      still serving the cache, and the target-signature comparison reached independently of the guard
- [ ] Integration PostgreSQL + SQL Server — routing against three real leased databases:
      `Given_{Postgresql,Mssql}_DerivativeRouting`. Every routing claim rests on making the databases
      a request must not touch unreachable *at the server* with their connection strings untouched, so
      a request that succeeded cannot have opened one. Covers replica-served reads, snapshot
      outranking replica, GET-by-id, `/availableChangeVersions`, a fully walked `/partitions` cursor,
      the typed not-found for an unconfigured snapshot, all three invalid mutation shapes with the
      exact generic 405 body and `Allow` set, writes landing on the parent, replacement and removal
      across a configuration change, and twelve in-flight requests held at a gated hydrator while a
      new configuration is published under them
- [ ] Integration PostgreSQL + SQL Server — `Given_{Postgresql,Mssql}_ValidationPhasePrecedence`: the
      full nine-cell hoisted-verdict matrix including `/partitions`, each fixture breaking one
      database-validation stage before the host boots and carrying a control that must still get 503
- [ ] Integration PostgreSQL + SQL Server — `Given_{Postgresql,Mssql}_DerivativeDocumentCache`,
      `_DerivativeParentAuthorization` (real namespace authorization plus a district route qualifier,
      asserting both the allowed and the denied outcome while a derivative serves),
      `_DerivativeTrackedChanges` (distinguishable tracked-delete and key-change counts per database),
      `_DerivativeStartupIsolation` (a realization recorder asserting zero realizations for an
      unusable snapshot and replica while the primary is non-zero)
- [ ] Integration SQL Server — `Given_Mssql_DerivativeAvailabilityRecovery`: with the snapshot offline
      the routed read fails, and once it is online the very next request reaches the server at the
      same effective connection string rather than being refused for SqlClient's blocking period
- [ ] Regression — the existing backend integration suites, which reach the primary through
      `PrimarySelectingTestDataStoreSelection` in `Backend.Tests.Common` (registered in one place, the
      relational integration-test seam; production never registers it)


src/dms/core/EdFi.DataManagementService.Core/ApiService.cs
```diff
@@ -505,13 +615,9 @@ private PipelineProvider CreateMethodNotAllowedPipeline()
...
         var steps = GetCommonInitialSteps();
-        steps.AddRange([
-            new ParsePathMiddleware(_logger),
-            new ApiSchemaValidationMiddleware(_apiSchemaProvider, _logger),
-            new ProvideApiSchemaMiddleware(_effectiveApiSchemaProvider, _logger),
-            new ValidateEndpointMiddleware(_logger),
-            new MethodNotAllowedMiddleware(_logger, _isTrackedChangeRoute: false),
-        ]);
+        steps.Add(new ParsePathMiddleware(_logger));
+        steps.AddRange(GetEndpointValidationSteps());
```
bradbanister, line 630: **High:** Unsupported HTTP verbs bypass snapshot mutation policy. The catch-all method-not-allowed pipelines start at `src/dms/core/EdFi.DataManagementService.Core/ApiService.cs:605` and proceed directly from endpoint validation to `MethodNotAllowedMiddleware`, so `PATCH` or `HEAD` with `Use-Snapshot: true` never produces `RejectedAsMutation`. The tests explicitly enforce this behavior (`src/dms/core/EdFi.DataManagementService.Core.Tests.Unit/Pipeline/PipelineOrderingTests.cs:1046`). Jira and the story require typed mutation rejection for "any non-GET resource or descriptor request" (`reference/design/backend-redesign/epics/10-update-tracking-change-queries/39-snapshot-read-replica-runtime-routing.md:51`), although the endpoint summary more narrowly names `POST`/`PUT`/`DELETE`. Reusing the selection step after endpoint validation would preserve unknown-resource precedence and let DMS-1368 supply the correct snapshot response.

src/dms/core/EdFi.DataManagementService.Core/ApiService.cs
```diff
@@ -329,10 +405,16 @@ private PipelineProvider CreateQueryPipeline()
     private PipelineProvider CreateGetPartitionsPipeline()
```
bradbanister, line 412: **Medium:** `/partitions` derivative routing is an undocumented public-surface expansion. This pipeline is now snapshot- and replica-eligible, and integration coverage proves it (`src/dms/tests/EdFi.DataManagementService.Tests.Integration/Scenarios/DerivativeRoutingScenario.cs:87`). Neither Jira nor Story 39 includes `/partitions` in endpoint coverage (`reference/design/backend-redesign/epics/10-update-tracking-change-queries/39-snapshot-read-replica-runtime-routing.md:76` and `:78`), and the story describes selection in eight pipelines rather than the nine now implemented. The behavior is defensible after the DMS-1394 rebase because partition sizing and page reads should use the same database, but it needs an explicit story/Jira amendment; otherwise this is scope creep and expands E2E setup for unrelated lanes. The partition design also says snapshot-consistent export is a non-goal (`reference/design/backend-redesign/design-docs/partitioned-cursor-paging.md:103`).

src/dms/backend/EdFi.DataManagementService.Backend/DocumentCacheReadAccelerationCoordinator.cs
```diff
@@ -969,7 +970,15 @@ out directFillSkipOutcome
             return false;
         }
 
-        if (!TargetMatchesSelectedDataStore(selectedDataStore, resolvedTargetContext))
+        // The database this request will actually read. Reaching here means either no effective target
```
bradbanister: **Medium:** Document-cache read acceleration can silently fall back to primary when no effective target is set. This path uses the parent connection string if `_dataStoreSelection.IsEffectiveTargetSet` is false. The story's selection contract is intentionally fail-fast: backend consumers should read the effective target and not silently serve primary. Current read pipelines appear to select first, but this fallback can hide a missing-selection composition bug on a cache hit. Simpler fix: require `IsEffectiveTargetSet` for request read acceleration and bypass/fail when absent; keep non-request background behavior behind a separate path.

src/dms/backend/EdFi.DataManagementService.Backend.Postgresql/NpgsqlDataSourceCache.cs
```diff
@@ -3,87 +3,341 @@
...
-                );
+                return LeaseLocked(connectionString, existing);
+            }
+        }
 
-                NpgsqlDataSourceBuilder builder = new(cs);
-                var csb = builder.ConnectionStringBuilder;
+        // Built outside the lock. This is where Npgsql parses, and where a provider-invalid string
+        // throws - inside the acquisition boundary, with no shared state touched.
+        NpgsqlDataSource candidate = Build(connectionString);
```
bradbanister, line 110: **Medium:** Npgsql data-source acquisition can leak a candidate during a dispose race. `AcquireLease` builds a candidate outside the lock, then checks `_disposed` inside the second lock. If the cache is disposed after `Build(connectionString)` and before that second check (`src/dms/backend/EdFi.DataManagementService.Backend.Postgresql/NpgsqlDataSourceCache.cs:111`), the method throws while `loser` is still null, so the `finally` does not dispose the just-built `NpgsqlDataSource`. Initialize the disposable candidate as the default cleanup target and clear it only after ownership is transferred.

src/dms/core/EdFi.DataManagementService.Core/Pipeline/RequestInfo.cs
```diff
@@ -194,6 +195,14 @@ public RequestMethod Method
     /// </summary>
     public ProfileContext? ProfileContext { get; set; }
 
+    /// <summary>
```
bradbanister: **Low:** Request-scoped selection-result state is dead production plumbing. `RequestInfo.EffectiveTargetSelection` is written but never read by production code. The selected case duplicates `IDataStoreSelection`; rejected cases are already distinguished by separate response-factory methods; and their `ParentDataStoreId` fields are only asserted by tests while middleware uses the parent directly. Removing this state and the unused IDs would reduce coupling unless DMS-1368 has a concrete consumer. Related lines: `src/dms/core/EdFi.DataManagementService.Core/Middleware/EffectiveTargetSelector.cs:63` and `src/dms/core/EdFi.DataManagementService.Core/Middleware/SelectEffectiveDataStoreTargetMiddleware.cs:47`.

src/dms/core/EdFi.DataManagementService.Core/Configuration/ConfigurationServiceDataStoreProvider.cs
```diff
@@ -359,28 +505,128 @@ CancellationToken cancellationToken
...
+            {
+                // The decryption service returns null only for a null or empty input, which the check
+                // above has already excluded, so normalizing to an empty string here changes no
+                // outcome: the blank check below treats null, empty, and whitespace alike as not
+                // configured, and this keeps the local non-nullable.
+                plainText =
+                    _connectionStringDecryptionService.DecryptFromBase64(derivative.ConnectionString)
+                    ?? string.Empty;
+            }
+            catch (InvalidOperationException ex)
```
bradbanister, line 593: **Low:** Derivative decryption logging relies on the current decryptor's sanitized exception messages. This catch logs the caught derivative decryption exception object. The current concrete decryptor uses sanitized messages, but the no-secret guarantee is stronger than that implementation detail. To keep the provider boundary responsible, log derivative type/tenant/data-store id and possibly exception type, but not `ex`.

src/dms/backend/EdFi.DataManagementService.Backend.Mssql/MssqlServiceExtensions.cs
```diff
@@ -27,6 +28,27 @@ IConfiguration configuration
 
         services.AddRelationalMappingSetServices(configuration, SqlDialect.Mssql, new MssqlDialectRules());
```
bradbanister, line 31: **Low:** Provider lifecycle registration is duplicated across composition roots. The SQL Server acquisition and reconciler registrations are repeated in `AddMssqlDatastore` and `AddMssqlReferenceResolver`; PostgreSQL repeats the same cache/reconciler block. The standalone composition requirement is valid, but a shared registration helper would express it once and prevent future lifetime or implementation-type drift. Related lines: `src/dms/backend/EdFi.DataManagementService.Backend.Mssql/MssqlReferenceResolverServiceCollectionExtensions.cs:24` and `src/dms/backend/EdFi.DataManagementService.Backend.Postgresql/PostgresqlServiceExtensions.cs:35`.

src/dms/backend/EdFi.DataManagementService.Backend.Postgresql/NpgsqlDataSourceLease.cs
```diff
@@ -0,0 +1,153 @@
...
+    /// Every owner that holds a transaction over a leased connection needs this exact ordering, and
+    /// getting it wrong strands a lease for the life of the process rather than failing visibly, so it
+    /// lives here once rather than being rewritten at each owner.
+    /// </remarks>
+    internal static async ValueTask DisposeOwnedAsync(
+        IAsyncDisposable precedingResource,
+        LeasedNpgsqlConnection? owned
+    )
+    {
+        try
```
bradbanister, line 116: **Low:** Npgsql owned-disposal helper can mask the first exception. `LeasedNpgsqlConnection.DisposeOwnedAsync` says the first disposal exception propagates, but the `finally` can mask it if `owned.DisposeAsync()` also throws. SQL Server already has `DisposeWithoutMaskingAsync`; Postgres should use the same pattern or update the claim.

docs/CACHING-STRATEGY.md
```diff
@@ -444,6 +514,8 @@ Default values: ClaimSets, AppContext, and data store = 10 minutes; Token = 25 m
 | NpgsqlDS     | ConcurDict   | Sing. | None   | N/A    | No       | Shutdown     |
```
bradbanister: **Low:** The cache documentation still describes obsolete Npgsql lifecycle behavior. The summary table says `NpgsqlDS` is invalidated only at shutdown, but the implementation now retires and disposes data sources after configuration ownership disappears and leases finish (`src/dms/backend/EdFi.DataManagementService.Backend.Postgresql/NpgsqlDataSourceCache.cs:177`). This table should reflect configuration reconciliation plus shutdown.

bradbanister:
Static review of DMS-1367 focused on correctness, simplicity, duplicated/dead code, and scope creep. Found 1 high, 3 medium, and 5 low issues. No local tests were run.

bradbanister (approved):

PR #1217, opened 2026-08-31
[DMS-1395] DMS Performance tuning with SQL Server backend

axelmarquezh:
**Why these changes**

Load testing the relational backend on SQL Server (volume suite, 10 users) showed MSSQL sustaining ~550 req/s where the equivalent PostgreSQL deployment sustains ~1,258. Profiling with Query Store and XE traced the gap to three MSSQL-specific mechanisms, each addressed by one change in this set:

1. **Statement recompile storm.** MSSQL delete plans embed a NO ACTION referential-check apparatus spanning every table that references the deleted row's cascade targets (~58 tables for `dms.Document`). Any auto-stats refresh on any of those tables invalidates every per-resource delete batch — measured at ~44 recompiles/s, 90% of all compilation, with the Document DELETE alone recompiling 18k+ times per run at ~46 ms each. These statements are fixed unique-key seeks whose optimal plan cannot change with statistics, so `OPTION (KEEPFIXED PLAN)` on the delete path and on the identity-trigger UPDATE suppresses the recompiles without affecting plan quality. (Freezing stats database-wide proved the diagnosis on the rig — 550→1,071 req/s — but isn't shippable; this hint is the targeted equivalent.)

2. **Serializable range locks on the document version lock.** `WITH (UPDLOCK, HOLDLOCK, ROWLOCK)` takes key-range locks (`LCK_M_RS_U`) on `PK_Document`, producing lock convoys and hundreds of deadlocks per run under concurrent writes. The lock only needs to pin one existing row, so it becomes `WITH (UPDLOCK, ROWLOCK)` — which also matches the PostgreSQL path, where `FOR UPDATE` likewise takes no range lock. The `DocumentProjectionWork` queue claim keeps HOLDLOCK, as it genuinely needs insert-if-absent semantics.

3. **Identity-maintenance triggers doing full-cost work for zero rows.** The abstract-identity trigger UPDATE (e.g. `TR_School_AbstractIdentity`) carries the referential-check apparatus for every FK referencing the identity table (~90 for `EducationOrganizationIdentity`), costing 30–40 ms of CPU per execution even when its `@changedDocs` workset is empty — the overwhelmingly common case, since identity values rarely change. The trigger bodies now skip the maintenance DML entirely when the workset is empty, and the insert-branch UPDATE (only ever live in reseed/repair scenarios) runs behind a cheap PK-seek existence probe.

PostgreSQL output is byte-identical; the changes are MSSQL-only. Golden fixtures and DDL manifests regenerated accordingly.


src/dms/backend/EdFi.DataManagementService.Backend/RelationalDocumentLockCommandBuilder.cs
```diff
@@ -85,7 +85,7 @@ FOR UPDATE
             SqlDialect.Mssql => """
                 SELECT
                     document.[ContentVersion] AS [ContentVersion]
-                FROM [dms].[Document] document WITH (UPDLOCK, HOLDLOCK, ROWLOCK)
+                FROM [dms].[Document] document WITH (UPDLOCK, ROWLOCK)
```
axelmarquezh, line 88: This is the change number 2 (see the PR description for more information).

src/dms/backend/EdFi.DataManagementService.Backend/OrderedDeleteCommandBuilder.cs
```diff
@@ -121,13 +126,23 @@ DELETE FROM dms."Document"
             SqlDialect.Mssql => """
                 DELETE FROM [dms].[Document]
                 OUTPUT DELETED.[DocumentId]
-                WHERE [DocumentId] = @documentId;
+                WHERE [DocumentId] = @documentId
+                OPTION (KEEPFIXED PLAN);
```
axelmarquezh, line 130: This is the change number 1 (see the PR description for more information).

src/dms/backend/EdFi.DataManagementService.Backend.Ddl/RelationalModelDdlEmitter.cs
```diff
@@ -2547,43 +2571,38 @@ bool isInsert
...
-        writer.Append("SET ");
-        for (int i = 0; i < mappings.Count; i++)
+        if (isInsert)
         {
-            if (i > 0)
+            // On the insert path a matching identity row only exists in reseed/repair scenarios,
+            // yet the UPDATE's plan carries a referential-check apparatus spanning every FK that
+            // references the identity table. Probe with a cheap PK seek and skip the UPDATE in the
+            // (overwhelmingly common) no-match case.
+            writer.Append("IF EXISTS (SELECT 1 FROM ");
```
axelmarquezh, line 2580: This is the change number 3 (see the PR description for more information).

samuellugo-ship-it (approved):

PR #1218, opened 2026-09-02
[DMS-1323] Add Explicit Local/Bootstrap Connector Registration

JBrenesSimpat:
## Summary

  - Adds a new `Backend.Cdc.Control` library holding the deployment-owned CDC control plane: the setup controller, a Kafka admin adapter, a Kafka Connect REST adapter,
  a Jolokia lag reader, the eligibility probe, the projection-correlation collector, provider setup-input/provisioning-proof factories, and provider artifact teardown.
  It sits outside `Backend.Cdc` and outside the CLI so runtime DMS never gains a control-plane dependency and the orchestration is not exempted from coverage.
  - Extends `dms-document-cache` with a `cdc` verb group — `enable`, `status`, `restart`, `adopt`, `replace-source`, `retire` — rather than adding a third CLI, reusing
  that tool's target resolution, provider runtime services, JSON output, exit-code mapping, and log sanitization. Verb identity is scoped, since `cdc status` is not the
  DocumentCache `status` command.
  - Provisioning evidence is caller-supplied and never inferred: `enable` and `replace-source` require the exact `--database-creation-mode
  created-for-initial-cdc-provisioning` and `--write-admission closed-never-opened` tokens, destructive verbs require their own confirmation token, `replace-source`
  names the generation it supersedes explicitly, and `--max-record-bytes` has no default so an absent value fails closed instead of inheriting a broker default.
  - `replace-source` fences the outgoing generation's connector — stopped, not deleted, so its configuration and committed offsets survive for the retirement that
  removes them in order — and then runs the replacing generation through the same initial readiness sequence `enable` runs. Every refusal it can decide without reading
  state it would itself create is decided before the fence, so a target that cannot be replaced is never left with its publication stopped. It refuses a replacing
  source whose identity has not been rotated away from the generation it replaces, because binding a new generation to an unrotated `dms.DataStoreIdentity` would
  publish one physical source under two generations.
  - Wires the E18 administrative gate to real evidence: a production `IDocumentCacheDownstreamPublicationHistoryProvider` reads the durable binding state store and
  reports `internalOnly` only for a complete listing that holds at least one readable binding and none for the target. `active`, `historical`, `unknown`, an unreadable
  record, a failed listing, and an empty listing all keep activate-offline, deactivate-offline, and recover-cache-ahead rejected with no mutation — replacing the
  default that always answered `unknown`.
  - Adds cluster-scoped Connect offset-store provisioning/validation (compaction, durability, worker-only ACLs) and binding-scoped topic, durability, record-size, and
  ACL provisioning/validation. `status` and `restart` only observe — they never provision, so an interrupted retirement is not silently undone.
  - Every provisioning gate is applied before the side effect it guards, not after. Provider artifacts are created, then read back through the same validate-only
  inspection every later status check uses, and a pass that is not an exact match ends the sequence there — before the shared offset store, the binding topics, or the
  connector registration. A connector registered against nonconforming artifacts would already be capturing from them by the time the final evaluation rejected the
  state.
  - A refused provider-setup pass reports the provider's own diagnostics rather than only its outcome. A missing connector principal, a refused grant, and an exhausted
  step budget are all one `Failed` outcome and are told apart only by what the provider said.
  - Retirement ordering is explicit: stop the connector, delete its committed offsets while it is stopped and still exists, delete the connector, then the governed
  topics and ACLs, provider capture artifacts, incident state, and the binding record last — only against a validated cleanup proof. A connector the worker does not
  have refuses retirement unless the operator asserts `--connector-already-absent`, and the proof records that as the operator's assertion.
  - A binding generation is never rebound. Creating a binding for a generation this deployment already retired is refused, and an existing retirement record must
  exact-match the binding being retired — only its latch time may differ — before it counts as an idempotent retry. The generation and its path are stable across a
  destructive volume removal, so accepting either would reassign an existing connector name, topic namespace, and consumer state to a different physical database.
  - The enable phase therefore allocates the generation from the store rather than fixing it at the target's first: one past the highest the store has ever held for the
  instance key, counting retirement records as well as live bindings, because retirement removes the binding record it retires and leaves the retirement record as the
  only trace. So a second local cycle binds generation 2 and publishes under its own connector, topics, and consumer state. The root is resolved the way Compose
  resolves the `/state` mount source, since that is the store the container will actually write to.
  - A terminal source-history loss that could not be latched durably is reported rather than dropped: it lands in the status's state-store diagnostics, which is what
  the binding state is evaluated as unavailable from, the connector stays fenced, and restart is refused. The loss is terminal and the record that must outlive the poll
  does not carry it, so no later step may read the binding as though it did.
  - Reads connector lag over a JMX-to-HTTP (Jolokia) bridge — Debezium's `MilliSecondsBehindSource` plus its P50/P95/P99 — never from the progress topic. An unavailable
  bridge yields `Unknown`, which keeps readiness false; a timeout never opens writes.
  - Polling budgets bound real elapsed time, not just the decision between observations. The Connect stop-wait and the controller's generic poller issue each
  observation under a token linked to the budget's remainder and clamp the wait to it; otherwise every poll got a fresh full request timeout and a slow-answering step
  took many multiples of the budget its own exhaustion message named.
  - Makes local Kafka and Kafka Connect engine-neutral and adds an `-EnableKafkaCdc` opt-in (with optional `-CdcBindingStatePath`) to `start-local-dms.ps1`, the
  bootstrap wrapper, and the DMS E2E setup wrapper. Previously Kafka was gated to postgresql; SQL Server is a first-class CDC provider, so neither the compose set nor
  the start sequence branches on the engine. Infrastructure opt-in alone configures no projection target and enables tracking on nothing.
  - The Kafka Connect image is operator-supplied via `DMS_CDC_CONNECT_IMAGE` and must be digest-qualified — a tag is rejected before Compose runs, because a moving tag
  makes a registered connector's runtime unreproducible. The non-CDC Kafka/Kafka-UI path keeps its `:pre` tag default.
  - Runs the control-plane verbs as a one-shot container (`cdc-setup.yml` + `DocumentCacheAdmin.Dockerfile`) on the `dms` network, since the instance database is
  registered in CMS under its container alias and the broker advertises a container-internal listener. One shared builder constructs those compose arguments for every
  verb, so the enable phase and the destructive teardown cannot drift on the connector base URL, record-size budget, binding-state path, durability profile, user
  override, setup principal, or build flags. The durable binding state store is a persistent bind mount outside the `.bootstrap` manifest, and the image clears the
  group/other-write bits Docker Desktop forces on any bind mount so the store does not refuse its own root.
  - Provisions a dedicated connector database principal (`provision-cdc-principal.ps1`), separate from the admin login. Its password reaches the worker as
  `CDC_DATABASE_PASSWORD` and the connector config references `${env:CDC_DATABASE_PASSWORD}`, so the secret never lands in the config topic or in any read-back the
  control plane compares against.
  - Destructive teardown (`-d -v`, including E2E) retires bindings before volumes are removed, in the same pass that removes every artifact the record governs; a normal
  stop retains the binding, connector, offsets, topics, ACLs, and capture artifacts. **A retirement that fails now aborts the volume removal and leaves the stack
  running**, and a file under the store that cannot be read as a binding record is fatal rather than skipped — removing the volumes around a surviving binding record
  destroys the connector, offsets, topics, and capture artifacts it still names, which is the one outcome the cleanup rule forbids and which destroys everything a
  retirement retry would act on. Abandoning that state is an explicit operator decision: `-AbandonCdcBindingState` on `start-local-dms.ps1` and
  `bootstrap-local-dms.ps1`, which requires `-d -v`, reports what it could not retire, and proceeds. It is never inferred from a retirement that could not run.
  - Core fix: the Connect offset-store policy validator now accepts an explicit `ConnectOffsetStoreInvalid` diagnostic as an invalid-policy fact, so a replica or
  min-ISR value that is positive but below the active durability profile is expressible.
  - Docs: new story doc `19-cdc-kafka/04-bootstrap-enable-kafka-cdc.md` recording the placement, command-surface, evidence, teardown-ordering, generation-allocation,
  and entry-point resolutions; the E18 CLI doc no longer claims the production history default stays `unknown`; operator tables added to the docker-compose and tool
  READMEs plus `.env.example`.

  ## Test plan

  - [ ] `./build-dms.ps1 UnitTest` — picks up the new `Backend.Cdc.Control.Tests.Unit` project (491 `It_` cases, 614 executed with `TestCase` expansion) automatically
  via the `*.Tests.Unit` glob: options validation, Connect REST adapter and secret redaction, observation mappers (configuration/runtime/offset/lag), Kafka
  topic/ACL/record-size/offset-store/schema-history/teardown policy, eligibility probe, downstream-publication-history provider, projection-target proof, and
  setup-controller preflight/sequence/retry/status/adoption/replace-source/teardown — including the provider validate-only guard, the provider-diagnostic pass-through,
  and the non-durable source-history latch.
  - [ ] CLI surface: `Given_DocumentCacheAdminCdcCommandSurface`, `...CdcCommandDispatcher`, `...CdcCommandRequests`, `...CdcExitCodes`,
  `Given_LocalCdcOptInConfiguration`, and the `Given_DocumentCacheAdminCdcJsonContracts` integration fixture — verb parsing, exact-token refusals, `replace-source`
  dispatch carrying the operator's superseded generation, exit-code mapping, and the JSON contract shapes.
  - [ ] Core: `Given_LocalCdcStateStore` — the retired-generation refusal and the retirement-record exact match, alongside the existing create/read/latch/cleanup cases.
  - [ ] Provider integration on both engines (these land in the existing engine-sharded CI jobs): `PostgresqlCdcControlReadinessTests` / `MssqlCdcControlReadinessTests`
  (`CdcControlReadiness`, shard 4) and `PostgresqlCdcEligibilityProbeTests` / `MssqlCdcEligibilityProbeTests` (`CdcEligibilityProbe`, shard 1) — initial readiness plus
  post-enablement lifecycle, and the retry classification (exact binding + `Disabled` + clear latch retries; exact binding + `Tracking` + clear latch + empty tables
  resumes; set cache-ahead latch, unbound `Tracking`, binding mismatch, or pre-capture rows fail closed).
  - [ ] Broker-backed suite, run locally — `--filter "Category=CdcControlBrokerBacked"` and `Category=CdcControlDiagnostics` in the new
  `Backend.Cdc.Control.Tests.Integration` project. These categories are not registered in any workflow, so they need a deliberate local run; they also require the
  qualified Ed-Fi Kafka Connect image by digest (the unqualified local `ed-fi-kafka-connect:pre` lacks the Ed-Fi partitioner and fails a handful of cases for that
  reason, not a code defect).
  - [ ] Pester, run locally — `Invoke-Pester eng/docker-compose/tests/BootstrapEnableKafkaCdc.Tests.ps1` (135 cases covering the opt-in refusals, the three-way `/state`
  path agreement, digest gating, generation allocation, and teardown ordering). This file is not in the CI bootstrap Pester lane's explicit path list; the four CDC
  cases in the registered `BootstrapEntryPointWorkflow.Tests.ps1` — engine-neutral Kafka selection, the teardown forwarding whitelist, and the binding-state abandonment
  switch with its `-d -v` guard — do run on every PR.
  - [ ] Turnkey local bring-up on PostgreSQL: `./start-local-dms.ps1 -EnableKafkaCdc` with `DMS_CDC_CONNECT_IMAGE` set to a digest — confirm the projection target and
  status role are configured before DMS starts, the enable workflow reaches `admitted`, and `cdc status` reports ready.
  - [ ] Same bring-up with `-DatabaseEngine mssql`, confirming Kafka/Connect start and a SQL Server connector registers and validates.
  - [ ] Second cycle on the same checkout: `-d -v`, then `-EnableKafkaCdc` again — the new binding is generation 2 against the fresh database, its governed names carry
  `-g2`, and the retirement record for generation 1 remains.
  - [ ] Fail-closed checks: `-EnableKafkaCdc` with a tag-only `DMS_CDC_CONNECT_IMAGE` (refused before Docker runs), with `-IdentityProvider keycloak` (warned/refused),
  with more than one configured data store or a route-qualified one, and `-CdcBindingStatePath` on a run that neither enables nor tears down CDC.
  - [ ] Teardown fail-closed: a `-d -v` whose retirement cannot succeed (stop the Connect worker first) leaves the volumes and the binding record in place and fails;
  the same run with `-AbandonCdcBindingState` reports the unretired binding and removes the volumes; `-AbandonCdcBindingState` without `-d -v` is refused.
  - [ ] E18 gate on the real provider: activate-offline / deactivate-offline / recover-cache-ahead admitted only after a readable binding exists for another target, and
  rejected without mutation against an empty store, an active binding, and a historical binding.
  - [ ] Retirement paths: `cdc retire` against a live binding (proof accounts for every governed artifact, offset store untouched), against an already-absent connector
  without the flag (refused, binding retained), and with `--connector-already-absent` (proceeds, proof records the operator assertion).
  - [ ] `./start-local-dms.ps1 -d` retains the binding and its artifacts; `-d -v` retires before removing volumes.
  - [ ] E2E: `setup-local-dms.ps1 -EnableKafkaCdc` on a fresh database — enablement admitted with write admission still closed, then the suite passes and
  `teardown-local-dms.ps1` retires cleanly.
  - [ ] Package smoke: `eng/ci/Invoke-DocumentCacheAdminPackageSmoke.ps1` — the `cdc` group and `cdc enable`'s evidence/state-path/durability options resolve on the
  installed tool, proving the `Backend.Cdc.Control` reference survived packaging.


eng/docker-compose/local-dms.yml
```diff
@@ -69,6 +69,14 @@ services:
       AppSettings__DomainsExcludedFromOpenApi: ${DOMAINS_EXCLUDED_FROM_OPEN_API:-}
       AppSettings__RouteQualifierSegments: ${ROUTE_QUALIFIER_SEGMENTS:-}
       SCHEMA_PACKAGES: ${SCHEMA_PACKAGES:-}
+      # Deployment-owned CDC. All three are blank unless the bootstrap CDC opt-in wrote them:
+      # a blank tenant/data-store pair binds to no projection target at all, and a blank role
+      # leaves the DocumentCache status endpoint unmapped, so a non-CDC run is unaffected.
+      # Targets binds through indexed keys and this wiring carries exactly one, matching the
+      # design's one connector per instance database - it is not multi-target support.
+      DataManagement__DocumentCache__Targets__0__TenantKey: ${DMS_CDC_TARGET_TENANT_KEY:-}
```
bradbanister, line 89: **High:** This always materializes `DocumentCache:Targets[0]`, even when CDC is disabled and both variables are blank. The existing binder/validator does not interpret a blank indexed entry as no target: the blank `DataStoreId` cannot bind to `long` (and zero is invalid). The same wiring exists in `published-dms.yml`, so the default non-CDC stacks regress. Put these target keys in a CDC-only override/profile so the collection is omitted entirely when CDC is off.

src/dms/backend/EdFi.DataManagementService.Backend.Cdc.Control/CdcSetupController.cs
```diff
@@ -0,0 +1,3195 @@
...
+
+        CdcObservationContext context = new(
+            request.OperationId,
+            target.ToTargetIdentity(),
+            physicalSourceFingerprint
+        );
+
+        evaluation = evaluation with
+        {
+            ConnectOffsetStore = await kafkaAdmin
```
bradbanister: **High:** These ensure calls return policy observations, but the controller never checks whether either policy is `Satisfied`; it proceeds to `PutConnectorConfigAsync` even for `Invalid` or `Unknown`. That violates the design's before-registration gate and can start publication with invalid durability, record-size, or ACL policy. Return a fail-closed admission before registration unless both observations are satisfied, and add invalid/unknown tests proving registration is not called.

eng/docker-compose/start-local-dms.ps1
```diff
@@ -887,29 +1033,41 @@ else {
...
+            }
         }
 
-        if ($enableKafkaInfrastructure -and $DatabaseEngine -eq "postgresql") {
+        if ($enableKafkaInfrastructure) {
             Write-Output "Starting Kafka infrastructure..."
+            # kafka-postgresql-source is the Kafka Connect service. The name predates the
+            # engine-neutral CDC workflow and is kept: renaming it would break existing local
+            # workflows and any external reference to the container name.
             docker compose $files --env-file $EnvironmentFile -p dms-local up $upArgs kafka kafka-postgresql-source
```
bradbanister, line 1261: **High:** Starting Kafka Connect here alongside Kafka lets the worker auto-create its shared offset topic before the control plane can provision it. The checked-in broker test explicitly proves that Connect-first topic is invalid because `min.insync.replicas` is inherited rather than an explicit topic override, and `EnsureConnectOffsetStoreAsync` validates rather than repairs an existing topic. Split startup into Kafka -> pre-create/validate the offset store -> Kafka Connect.

src/dms/backend/EdFi.DataManagementService.Backend.Cdc.Control/CdcSetupController.cs
```diff
@@ -0,0 +1,3195 @@
...
+            );
+        }
+
+        logger.LogDebug(
+            "CDC source replacement fenced generation {Generation} and is enabling generation {Replacement}.",
+            request.PreviousGeneration,
+            target.Generation
+        );
+
+        return await EnableAsync(
```
bradbanister: **High:** Source replacement cannot be implemented by re-entering initial enablement. This path never rotates `DataStoreIdentity.SourceIdentity` or checks that the replacement fingerprint differs, rejects a restored nonempty replacement through the initial-empty classifier, and produces another exact-baseline admission instead of eventual operational status. Implement the design's separate guarded replacement workflow and add tests rejecting a retained source identity and accepting the intended restored-source shape.

src/dms/backend/EdFi.DataManagementService.Backend.Cdc.Control/CdcSetupController.cs
```diff
@@ -0,0 +1,3195 @@
...
+                );
+            }
+
+            // The proof records whose assertion this is, because the worker made none. A reason that
+            // spoke for the worker here would read as evidence in a proof that outlives the record.
+            governedArtifacts.Add(
+                new CdcGovernedArtifact(
+                    CdcGovernedArtifactKind.ConnectSourceOffsets,
+                    inventory.ConnectorName,
+                    CdcCleanupState.NotFound,
```
bradbanister, line 2024: **High:** This records source offsets as `NotFound` solely from an operator assertion, after the code itself explains that connector absence is not evidence that its committed offsets are absent. The proof validator then permits deletion of the binding. A crash before the subsequent Kafka-volume removal leaves an unowned offset record that a reused connector name may resume. Persist objective cleanup progress or inspect/delete the raw offset state; do not delete the binding from this assertion alone.

src/dms/backend/EdFi.DataManagementService.Backend.Cdc.Control/CdcSetupController.cs
```diff
@@ -0,0 +1,3195 @@
...
+                    ),
+                    eligibility.Diagnostics
+                );
+            }
+        }
+
+        // Step 5: provider artifacts first, then the shared Connect offset store, then the binding's
+        // own topics and ACLs.
+        await using DbConnection connection = connectionFactory.Create(provider, request.ConnectionString);
+        await connection.OpenAsync(cancellationToken).ConfigureAwait(false);
```
bradbanister: **Medium:** Connection establishment occurs before the configured provider-setup budget and its timeout-to-contract handling. The same pattern appears in adoption and status, so an unreachable database can exceed the step budget or throw before a shared CDC result is produced. Apply one linked provider budget to connection creation/open plus the provider pass, as the retirement path already does.

src/dms/backend/EdFi.DataManagementService.Backend.Cdc.Control/CdcDownstreamPublicationHistoryProvider.cs
```diff
@@ -0,0 +1,219 @@
...
+        if (listResult.Status != CdcControlPlaneOperationStatus.Succeeded)
+        {
+            return Unknown(
+                targetKey,
+                currentPhysicalSourceFingerprint,
+                "The CDC binding state store could not be listed, so downstream publication history is unavailable."
+            );
+        }
+
+        if (listResult.States.Count == 0)
```
bradbanister: **Medium:** With an empty store this returns `Unknown`, while `InternalOnly` requires a readable binding for some other target. In the supported single-target bootstrap lifecycle that makes the positive state unreachable: before CDC and after retirement the store is empty, and while bound this target is active/historical. Add durable store-initialization/deployment-authority evidence so a verified authoritative empty store can prove no publication history without trusting an arbitrary mount.

reference/design/backend-redesign/epics/19-cdc-kafka/04-bootstrap-enable-kafka-cdc.md
```diff
@@ -42,12 +42,239 @@ needed to provision, validate, start, stop, and retire a target.
...
+  binding evidence proves the same normalized target key and physical-source fingerprint were
+  internal-only; active, historical, possible, unknown, missing, or mismatched evidence keeps
+  the E18 commands rejected with no mutation.
 - Add cluster-scoped Kafka Connect offset-store provisioning/validation and binding-scoped
   Kafka topic, durability, record-size, and ACL provisioning/validation.
 - Add Kafka Connect registration, live validation, status polling, restart, guarded
   adoption/source replacement, and teardown operations.
 - Expose the same workflow to the E2E harness.
 
+## Resolved Bootstrap CDC Scope and Integration Contract
```
bradbanister, line 56: **Medium:** This expanded section creates a second normative readiness/recovery contract, although `cdc-streaming.md` explicitly limits stories to implementation scope and acceptance evidence. The replacement and retirement rules now conflict with that owner, and the later claim that the published-image stack can honor this opt-in conflicts with the wrapper explicitly rejecting it. Move behavioral decisions into the owning design, reconcile them there, and keep this story scoped to components and evidence.

bradbanister:
Reviewed DMS-1323 against the normative CDC design and the existing main-branch contracts. The overall control-plane layering is sound, but five high-severity sequencing/recovery issues and three medium integration/design-contract issues should be resolved before merge. Targeted controller tests passed (42 tests), and rendered Compose configuration confirmed the default stacks emit blank indexed DocumentCache target keys. See the inline findings for the actionable details.

eng/docker-compose/local-dms.yml
JohnnyBrenes: **Not reproducible — the binder drops the element rather than materializing it.**

I checked this against the actual binding path before changing anything. `AddDmsDocumentCacheOptions` uses plain `.Bind()` (reflection binder — `EnableConfigurationBindingGenerator` is set nowhere in the tree), and `ConfigurationBinder.BindCollection` swallows the per-element conversion failure and never adds the item. So `Targets:0:DataStoreId=""` yields `Targets.Count == 0`, not a `Targets[0]` carrying an invalid `DataStoreId`, and `DocumentCacheOptionsValidator` never sees a target to reject. Verified with a standalone net10.0 probe against `Microsoft.Extensions.Configuration.Binder` 10.0.1 (the pinned version), then in-repo.

The claim in the compose comment was correct but untested, which is the real gap here. It is now locked by `DocumentCacheOptionsBindingTests.It_binds_blank_indexed_target_keys_to_no_projection_target`: blank `TenantKey`/`DataStoreId`/`RequiredRole` bind to zero targets, leave the DocumentCache status endpoint unmapped, and pass options validation. The last part matters because the DMS host validates on start, so a failure there would refuse to start every non-CDC stack — which is the regression you were reaching for.

No CDC-only profile added: there is nothing on the default path to move out of it.

src/dms/backend/EdFi.DataManagementService.Backend.Cdc.Control/CdcSetupController.cs
JohnnyBrenes: **Valid — fixed.**

Confirmed: both `Ensure*` results were assigned into `evaluation` and the sequence went straight on to plugin validation and `PutConnectorConfigAsync`. The nonconformance surfaced only in the admission composed at the end — that is, about a connector that was already registered and running against it. `cdc-streaming.md` is explicit that the store is validated "before accepting a worker, before connector registration or start/resume".

Added a fail-closed gate after each `Ensure*` call, returning `enableConnectOffsetStoreNotSatisfied` / `enableKafkaPolicyNotSatisfied` before anything is rendered or registered. Rather than restating the rule, the gate calls two new predicates on `CdcTargetStatusEvaluator` (`IsConnectOffsetStorePolicySatisfied`, `IsKafkaPolicySatisfied`) that reuse the evaluator's own `HasInvalidKafkaItem` / `HasUnknownKafkaItem` helpers, so the pre-registration answer and the composed status cannot drift apart.

New fixture `Given_CdcSetupControllerProvisioningGate` covers invalid *and* unknown for the store policy, the worker-only grants, the composed Kafka policy, and the record-size item, each asserting that neither `ValidateConnectorPluginConfigAsync` nor `PutConnectorConfigAsync` ran — plus that the binding's own artifacts are not provisioned when the shared store fails first.

eng/docker-compose/start-local-dms.ps1
JohnnyBrenes: **Valid — fixed. This one was load-bearing: `-EnableKafkaCdc` could not work at all.**

`cdc-streaming.md`: "Before bootstrap starts local Kafka Connect, it pre-creates and validates the configured shared offset topic and its worker-only ACLs … It never relies on Connect topic auto-creation or broker defaults." And `It_rejects_a_worker_created_offset_store_that_inherits_its_durability_from_the_broker` already proves the Connect-first topic is `Invalid`, because `min.insync.replicas` is not a topic-level override. Since `EnsureConnectOffsetStoreAsync` only creates when absent, that state was permanent.

Startup is now split in both paths that start the worker — the `-InfraOnly` `up kafka kafka-postgresql-source` and the full-stack `up` — into: start `kafka` alone, wait for the broker, provision the store, then start Connect. Provisioning creates the topic with `cleanup.policy=compact` plus an explicit `min.insync.replicas`, then applies both through `kafka-configs.sh --alter --add-config` whether or not the create found it present. That second step matters: a stack stopped without `-v` keeps the broker volume, so a run that opts into `-EnableKafkaCdc` after an earlier plain `-EnableKafka` run inherits a worker-created store, and repairing it is what makes the opt-in usable there. The policy is then read back and the run fails closed if it is not what the control plane requires.

The values are the `local` durability profile's own and live in `Get-LocalCdcDeploymentPolicy` next to `DurabilityProfile = 'local'`, so the script and the `cdc` verbs cannot name different ones; the topic name resolves from the same `DMS_CDC_CONNECT_OFFSET_STORAGE_TOPIC` the worker and the control plane read. Five Pester cases assert the ordering, both start paths, and the policy values; the compose README and the story doc record the ordering and why it matters.

src/dms/backend/EdFi.DataManagementService.Backend.Cdc.Control/CdcSetupController.cs
JohnnyBrenes: **Valid — partially fixed; the remainder is deliberately out of this PR, and I would rather say so than half-build it.**

You are right on the substance, and `cdc-streaming.md` backs both specifics: guarded replacement "rotates `SourceIdentity` through the binding-state operation", and "the new generation reports eventual operational status rather than another exact baseline".

Fixed now — the retained-source-identity case you asked for a test on. A restore, rollback, or copied backup carries the replaced database's own `dms.DataStoreIdentity` row, so its fingerprint is the outgoing generation's until the identity is rotated. `ReplaceSourceAsync` now refuses that with `SourceMismatch` / `observed: "retained"`, before the fence, so a replacement that replaced nothing cannot bind a second generation to one physical source. Test: `It_refuses_a_replacing_source_whose_identity_was_never_rotated`. The harness's previous-generation binding now carries its own distinct fingerprint — the realistic shape, and the reason three existing replace-source tests were passing on an identity that had not moved.

Not fixed here: (1) rotating `SourceIdentity` needs a write path to `dms.DataStoreIdentity` that exists nowhere in the repo — only the fingerprint *reader* does; (2) accepting a restored, non-empty source needs a replacement-specific classifier, since `EvaluatePreBindingEligibility` rejects any canonical/cache/work row by design; (3) "eventual operational status" means `ReplaceSourceAsync` stops returning `CdcAdmission`, which ripples through the CLI dispatcher, the exit-code mapper, the JSON contracts and their tests. That is a design-and-implementation package rather than a review fix. As it stands the verb gives a clear refusal instead of a wrong success.

src/dms/backend/EdFi.DataManagementService.Backend.Cdc.Control/CdcSetupController.cs
JohnnyBrenes: **Checked and not valid as stated — neither documented case can strand offsets.**

The switch is documented for exactly two shapes, and the control plane's own ordering rules both out:

- a connector that was never registered committed no offsets under that name;
- a connector removed by an earlier, interrupted retirement had its offsets deleted **first** — step (2) deletes them while the connector is stopped and still exists, and step (3) only then removes the configuration. So an absence this control plane produced is an absence of offsets too.

What is left is a connector deleted outside this control plane (Kafka UI, curl), which is precisely the judgement `--connector-already-absent` puts on the operator by name, and it is never defaulted.

The alternative you propose — inspecting or deleting the raw offset state — is ruled out by the design rather than by convenience: the offset topic grants literal `READ`/`WRITE`/`DESCRIBE` to the Connect worker principal only, and `It_fails_closed_when_the_offset_store_carries_a_grant_beyond_the_connect_worker` makes any grant beyond that `Invalid`. The control plane cannot be given the read access consuming it would require.

Behaviour unchanged. I did add the ordering argument as a comment at the branch and a sentence in the story doc, because the code previously said only whose assertion it was and not why the assertion is safe in the cases it covers — which is fair enough as a review finding.

src/dms/backend/EdFi.DataManagementService.Backend.Cdc.Control/CdcSetupController.cs
JohnnyBrenes: **Valid — fixed at all three sites.**

Confirmed: enable, adopt, and the shared status collection all did `connectionFactory.Create(...)` followed by `OpenAsync(cancellationToken)` outside any budget, and an unreachable database threw `DbException` straight out of the verb with no CDC contract produced at all. Retirement was the only path that already got this right.

Added `OpenProviderConnectionWithinBudgetAsync`, which creates and opens under a `CancellationTokenSource` linked to the caller and cancelled after `Timeouts.ProviderSetup`, mirroring the retirement path's catch set (`DbException`, `InvalidOperationException`, `ArgumentException`, and budget-only `OperationCanceledException`). Each caller now returns its own contract-shaped refusal — `enableProviderConnectionUnavailable`, an `AdoptionRefused`, `statusProviderConnectionUnavailable` — carrying only the rejection's type or `timedOut`, never the provider message, since those quote connection settings.

Three tests: a connection that never completes ends the enablement at the budget with `observed: "timedOut"` and no provider pass; a refused connection reports the exception type, with an assertion that the connection string never reaches the diagnostics; and the status path reports it as a readable status rather than throwing.

One deliberate narrowing of your suggestion: I kept each provider pass on its own budget rather than one budget spanning open plus both passes, so the existing create/validate semantics are unchanged and only the previously unbudgeted step joins them.

src/dms/backend/EdFi.DataManagementService.Backend.Cdc.Control/CdcDownstreamPublicationHistoryProvider.cs
JohnnyBrenes: **Checked — this is an explicitly adjudicated design trade rather than a gap, but your sharper version was not written down and now is.**

The story doc already owns the rule and its cost: "An empty listing rejects rather than proving the target was never published. A fresh volume, a mis-mounted state root, and a root pointed at another deployment all list empty, so an absent binding means something only once a readable one has established that this is the populated, authoritative store … that is the fail-closed side of the trade, and it matches the rule that no operator assertion may stand in for durable evidence."

Your observation goes further than what was recorded, and it is correct: in the single-target bootstrap lifecycle the positive state is not merely hard to reach, it is *unreachable* — the store is empty before enablement and after retirement, and holds only this target's binding in between. `internalOnly` is reachable for a target of a deployment that has bound some **other** target, which is the only shape in which an absent binding is evidence rather than an empty store. That consequence is now stated explicitly in the story doc.

I have not added the durable store-initialization / deployment-authority record you suggest: that is a new artifact in the binding state store and a change to the owning rule, not something to slip in on a review pass. Behaviour unchanged.

reference/design/backend-redesign/epics/19-cdc-kafka/04-bootstrap-enable-kafka-cdc.md
JohnnyBrenes: **Split verdict: the structural claim does not hold, but the contradiction you spotted was real and is fixed.**

On the section itself — "Resolved … Scope and Contract" is an established convention in this repo, not something this story invented. It appears in `10-update-tracking-change-queries/29-snapshot-support.md`, six `18-document-cache` stories, and two sibling `19-cdc-kafka` stories already on main (`01-cdc-ddl-support.md`, `02-connector-template-generation.md`). The section also opens by naming the owning documents as normative for eligibility, sequencing, topic policy, readiness, continuity and operations, and claims only the controller, its adapters, and the entry points. There is an existing Pester case asserting it carries such a section "in the sibling story's style", and another asserting the normative design carries no caveat about this story. Restructuring it would put this story out of line with its siblings, so I have left it as it is.

The published-image inconsistency was genuine. That bullet conflated two things and now separates them: the one-shot container is independent of the DMS image, so nothing in the *control plane* ties it to the local stack — and separately, the `-EnableKafkaCdc` switch is local-only, `start-published-dms.ps1` declares neither it nor the binding-state root, and the wrapper refuses it on the published path. A published deployment that wants projection sets the three DocumentCache variables in its own environment file, which is why `published-dms.yml` reads them the way `local-dms.yml` does. No claim that the published stack honours the opt-in remains.

Two other things from this review are folded into the same section: the offset-store pre-creation ordering, and the single-target consequence for the E18 gate.

PR #1220, opened 2026-09-02
[DMS-1060] Implement ownership-based authorization for GET-by-id, POST, PUT, and DELETE

samuellugo-ship-it:
## What

Implements ownership-based authorization for the single-record operations GET-by-id, POST, PUT and DELETE on relationally stored resources, per `auth.md` sections 2.13 and 2.14.

## Why

`OwnershipBased` was recognized but not enforced: a claim set configuring it fell through to a fail-closed 501. DMS-1373 already loads and caches `CreatorOwnershipTokenId` and `OwnershipTokenIds` from the tenant-qualified CMS `ApplicationContext`, so the values were available but nothing consumed them. This story consumes them, so a client configured with `OwnershipBased` is authorized against the stored `CreatedByOwnershipTokenId` rather than refused.

GET-many is deliberately left out and stays 501 — it is a page filter rather than a single-record check, and belongs to DMS-1410.

## Changes

Contracts and responses:
- `OwnershipAuthorizationFailure` with `OwnershipTokenMismatch` and `StoredOwnershipTokenUninitialized`, and the four `…FailureOwnershipNotAuthorized` result members on `GetResult`, `UpsertResult`, `UpdateResult` and `DeleteResult`.
- `OwnershipAuthorizationFailureResponse` producing the section 2.13 and 2.14 ProblemDetails, plus the dispatch arm in each of the four handlers. Neither body discloses a token value or the configured strategy index.

Planning and SQL:
- `OwnershipAuthorizationPlanner`, `OwnershipAuthorizationSqlCompiler` for both dialects, and the `own1` `AUTH1` payload family with its codec and dispatcher arm. The payload carries the CMS-configured strategy index, and a payload whose index does not match the planned check maps to a 500 rather than to a denial.
- `OwnershipTokenParameterization` binds one array parameter on PostgreSQL and one scalar per token on SQL Server; no TVP. `OwnershipTokenLimitExceededException` fails closed at 2,000 or more tokens on both dialects, guarded before deduplication so a padded list cannot slip under the cap.
- `RelationalAuthorizationPlanner` gains the enablement gate, the cap terminal, and its precedence: after both namespace terminals, ahead of every relationship terminal, and yielding to a custom-view configuration failure.

Execution:
- `OwnershipAuthorizationExecutor` and `OwnershipAuthorizationProviderFailureMapper`.
- `RelationalCompositeStoredAuthorization.TryAppendOwnership` places the statement after the custom-view and namespace statements and before the relationship statement, so statement order is precedence order. DELETE co-batches it into the command carrying the deletes, so an `AUTH1` abort leaves the row present. POST and PUT place it in the write's first-phase command, ahead of the second command that carries the DML. Both appends are budget-guarded and fall back to ordered same-session segments.
- Stamping on `RelationalDocumentRowCommandBuilder` applies to every create, including resources not configured with `OwnershipBased`; a missing creator token stamps null, and a PUT never changes the stored value.

Deliberate non-goals, each behind a named predicate rather than an accidental gap:
- Descriptors stamp but do not enforce — a descriptor action configured with `OwnershipBased` stays 501, via the `ResourceStorageKind.SharedDescriptorTable` arm of `EnforcesOwnershipChecks`.
- A POST resolving to create is never denied by ownership, pinned by the carrier row guard and by a wire-level 201 for a caller holding no tokens.
- GET-many and change queries keep their existing 501 and 500.

Documentation:
- `11-ownership-auth-strategy.md` acceptance criteria now separate the read path from the write and delete paths, separate descriptor stamping from descriptor enforcement, and record the budget-guarded segment fallback.
- `08-write-roundtrip-batching.md` records the GET-by-id one-added-command deviation next to the existing authorized-GET-by-id one, and qualifies the write/delete co-batch claim.

No schema change: `CreatedByOwnershipTokenId` already exists on `main`.

This PR also keeps the GET-by-id preflight terminal custom-view validation fix in DMS-1060 because ownership made the pre-existing accelerated-path gap reachable: the ownership token-cap terminal carries every configured custom view, since OwnershipBased executes last among AND strategies. Both regular and read-accelerated GET-by-id now complete preflight terminals through CompleteGetByIdPreflightStopAsync so custom-view validation behavior is consistent.

## Testing

About 16,000 tests, one failure, which is inherited — see below.

- Unit, all 11 projects: 14,138 passed.
- Backend integration `Category=Authorization`: 235 passed on PostgreSQL, 285 on SQL Server.
- API integration, full in-process HTTP pipeline on both providers: 418 passed, 1 failed.
- E2E shards 1 through 4: 909 passed, 0 failed, 13 skipped across 922 scenarios. No skip is an authorization scenario.
- `dotnet csharpier check src/dms` clean.

Each enablement gate was reverted individually to confirm its own test fails and no branch is covered only incidentally. That exercise found one branch covered only by its extracted predicate and not through the planner, and added the missing test.

## Expected CI noise, not hidden

`Given_DocumentCacheStatusEndpointProductionService.It_keeps_normal_api_routing_open_when_later_projection_work_makes_cdc_status_not_caught_up` fails with a 401 where it expects a 201. This is pre-existing on `main` and unrelated to this PR:

- Every file on that request path is byte-identical to `main` tip: `ApiService.cs`, `JwtAuthenticationMiddleware.cs`, `JwtRoleAuthenticationMiddleware.cs`, `ExternalDoublesRegistration.cs`, and the failing test itself.
- `main` is an ancestor of this branch, so `main` runs that test with identical inputs.
- The 401 is returned during authentication, upstream of anything this PR touches. The test's own fixture replaces `IJwtValidationService` with a document-cache status double and then makes a data-path request through it.
- It is invisible on `main` because the API integration project runs only in `on-dms-pullrequest.yml`, which fires on PRs.

It deserves its own ticket rather than a fix folded into this PR.

## Not included

Phase 9, Docker E2E ownership coverage, is optional and separable and is not part of this PR. It would add E2E step definitions for `POST /v3/ownershipTokens/` and `PUT /v3/apiClients/{id}/ownership` plus an `OwnershipAuthorization.feature`. It was scoped out because `AuthorizationDataProvider` may not currently surface the CMS apiClient id the assignment needs, and extending it is potentially broad. Enforcement is already proven at the unit, backend integration and API integration levels on both providers, and the existing E2E authorization shards prove no regression.

### DIFF
--- a/reference/design/backend-redesign/design-docs/ownership-token-maintenance.md
+++ b/reference/design/backend-redesign/design-docs/ownership-token-maintenance.md
@@ -275,6 +275,14 @@ the creator token in the read/modify collection.
 An ownership token may be assigned to multiple API clients within the same tenant. Sharing a token
 grants those clients access to documents stamped with that token.
 
+### Descriptors
+
+Descriptors: stamping is in scope, enforcement is not. Descriptor creates stamp
+`CreatedByOwnershipTokenId` like any other create, but descriptor GET-by-id, POST, PUT, and DELETE
+configured with `OwnershipBased` remain fail-closed at `501` in DMS-1060. Ownership enforcement for
+descriptors is not implemented in this story. The GET-by-id, PUT, and DELETE statements above
+therefore describe relationally stored resources.
+
 ## CMS Persistence Contract
 
 CMS will add the following PostgreSQL and SQL Server structures.
@@ -689,9 +697,13 @@ The approved handoff:
 2. adds the CMS application-context paragraphs above to both tickets: DMS-1060 consumes both
    ownership fields for POST stamping and single-record authorization, while DMS-1410 consumes
    `OwnershipTokenIds` for GET-many filtering only;
-3. retains DMS-1060's POST stamping and single-record authorization ProblemDetails, batching, and
+3. retains DMS-1060's POST stamping, single-record authorization ProblemDetails, and
    database-provider acceptance criteria, while assigning GET-many filtering and removal of the
-   temporary GET-many 501 only to DMS-1410; and
+   temporary GET-many 501 only to DMS-1410. Batching is not restated here: the accepted batching
+   deviations, including GET-by-id ownership running as one added command ahead of hydration, are
+   owned by the [DMS-1060 story mirror](../epics/14-authorization/11-ownership-auth-strategy.md)
+   and recorded in
+   [`08-write-roundtrip-batching.md`](../epics/07-relational-write-path/08-write-roundtrip-batching.md); and
 4. retains the defensive SQL Server failure at 2,000 or more tokens for both consumers.
 
 ## Acceptance Criteria Traceability
--- a/reference/design/backend-redesign/epics/07-relational-write-path/08-write-roundtrip-batching.md
+++ b/reference/design/backend-redesign/epics/07-relational-write-path/08-write-roundtrip-batching.md
@@ -237,6 +237,7 @@ has a recorded command count yet.
 | Namespace authorization only | 4 (adds the authorization command and `ShouldRetryPostHydrationReadBoundaryAsync`) |
 | Namespace and relationship authorization | 5 |
 | Custom-view authorization | unrecorded. A regular resource adds one command for the view-contract probe that precedes the membership run; the `cv1` statement itself joins the authorization command. A descriptor GET-by-id has no authorization command to join — it evaluates namespace in memory — so its membership query is an added command alongside that probe |
+| Ownership authorization | one command more than the same variant without it. The ownership check does not join the authorization command; see the deviation recorded below |
 
 ### Descriptor Writes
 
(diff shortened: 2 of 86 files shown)


samuellugo-ship-it:
Two accepted scope decisions for DMS-1060, recorded here so they are not re-raised as findings.

**Descriptor ownership: stamping is in scope, enforcement is not.** Descriptor creates stamp `CreatedByOwnershipTokenId` like any other create, but descriptor GET-by-id, POST, PUT, and DELETE configured with `OwnershipBased` remain fail-closed at `501` in DMS-1060. Ownership enforcement for descriptors is not implemented in this story.

**GET-by-id ownership runs as one added command ahead of hydration**, rather than inside the read roundtrip. POST and PUT run the check in the write's first-phase command with the DML as a later command, DELETE co-batches it with the deletes, and both arrangements are budget-guarded. This is the accepted deviation recorded in `reference/design/backend-redesign/epics/07-relational-write-path/08-write-roundtrip-batching.md`.

Both decisions are reflected in the DMS-1060 Jira acceptance criteria, `11-ownership-auth-strategy.md`, and `ownership-token-maintenance.md`. No implementation change is planned for either.

bradbanister (approved):

PR #1221, opened 2026-09-03
[DMS-1413] Define identity management design

simpat-adam:
Spike [DMS-1413](https://edfi.atlassian.net/browse/DMS-1413) defines the scope and story breakdown for
adding the Ed-Fi Identities API to DMS. **Design documents only — no application code.**

The result is an optional, pluggable pass-through API under `/identity/v2`. DMS owns the routes, the
Discovery entry, the OpenAPI document, authorization, the feature toggle, the pipeline, and the response
mapping; an implementer-provided plugin supplies the identity backend through the plugin architecture
[plugins-DMS-1462](reference/design/plugins-DMS-1462/design.md) designs. DMS ships no identity backend.

## Changes

| File | |
| --- | --- |
| `reference/design/identity-DMS-1413/design.md` | new — the companion design |
| `reference/design/identity-DMS-1413/README.md` | new — the manifest |
| `reference/design/identity-DMS-1413/01`–`08*.md` | new — eight story drafts, in dependency order |
| `reference/design/plugins-DMS-1462/README.md` | one line — points the `Identity companion` row at the new document |

Eleven Markdown files under `reference/design`. Nothing else is touched: no C#, project files, build
scripts, compose files, or tests.

## Design decisions worth a reviewer's attention

- **DMS owns the HTTP surface; the plugin owns the answer.** `IIdentityService` is a DMS-1462 `Replace`
  contract — a DMS-supplied `NoIdentityService` host default plus at most one plugin, with two claimants
  fatal. Plugins contribute no routes, which DMS-1462 already deferred citing DMS-1412.
- **`AppSettings:EnableIdentityManagement` is off by default.** When disabled the routes and all identity
  metadata are absent; when enabled without a plugin, DMS starts cleanly and every operation answers
  operation-unsupported `404`.
- **A capability miss is `404`, not ODS's `501`**, per the Jira acceptance criteria — and it carries its own
  problem-detail type so it stays distinguishable from tenant-not-found, identity-not-found, and feature-off
  `404`s.
- **Only find and search can answer `202`.** They return `IdentityAsyncResult`, which carries a request
  token; the other three return `IdentityResult`, which has no such member, so an async token from create,
  get-by-id, or results is unrepresentable rather than runtime-checked.
- **A 14-row divergence ledger** records every measured ODS/API v7.3 behavior DMS deliberately does not
  reproduce, so vendors get one table instead of a surprise.

## Story breakdown

Nine stories, filed from the companion README once this spike is approved. The graph has two lanes:
stories 01–04 build the contract, host default, pipeline, endpoints, document, and Discovery, and depend on
**no plugin foundation story at all**. Only story 05 (the plugin contract registry entry, behind
[DMS-1498](https://edfi.atlassian.net/browse/DMS-1498) and
[DMS-1499](https://edfi.atlassian.net/browse/DMS-1499)) and story 06 (the fixture-plugin proof) need the
loader. No story claims deploy-time plugin replacement before 05.

## Boundaries

[DMS-1414](https://edfi.atlassian.net/browse/DMS-1414) (UniqueId validation on person writes) stays
separate — the Jira description asked that this boundary be confirmed, and it is. Model 2 needs no work.

## Verification

No tests, builds, or restores were run: the change is Markdown only, and the repository has no Markdown
linter. What was checked instead:

- Every ODS/API claim is cited to raw source at the commit
  [ods-precedent.md](reference/design/plugins-DMS-1462/ods-precedent.md) pins
  (`90c75ffed3fc2bc0dafa14a2600b3a0d050f82e9`) or to a live `v7.3.2` response — not to a documentation
  summary.
- Published schemas were compared property-by-property and type-by-type against the pinned ODS identity
  OpenAPI document; four drifts were found and corrected during review (`Responses`, `Score` as
  `number`/`double`, `BirthDate` as `date-time`, and the `required` lists, the last recorded as a
  deliberate divergence).
- All 28 relative links resolve and every Markdown table has consistent pipe counts.

## Notes for reviewers

- The eight story drafts are **not yet filed** in Jira; the companion README's filing gate opens on approval
  of this spike.
- Four unrelated defects surfaced while designing and are recorded in the design's "Observations Outside
  This Ticket" as candidates for their own tickets: the inert `AppSettings:EnableManagementEndpoints` flag,
  `JwtRoleAuthenticationMiddleware` registered into no pipeline, two fixed-service facade paths that pass no
  cancellation token, and tenant existence having no single implementation.
- Open questions are listed in the design with a recommended default for each; none blocks the follow-up
  stories.

[DMS-1413]: https://edfi.atlassian.net/browse/DMS-1413?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DMS-1498]: https://edfi.atlassian.net/browse/DMS-1498?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


reference/design/identity-DMS-1413/design.md
```diff
@@ -0,0 +1,616 @@
...
+A tenant-independent pre-authorization mechanism would be a new design, not part of DMS-1413.
+
+## Tenant and Route-Qualifier Boundary
+
+`TenantValidationMiddleware` remains responsible for tenant presence and syntax.
+Add a Core identity-only `ValidateTenantExistsMiddleware` that uses the same cache-then-reload behavior as the frontend `TenantValidator`.
+When multitenancy is enabled and the tenant does not exist, it returns `404` without calling the claim-set provider or identity provider.
+When multitenancy is disabled, it is a pass-through.
+
+`ResolveDataStoreMiddleware` stays omitted.
```
bradbanister, line 587: **High:** Omitting datastore resolution does not establish that the authenticated client is allowed to address the URL tenant. JWT validation currently extracts client authorizations without binding them to the tenant, so a token from tenant A could reach tenant B when both tenants expose the same claim-set name. Add an explicit client-to-tenant authorization check before service-claim evaluation and test two tenants with identical claim sets.

reference/design/identity-DMS-1413/design.md
```diff
@@ -0,0 +1,616 @@
...
+- Runtime response-payload validation: do not add it in v1.
+- Tokens containing path separators: keep rejected unless a future story measures deployment behavior and changes the contract.
+
+## Follow-Up Items Outside Identity Scope
+
+- `AppSettings:EnableManagementEndpoints` appears to be configured but not read by production code.
+- `JwtRoleAuthenticationMiddleware` is registered but not composed into a pipeline.
+- Existing fixed-service facade paths pass no request cancellation token.
+- Tenant existence should eventually be unified behind one Core-side helper used by both frontend fixed routes and the identity Core middleware.
+- The Configuration Service claim-set provider is a singleton that mutates one shared `HttpClient`'s `Tenant` and `Authorization` default headers per call, while the claim-set cache's stampede lock is keyed per tenant. Concurrent first-time misses for different tenants can therefore race on those headers. Identity service-claim authorization adds another caller to that path.
```
bradbanister: **High:** This existing shared-HttpClient header race is on the Identity authorization path, not merely a cleanup item: concurrent first-time claim-set loads for different tenants can send the wrong Tenant header and cache the result under the wrong tenant key. Make per-request HttpRequestMessage headers and a concurrent-miss test a prerequisite for multitenant Identity support.

reference/design/identity-DMS-1413/design.md
```diff
@@ -0,0 +1,616 @@
...
+## Authorization
+
+The service claim already exists in shipped claim documents:
+
+```text
+http://ed-fi.org/identity/claims/services/identity
+```
+
+DMS does not need a CMS migration for this ticket.
+The identity authorization middleware should use the existing claim-set graph and map operations to actions:
```
bradbanister, line 442: **Medium:** The design maps the service claim to Create/Read but does not define how configured authorization strategies are handled. A claim action can carry NamespaceBased, relationship, unknown, or empty strategies; silently checking action membership would bypass CMS security metadata. State which strategy is supported for Identity (likely NoFurtherAuthorizationRequired) and fail closed with the existing security-configuration behavior for the rest.

reference/design/identity-DMS-1413/design.md
```diff
@@ -0,0 +1,616 @@
...
+
+public sealed record IdentityAsyncResult
+{
+    public required IdentityResultStatus Status { get; init; }
+    public JsonNode? Payload { get; init; }
+    public string? RequestToken { get; init; }
+    public IReadOnlyList<IdentityError> Errors { get; init; } = [];
+}
+
+public interface IIdentityService
```
bradbanister, line 169: **Medium:** The public contract does not specify the lifetime or resolution boundary for IIdentityService. ApiService is a singleton and its pipelines are cached, while the plugin design permits scoped registrations; constructor injection would either fail scope validation or capture a request-scoped service. Specify allowed lifetimes and resolve the provider from the request scope, then test a provider with a scoped dependency.

reference/design/identity-DMS-1413/design.md
```diff
@@ -0,0 +1,616 @@
...
+4. expected JSON shape: object for create, array of strings for find, and array of objects for search.
+
+The two outbound protocol checks are:
+
+1. required values are present for the status returned;
+2. an async request token can be carried in one URL path segment.
+
+## Async Token Rule
+
+Find and search may return asynchronous `Success` with a request token instead of a payload.
```
bradbanister, line 724: **Medium:** The token syntax is well defined, but async results have no contract for ownership, tenant/qualifier binding, expiration, retention, terminal failure, restart behavior, or replica portability. A provider could expose a global token store or lose every job on restart. Require context binding and document the lifecycle; include client/tenant isolation and multi-instance/restart tests.

reference/design/identity-DMS-1413/design.md
```diff
@@ -0,0 +1,616 @@
...
+
+That ordering is deliberate.
+Unauthenticated callers still receive `401` and do not reach tenant existence.
+Authenticated callers with a valid token can receive tenant `404` before service-claim `403`; that is accepted because any middleware that calls `IClaimSetProvider.GetAllClaimSets(tenant)` must not run before DMS knows the tenant exists.
+A tenant-independent pre-authorization mechanism would be a new design, not part of DMS-1413.
+
+## Tenant and Route-Qualifier Boundary
+
+`TenantValidationMiddleware` remains responsible for tenant presence and syntax.
+Add a Core identity-only `ValidateTenantExistsMiddleware` that uses the same cache-then-reload behavior as the frontend `TenantValidator`.
```
bradbanister: **Medium:** Reusing TenantValidator's cache/reload behavior also reuses its failure collapse: general Configuration Service errors are caught and returned as false, which would turn an upstream outage into tenant-not-found 404 (and can swallow cancellation). Define typed exists/absent/unavailable outcomes so only confirmed absence maps to 404 and upstream failure remains distinguishable.

reference/design/identity-DMS-1413/design.md
```diff
@@ -0,0 +1,616 @@
...
+4. Let DMS broker requests and responses without interpreting identity semantics, match scores, or custom person fields.
+5. Support synchronous `200` responses and asynchronous `202` responses with a `Location` header and later results polling.
+6. Publish OpenAPI and Discovery entries only when the feature is enabled.
+7. Break the work into stories that respect the current plugin-foundation dependency graph.
+
+### Non-Goals
+
+- No concrete identity-system integration and no DMS-shipped identity backend.
+- No person matching, scoring, or UniqueId issuance logic in DMS.
+- No validation of UniqueIds on person-resource writes; that belongs to DMS-1414.
```
bradbanister: **Medium:** Deferring UniqueId validation to DMS-1414 is reasonable, but the Identity contract still needs issuance compatibility rules. The backend redesign has different equality semantics across PostgreSQL and SQL Server; a provider issuing IDs that differ only by case can collide with a person natural key on SQL Server. Document uniqueness scope, casing/equality, length, character, and stability requirements, and exercise an issued ID through person-resource writes on both engines.

reference/design/identity-DMS-1413/design.md
```diff
@@ -0,0 +1,616 @@
...
+- The full plugin path cannot be proven until DMS-1499 exists.
+- `404` has several meanings: feature off, tenant not found, unsupported operation, and subject not found. Tests must distinguish by route presence or problem-detail type.
+- The create response differs from running ODS code by omitting `201 Location`.
+- A published contract is permanent and additive-only.
+- Feature-toggle rot is a real risk; absence tests must cover routes, metadata, listing, and Discovery.
+- `IApiService` grows five methods. Interface fakes absorb that automatically, but the `ApiService` constructor gains a dependency and 13 test sites construct that class directly.
+- Plugins run fully trusted in process.
+- Capabilities are deployment-wide in v1.
+- Parameterizing content-type validation touches existing resource write pipelines.
+- Payload shape is documented but not enforced at runtime.
```
bradbanister, line 1103: **Medium:** Because response payloads are intentionally opaque at runtime, fixture tests alone cannot protect implementers from serving unusable payloads. Provide reusable provider-conformance tests (or an executable schema validator outside the request path) covering the documented required fields, status shapes, ordered groups, and search scores.

reference/design/identity-DMS-1413/design.md
```diff
@@ -0,0 +1,616 @@
...
+    public JsonNode? Payload { get; init; }
+    public string? RequestToken { get; init; }
+    public IReadOnlyList<IdentityError> Errors { get; init; } = [];
+}
+
+public interface IIdentityService
+{
+    IdentityCapabilities Capabilities { get; }
+
+    Task<IdentityResult> CreateAsync(
```
bradbanister, line 173: **Medium:** Create is a side-effecting operation, but the contract has no idempotency or reconciliation rule. If the provider allocates a UniqueId and the response is lost, retrying can allocate a second identity. Define whether a client-supplied idempotency key is supported, how retries are correlated, and how an ambiguous create is reconciled; TraceId alone is not sufficient.

reference/design/identity-DMS-1413/02-add-identity-api-surface-pipeline-toggle-openapi-discovery.md
```diff
@@ -0,0 +1,84 @@
...
+
+### Frontend Endpoints and Feature Toggle
+
+- `AppSettings:EnableIdentityManagement` exists and defaults to `false`.
+- The five identity routes are mapped only when the feature is enabled.
+- When disabled, all five routes answer routing `404`.
+- The route prefix supports the same tenant and route-qualifier prefixing used by other fixed-service endpoints.
+- `GET /identity/v2/identities/results` routes to get-by-id with id `results`.
+- `GET /identity/v2/identities/results/{token}` routes to results polling.
+- POST handlers request body parsing through the existing frontend body extraction path.
```
bradbanister, line 99: **Low:** The story says capability and authorization run before body parsing, but the required frontend path already buffers and parses JSON in AspNetCoreFrontend before Core receives RequestInfo. This ordering cannot be achieved by Core middleware alone. Clarify that authorization gates body-validation responses, or redesign frontend extraction if parsing itself must be deferred.

bradbanister:
Static design review: the Identity proposal integrates cleanly with the relational redesign, but it has 10 issues to resolve before implementation: tenant binding, multitenant claim-set safety, authorization strategy semantics, provider lifetime, async job isolation/lifecycle, typed tenant failure handling, UniqueId compatibility, provider conformance/idempotency, and the stated pipeline ordering.

reference/design/identity-DMS-1413/design.md
simpat-adam: Agreed, and fixed in `6d8c105bd`.

The design now adds an explicit client-to-tenant binding step between tenant existence and service-claim authorization, using `IApplicationContextProvider.GetApplicationByClientIdAsync(clientId, tenant)`. That provider already sends the tenant per request and already distinguishes "client absent from this tenant" from "Configuration Service unavailable", so it supplies the binding and the typed outcomes together: `401` for a client not resolvable in the tenant, `503` for an unanswerable check. See the "Client-to-Tenant Binding" section.

Your framing that the omission *removed* the check rather than merely not adding one is what made this land. `ResolveDataStoreMiddleware` was the only thing binding a client to the URL tenant, and it was incidental.

One subtlety worth recording, because it is the way an implementation would plausibly get this wrong: the lookup is client-scoped, not datastore-scoped. `ApplicationContext.DataStoreIds` may legitimately be empty, and an implementation that rejected an empty list would break identity-only clients - the exact client shape this API exists to serve. The design and the API-surface story both say so explicitly, with a test.

The two-tenant test you asked for is an acceptance criterion: two existing tenants with identically named claim sets both granting identity access, asserting tenant A's token cannot reach tenant B and that the identity provider is never called.

reference/design/identity-DMS-1413/design.md
simpat-adam: Agreed, and promoted further than a cleanup item: it is now a named prerequisite with its own story draft, `00-send-tenant-header-per-request-in-cms-claim-set-provider.md`, and the identity API-surface story declares it as a dependency.

It is filed separately rather than folded into the feature story because it is a security defect on the existing resource authorization path, so it should be reviewable and releasable on its own rather than buried in identity work.

Verifying it turned up one detail worth adding: `ConfigurationServiceClaimSetProvider` is the lone outlier in Core. The application, data store, and profile providers already use per-request `HttpRequestMessage` headers, the last with an explicit comment saying it does so for thread safety. So this is a one-file change onto an established in-repo pattern, not a new approach.

Also worth noting for scheduling: the window is not the startup path. `CacheClaimSetsTask` warms tenants sequentially at boot, so concurrent cold misses happen after cache expiry or invalidation, which recurs for the life of the process. The concurrent two-tenant cold-miss test is an acceptance criterion.

reference/design/identity-DMS-1413/design.md
simpat-adam: Agreed, and specified in `6d8c105bd` under "Authorization Strategies".

v1 policy: an action whose effective strategy list is exactly `NoFurtherAuthorizationRequired` is authorized. An empty list, an unrecognized strategy name, and any recognized strategy other than `NoFurtherAuthorizationRequired` all fail closed as invalid security configuration - a `500`, explicitly distinct from a `403`. That follows `ResourceActionAuthorizationMiddleware`, which already validates the strategy list rather than treating action membership as the whole decision, and it preserves the redesign's separation of permission denial from invalid configuration.

Your point about CMS `authorizationStrategyOverrides` is the reason this needed stating at all: the shipped claim declares NFAR on every action, so without the override path it would have looked like a non-question.

One related gap your comment surfaced indirectly: the shipped claim declares NFAR on Create, Read, **and Update**, but no identity operation maps to `Update`. That was unstated and is now explicit.

reference/design/identity-DMS-1413/design.md
simpat-adam: Agreed, and specified in `6d8c105bd` under "Provider Lifetime and Resolution".

All three lifetimes are supported. `IIdentityService` is resolved once per request from `RequestInfo.ScopedServiceProvider`, following the existing `UpsertHandler` pattern, and is explicitly not a constructor dependency of the singleton `ApiService`. DMS never disposes the instance itself. The contract also states that a provider must be safe for concurrent calls across requests, since a singleton or transient registration gives DMS no per-request isolation, and that a scoped registration is called at most once per request but its captured dependencies may still be shared.

Your "test a provider with a scoped dependency" is an acceptance criterion, and the design says why the weaker version is not enough: a singleton test fixture cannot detect a captured-scope defect. Worth flagging that ASP.NET scope validation is Development-only, so a captured-scope bug of this shape fails on a dev box and ships silently otherwise - which is precisely why the fixture has to be scoped rather than the check relying on the framework.

reference/design/identity-DMS-1413/design.md
simpat-adam: Agreed, and this was the most useful of the medium findings, because the token rules were purely syntactic - nothing bound a token to its issuing context.

Fixed in `6d8c105bd`: `IdentityRequestContext` now carries a required `ClientId` alongside `Tenant` and `RouteQualifiers`, and a new "Async Job Obligations" section states what the provider owns: bind each accepted job to the tenant, qualifiers, and client of the creating request and answer a mismatched poll as `NotFound`; document retention and post-expiry behavior; answer repeated polls of a complete job identically, since polling is a `GET`; represent terminal failure as its own answer rather than indefinite `Incomplete`; make results retrievable from any replica or document the integration as single-replica; and document restart behavior.

The v1 decision is that results are scoped to the issuing client rather than shared across the tenant. That direction is deliberate: loosening it later is a compatible change, tightening it later is not.

Isolation and restart tests are acceptance criteria, including the fixture-plugin alternative where the fixture documents itself as in-memory only and the test asserts the documented `404`.

reference/design/identity-DMS-1413/design.md
simpat-adam: Agreed on the diagnosis, and the fix went further than the original comment after a later pass.

`6d8c105bd` replaced the bare `bool` with typed exists / absent / unavailable outcomes, where only confirmed absence maps to `404`, an unanswerable check is `503`, and `OperationCanceledException` for the request's own token propagates rather than being converted into an outcome.

A subsequent review round showed the typed enum alone was not sufficient, and `adc7ff4e9` corrects the mechanism. The confirmed-absent signal never reached the validator in the first place: `LoadDataStores` rewraps every `HttpRequestException` as `InvalidOperationException` before it can propagate, and `ConfigurationServiceResponseHandler` throws with the message-only constructor so `StatusCode` is null regardless. `TenantValidator`'s `catch ... when (ex.StatusCode == NotFound)` clause is therefore unreachable dead code today, and a typed result over that path would have been a typed result over a single undifferentiated failure.

The identity check now uses the existing `IDataStoreProvider.LoadTenants()` tenant-name path instead: names only from `v3/tenants/`, no `Tenant` header, no datastore projection, no connection-string decryption. Confirmed absence is set membership over a successfully fetched list rather than an error status, so no Configuration Service change and no change to the shared response handler is needed.

That also removed a coupling worth naming: the datastore reload decrypts each store's primary connection string and fails the whole tenant load on an undecryptable primary, so reusing it would have let an unrelated datastore misconfiguration block identity requests that need no datastore at all.

The response-handler status discard and the dead catch clause are recorded as defects outside this ticket's scope.

reference/design/identity-DMS-1413/design.md
simpat-adam: Accepting most of this and pushing back on one part.

**Accepted.** The issuance constraints are now in the contract, extended in `adc7ff4e9` beyond what `6d8c105bd` first added:

- Length. Person resources declare `"maxLength": 32` for `studentUniqueId`, `staffUniqueId`, and `contactUniqueId` - every occurrence, across core DS 5.2 and the shipped TPDM and sample extensions. The wire type is an unbounded string, so a provider could satisfy every other rule and still issue ids no person write accepts. A canonical 36-character GUID does not fit; the same GUID hyphen-free does. The rule is stated as "fits the `maxLength` the deployment's ApiSchema declares", with 32 as its current value, rather than as a contract constant - the identity package versions independently of both the DMS release and the Data Standard, so freezing a data-model constant into an additive-only public artifact would have it assert something about a model it does not own.
- Equality. Your casing point is right, and case-insensitive distinctness turned out to be necessary but not sufficient. `OrdinalIgnoreCase` is documented in the redesign as an in-process approximation of `SQL_Latin1_General_CP1_CI_AS`, not an emulator, and the accepted residue records equivalences beyond case. The contract therefore defines a guaranteed repertoire - ASCII letters and digits - inside which the approximation is exact, and states that values outside it move the uniqueness obligation to the provider under each store's actual equality.
- Uniqueness scope, whitespace, single-path-segment, and stability are all stated.

**Pushed back:** exercising an issued id through person-resource writes on both engines. That is DMS-1414's validation path, and this design deliberately adds no ApiSchema or relational dependency to the identity pipeline. Putting the cross-backend write test here would require the epic to take on a dependency it otherwise does not have, to assert something the owning story will assert anyway. This design contributes the constraint that story asserts against.

reference/design/identity-DMS-1413/design.md
simpat-adam: Pushing back on this one, though the underlying concern is real and is recorded rather than dismissed.

A published conformance suite would be a new public artifact with its own version, publication lane, and additive-only policy - a different kind of commitment from the contract package, and one that should be scoped and scheduled deliberately rather than absorbed into this epic. It is recorded as a deferred follow-up, and it belongs to its own story if it is wanted.

What implementers get in v1: the served `/metadata/identity/v2/swagger.json` document is the artifact they validate payloads against, story 04's implementer chapter documents the payload obligations, and the fixture-plugin story asserts fixture payloads validate against the served schemas.

A middle option was considered and set aside in `adc7ff4e9`: embedding JSON Schema documents and examples in the contract package itself. It was briefly accepted, then reverted, because the served OpenAPI document already carries the schemas and now also carries pinned `400` example bodies for each error-projection shape, so embedding a second copy would add a public artifact to keep in sync for coverage the epic already has.

If you think the deferral is wrong, the concrete question is whether a runnable suite should gate v1 publication of `EdFi.Api.Identity` - happy to file it as a blocking story if so.

reference/design/identity-DMS-1413/design.md
simpat-adam: Agreed that the hazard is real, and it is now documented explicitly rather than left implicit - but v1 deliberately does not add an idempotency key.

The "Timeout, Retry, and Idempotency" section states the consequence in the terms you describe: DMS wraps no resilience pipeline, so it never retries a provider call and imposes no timeout of its own. If the upstream system issues a UniqueId and the response to DMS is lost, the client sees `502` and a retry may issue a second id for the same person. `TraceId` is per-request and is explicitly not an ownership, cache, or idempotency key, so it cannot deduplicate the retry.

The contract instead requires each provider to document how its integration behaves on a repeated create for the same identifying data: either creation is idempotent on some upstream key, or duplicate issuance is possible and the reconciliation path is stated. An implementer whose upstream offers no such guarantee must say so, because a client cannot otherwise know whether retrying a failed create is safe.

The reasoning for not adding a host-level key in v1: DMS does not hold identity state and cannot deduplicate on the client's behalf, so a key it accepted but could not enforce would be worse than an documented absence. This is consistent with the product requirement that DMS does not prevent duplicate identity creation. It stays prominent in implementer and client guidance, and a client-supplied idempotency key remains an additive change if a provider ecosystem later wants one.

reference/design/identity-DMS-1413/02-add-identity-api-surface-pipeline-toggle-openapi-discovery.md
simpat-adam: Correct, and the wording is fixed in `6d8c105bd`.

The claim was true of *reporting* but not of *parsing*. `AspNetCoreFrontend.ExtractJsonBodyFrom` parses the request body and records duplicate-property paths before Core is entered; `ParseBodyMiddleware` and `DuplicatePropertiesMiddleware` report what it already computed rather than performing the parse themselves. So authorization genuinely precedes the reporting of body-parse and duplicate-property errors, but it does not precede parsing, and the design previously implied it did.

Both `design.md` and the API-surface story now say this, and the story carries an acceptance criterion stating that no criterion claims parsing itself is ordered after authorization - so the corrected wording cannot quietly regress into the original claim.

bradbanister (approved):

bradbanister (approved):

PR #1235, opened 2026-09-09
[DMS-1430] Store the PostgreSQL token expiration as an instant

samuellugo-ship-it:
## What

Stores `dmscs.OpenIddictToken.ExpirationDate` as a PostgreSQL instant (`timestamp with time zone`) instead of a session-time-zone wall clock, so the expired-token cleanup sweep no longer depends on the server's time zone.

## Why

DMS-1354 shipped the column as `timestamp without time zone` while both repository paths bind `DateTimeOffset` values, which Npgsql sends as `timestamptz`. The two paths then converted in opposite directions: the insert cast the instant down to a wall clock through the session time zone, and the cleanup predicate cast the stored column back up through it via `timestamp_le_timestamptz`. Those conversions do not round-trip on a DST-observing server. DMS-1354 recorded the issue and deliberately deferred it, because a correct fix has to move the insert, the sweep bound, and the column type together.

SQL Server was never affected: it compares `DATETIME2` against a UTC `DateTime` with the parameter type pinned explicitly.

### Correction to the failure direction in the ticket

The ticket describes this as the sweep deleting a row the validator still accepts, twice a year. That is not reproducible under a *stable* DST-observing zone, and the investigation is recorded in the design record and on the ticket. Every edge case in the reverse conversion resolves later or equal - a fall-back wall clock is ambiguous and PostgreSQL picks the later candidate instant, a spring-forward gap resolves forward - so the sweep could only *under*-delete, leaving rows to linger past their expiry.

Premature deletion of a live token is real but requires the session zone to *differ* between the write and the sweep (an operator changing the server `timezone`, setting `PGTZ`, or adding `Timezone=` to one connection string). Reproduced during investigation: a row written under `America/New_York` expiring 18:00Z, swept under UTC at 14:00Z, deleted four hours early. Both defects are removed by this change; the distinction affects only how the pre-fix severity is described.

## Changes

Schema (`ee7e45b6b`):

- `0016_Create_openiddict_Token_Table.sql` declares `ExpirationDate` as `timestamp with time zone` for clean installs.
- New `0032_Alter_OpenIddictToken_ExpirationDate_TimeZone.sql` migrates existing installs, PostgreSQL only. The `ALTER` is wrapped in a conditional on the current column type, so the `ACCESS EXCLUSIVE` rewrite runs at most once per database and is a no-op on fresh installs and under the shape test's script replay.
- The script deliberately carries no `USING` clause, and says why: the implicit assignment cast is the inverse of how Npgsql wrote the rows, whereas `USING "ExpirationDate" AT TIME ZONE 'UTC'` corrupts data on any non-UTC server.
- `DatabaseShapeTests` asserts the declared column type, pinning the DDL against regression.

Repository bindings (`bc76cb132`):

- `StoreTokenAsync` and `DeleteExpiredTokensAsync` both pin their parameter to `DbType.DateTimeOffset`, normalized with `ToUniversalTime()`, in a single commit. `StoreTokenAsync` moves to `DynamicParameters` to allow the pin, matching the SQL Server implementation's shape and closing the cross-engine asymmetry noted in DMS-1411 AC7.
- This is expressiveness and defence in depth rather than a behavior change: Npgsql already maps `DateTimeOffset` to `timestamptz`, so against the new column type the pinned and unpinned forms are equivalent. The column type carries the correctness. Verified against PostgreSQL's bind-parameter log under a non-UTC session, where a UTC session cannot distinguish the two wire types.
- `CreationDate` is restated with the same value and no explicit `DbType`, so its behavior is unchanged.

Tests (`c3e0f0b6a`):

- `Given_A_DST_Observing_PostgreSQL_Session_Time_Zone` runs the real repository against an `America/New_York` session connection, storing two expirations an hour apart that both land on local wall clock 01:30 across the 2026 fall-back. Asserts distinct stored instants, a sweep that deletes only the truly expired row, and an unchanged read-back through `GetTokenByIdAsync`.
- `Given_a_pre_DMS_1430_OpenIddictToken_PostgreSQL_upgrade` exercises `0032` against a journaled pre-upgrade database: exact recovery under an unchanged zone, `NULL` preservation, replay safety, and the two cases the migration cannot repair.

Documentation (`8f2c2c3b5`):

- `TOKEN-CLEANUP.md` gains a "Time-Zone Independence (DMS-1430)" section and its stale Current State paragraph is corrected. Records the corrected failure direction, the three migration cases, the bounded one-time rewrite cost, the deliberate scope limit, and the new coverage.

## Testing

| Suite | Result |
| --- | --- |
| PostgreSQL CMS integration (full) | 427 passed, 0 failed, 0 skipped |
| SQL Server CMS integration (full) | 432 passed, 0 failed, 0 skipped |
| SQL Server OpenIddict cleanup (targeted) | 6 passed, 0 failed, 0 skipped |

SQL Server ran against a temporary `mcr.microsoft.com/mssql/server:2025-latest` instance, so the cleanup coverage was genuinely executed rather than skipped. CSharpier is clean on all four touched C# files.

Fault injection confirms each half of the fix is independently covered, and no injected edit remains in the tree:

| Injected fault | Result |
| --- | --- |
| Column reverted to `timestamp without time zone`, code untouched | all 3 DST tests fail |
| Insert misbound as a wall clock, sweep left correct | all 3 fail |
| Sweep bound misbound as a wall clock, insert left correct | only the deletion test fails; storage and read stay green |

The third row is the one-sided change the ticket asked the suite to catch.

## Residual risks

- **Existing rows in a DST fall-back window are not repaired.** Two instants an hour apart were already stored identically by the old column type, so the migration recovers both as the later one. The write destroyed that distinction; the migration cannot recover it. The error direction is late, so such a token lingers rather than being swept early.
- **A database whose session zone changed since its rows were written reconstructs against the migration-time zone**, which can place a value *earlier* than the true instant. The zone a row was written under was never recorded, so the migration cannot detect this. Mitigation is operational: upgrade under the zone the rows were written under. On the shipped UTC containers this case does not arise.
- **`CreationDate` and `RedemptionDate` remain wall-clock columns** with the same latent dependence. Deliberately out of scope: no predicate compares them and no production path reads them. Recorded as possible follow-up cleanup.
- **The DST tests depend on stable `America/New_York` 2026 tzdata.** US DST rules are stable and the assertions are behavioral rather than on PostgreSQL internals, but a tzdata change to US DST rules would move them.

Full detail is in `reference/design/configuration-service/TOKEN-CLEANUP.md`, section "Time-Zone Independence (DMS-1430)", and in the DMS-1430 ticket description under Implementation Decision.

### DIFF
--- a/reference/design/configuration-service/TOKEN-CLEANUP.md
+++ b/reference/design/configuration-service/TOKEN-CLEANUP.md
@@ -49,14 +49,13 @@ A row expired beyond that skew is therefore unreachable by the only production r
 The DMS data API never reads the table at all: there are zero `OpenIddictToken` references
 anywhere under `src/dms`, because the data API validates bearer tokens statelessly via JWKS.
 
-Timestamps are UTC wall-clock values, stored as `timestamp without time zone` on PostgreSQL and
-`DATETIME2` on SQL Server.
+`ExpirationDate` is an instant, stored as `timestamp with time zone` on PostgreSQL and `DATETIME2`
+holding UTC on SQL Server.
 Any deletion predicate must therefore compare against UTC "now", never local time.
-On PostgreSQL the UTC-wall-clock property holds because writes pass `DateTimeOffset` values that
-Npgsql sends as `timestamptz`, which the server converts through the session time zone; the
-shipped containers run UTC, and the sweep's bound takes the identical conversion path as the
-pre-existing insert, so the two stay consistent under any single session time zone (see the
-latent observation below).
+As shipped by DMS-1354 the PostgreSQL column was `timestamp without time zone`, which made both the
+insert and the sweep convert through the session time zone; DMS-1430 replaced that with the
+`timestamptz` declaration described under Time-Zone Independence below, so neither path converts
+any more.
 A deletion predicate is already index-supported on both engines: `IX_OpenIddictToken_ExpirationDate`
 exists in both engines' DDL, in `0016_Create_openiddict_Token_Table.sql`.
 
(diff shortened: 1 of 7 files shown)


simpat-adam (approved):

samuellugo-ship-it:
Rebased onto latest main; the new token-limit count bound (`ActiveAsOf`) now binds as an instant alongside `ExpirationDate`.

simpat-adam (approved):

PR #1236, opened 2026-09-09
DMS-1457: Normalize the correlation ID once, and apply it everywhere

stephenfuqua:
> [!WARNING]
> This is a "duplicate" pull request, performing a head-to-head comparison of Claude Code on the web with GitHub Copilot Coding Agent in #1232. Only one of them will merge. Still working on the determination. SF Sep 9 11:49 AM

## Why

DMS accepts a client-supplied correlation ID from a host-configured request header and uses it to tie a client-visible error response to the server's log entries. That value was normalized inconsistently: the frontend request-logging layer applied a character filter and a length cap, but most error response bodies echoed the value exactly as the client sent it, and the Core and Backend layers re-narrowed it through a *stricter* allowlist before logging. The ID a client read from a failed request was therefore not reliably the ID that appeared in the logs — defeating the purpose of the correlation ID — and an unbounded, unfiltered client-controlled string reached multiple response paths.

This makes normalization uniform and single-sourced: one function, applied once at ingestion, so every downstream consumer — every log event and every error response body — carries the identical value.

Implements **FR-LOG-3** through **FR-LOG-6** of `docs/PRD-v8.1.md` §3.11.

## What changed

**One normalizer, one ingestion point.** New `CorrelationIdNormalizer` (the only new type) holds the whole policy. `AspNetCoreFrontend.ExtractTraceIdFrom` is the single place a `TraceId` is built from request data, so `FailureResponse`'s ~30 factories, all handlers and all log events receive an already-correct value without changes.

**A correlation-ID-specific allowlist, kept distinct from the strict one.** New `LogSanitizer.SanitizeCorrelationIdForLog`, surfaced as `LoggingSanitizer.SanitizeCorrelationIdForLogging`. `SanitizeForLogging` is behaviorally unchanged and `Method`/`Path`/`PathBase` still use it, which is correct for internally-controlled values.

**Normalization order: truncate first, then filter — deliberately.** This matches what `LoggingMiddleware` already did, and it is pinned by a test that fails if the order is swapped. A consequence, intended and documented: a value that is both over-length and contains disallowed characters ends up *shorter* than `CorrelationIdMaxLength`.

**The allowlist is all printable non-control characters** — a bare `!char.IsControl(c)` with no positive character enumeration. Every control character is removed, `\r`, `\n`, `\t` and `\0` included; everything else is preserved, including `+ = { } @ | , # ( ) [ ] < > " '` and non-ASCII. That is the point of FR-LOG-3: a client-supplied correlation ID normally originates in an upstream system's own identifier scheme, so narrowing to alphanumerics would defeat accepting one at all. The leading `ReplaceLineEndings` call also removes U+2028 and U+2029, which `char.IsControl` does not report — that call is load-bearing here (unlike in `SanitizeForLog`) and is documented as such.

**Applied at every use site, not just the frontend.** 32 log sites across 9 files in Core and Backend were re-narrowing the already-normalized value through the strict `Method`/`Path` allowlist, so a Core- or Backend-produced failure logged a different correlation ID than its own response body. All 32 now use the correlation allowlist, which is idempotent and keeps a CodeQL sanitizer marker at each site.

**Fixes included:**

- `Program.cs` `MapFallback` built its own trace ID inline and hardcoded a `?? "correlationid"` fallback header name, so a host that left `CorrelationIdHeader` empty still had client values honored on unmatched routes. Now routed through `ExtractTraceIdFrom`; the hardcoded fallback is removed.
- `HealthCheckEndpointModule` used a target-typed `new TraceId(...)`, bypassing normalization.
- A header consisting only of control characters normalized to the empty string and was emitted as `"correlationId": ""` everywhere, because emptiness was tested on the raw value. It now falls back to the server-generated identifier, as `docs/CONFIGURATION.md` already promised.
- Truncation could split a surrogate pair, leaving an orphan that survives the allowlist and reaches a response body as U+FFFD while a log sink receives the raw unit.
- `OAuthManager` passed the `TraceId` *record struct* into a log template instead of `.Value`, so its synthesized `ToString()` emitted `TraceId { Value = … }` while the `/oauth/token` 502 body carried the bare value.

## Behavior changes a reviewer should look at deliberately

1. **`/health/document-cache` 401/403 bodies now honor the configured correlation header.** That endpoint previously always used `HttpContext.TraceIdentifier` and ignored the client header. Routing it through `ExtractTraceIdFrom` is what FR-LOG-6's uniformity asks for, but it is a change beyond pure normalization and is called out rather than buried.
2. **A correlation header that normalizes to empty now yields the server-generated identifier** rather than an empty string.
3. **`MapFallback` no longer honors a `correlationid` header when `CorrelationIdHeader` is empty.** Leaving that setting empty now disables client-supplied correlation IDs on *every* path.

## Knowingly left alone

- **The `traceId` field name in the HTTP 500 body stays as-is.** `LoggingMiddleware` emits `traceId` while every other DMS error body emits `correlationId`. FR-LOG-6 tolerates either. The inconsistency is real — it is the narrow symptom of two competing 500 body contracts — and is tracked separately as [DMS-1518](https://edfi.atlassian.net/browse/DMS-1518). Not part of this work.
- **`src/config` (CMS) is untouched**, by design. CMS has no `CorrelationIdHeader` setting and reads no correlation header, so no client-controlled value reaches it; its own PRD's FR-ERROR-2 *requires* `correlationId` to equal `HttpContext.TraceIdentifier`.

## Test plan

`./build-dms.ps1 Build` — succeeds, 0 warnings, 0 errors. `./build-dms.ps1 UnitTest` — 15,984 passing. `dotnet csharpier format src/dms` — no diff.

Net +6 tests. New coverage:

- `CorrelationIdNormalizerTests` — passthrough; upstream punctuation survives; control characters removed; truncation at both the default and a configured bound; the truncate-then-filter order (a single equality that fails if swapped); idempotence; empty/whitespace/null-ish; non-positive max length; the surrogate-pair cases, including a pair ending exactly at the boundary.
- `ExtractTraceIdFromTests` — the ingestion choke point had **no test at all** before this change. Header selection, the disable path, the `TraceIdentifier` fallback, header present but empty, and header holding only control characters.
- `CorrelationIdParityTests` — in-process `WebApplicationFactory` fixture asserting log/response parity for a hostile, over-length ID across **404, 401, 403 and 429 from three producing layers**, compared against the expected normalized value rather than merely "not the raw input", plus a clean-ID control arm proving a malformed ID does not change the HTTP status.
- `RequestResponseLoggingMiddlewareTests` — a Core-layer parity test. Its absence is why the 32-site divergence went unnoticed.
- `OAuthManagerTests` — asserts the logged `TraceId` equals the correlation ID *exactly*, because the defect being guarded is that the wrong value *contains* the right one.
- `LoggingSanitizerTests` — the new allowlist, including a differential assertion against `SanitizeForLogging` that fails if the two allowlists ever converge, and U+2028/U+2029 removal.

Two pre-existing test expectations were updated, literals only, with inputs, structure and assertion counts unchanged: `LoggingMiddlewareTests` and `RequestResponseLoggingMiddlewareTests` asserted that `{` and `}` were stripped from a correlation ID, which is the strict `Method`/`Path` allowlist that FR-LOG-3 explicitly forbids conflating with this one. Both still assert that every control character is removed and that body, log entry and log scope agree.

`Features/General/CorrelationId.feature` `@API-061` is untouched and still passes — its `test-correlationId` value is allowlist-clean, so it round-trips byte-for-byte.

### Verification caveats, stated rather than glossed

The 6 failing `Core.Tests.Unit` tests visible locally are **pre-existing sandbox artifacts**, failing identically at the branch point: `EdFi.DataStandard*.ApiSchema` is only available on an Ed-Fi feed unreachable from the build sandbox, so content-only stand-ins were used. Consequence: the ApiSchema-dependent portion of the Core suite was not exercised on this branch. Nothing here touches ApiSchema, but CI should confirm. E2E and integration suites were not run (they need Docker and PostgreSQL), which is why the parity test is an in-process pipeline fixture rather than an E2E scenario.

## Recommended follow-ups (not in this PR)

- `OAuthManager` logs the upstream identity service's raw response body (`{Content}`) through no sanitizer at Warning level — a log-forging vector from an external source, one argument away from a fix in this PR. Not a correlation ID, so out of scope, but it deserves its own ticket.
- `TraceId` is a `record struct` with no `ToString()` override, so `logger.LogX("{TraceId}", traceId)` compiles silently and renders `TraceId { Value = … }` — the defect fixed here in `OAuthManager`. A guard would mean either changing a published contract assembly or adding a source-scanning test, both disproportionate here.
- `src/config/.../RequestLoggingMiddleware.cs:25` passes a trace value through the strict allowlist — the same defect class fixed in DMS by this PR, still present in CMS. Out of scope under this PR's architecture decisions, but worth a CMS-side look.

## Process notes

This change was produced by an autonomous coding run with AI assistance, driven through five rounds of implement → dual independent review → adjudicate. The per-round records are committed under `tasks/` (`review-round-1.md` … `review-round-5.md`, plus `declined-findings.md` with 17 adjudicated entries), including the two mid-loop defects the loop caught: the 32-site FR-LOG-6 divergence and the `TraceId`-struct log site. The loop exited on convergence — all four FRs met with both reviewers returning only Nit-level observations.

**Those `tasks/` files are ephemeral and should be removed with `git rm` before merge.** Two Nit-level items were accepted rather than fixed and are recorded in `review-round-5.md` if a human wants them in the same pass. All AI-generated analysis should be reviewed before it informs a decision; a human remains accountable for the merge.

---

TODO: Public documentation needs to cover the use of the correlation ID, including
the maximum length and the truncation of longer values. That work belongs in the
Ed-Fi documentation repository, not this one.

---
_Generated by [Claude Code](https://claude.ai/code/session_01JMqZKUUxbTKf1dundYGwQv)_

[DMS-1518]: https://edfi.atlassian.net/browse/DMS-1518?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


stephenfuqua:
# Team PR Review — DMS-1457: Normalize the correlation ID once, and apply it everywhere

> Spec used in place of a Jira ticket: `docs/PRD-v8.1.md` §3.11 (FR-LOG-3…6) + `tasks/plan.md` · `tasks/todo.md` · `tasks/declined-findings.md` (D-1…D-17 excluded from all findings)
> Reviewed by five specialists in parallel: Security · Functionality · Maintainability · Usability · Test Coverage
>
> ⚠️ **No build or test run was possible in the review environment** (no .NET SDK; `api.nuget.org` returns 403 by network policy). The PR's "0 warnings, 15,984 passing, csharpier clean" is unverified here — CI must be the gate.

---

## 🔴 Critical
None found.

## 🟠 High
None found.

## 🟡 Medium

- **[Test Coverage · Maintainability]** Task 5's guard test does not exist as a guard, and D-16 declined the stronger option on a factually incorrect premise.
  AD-2 accepts that nothing structurally prevents a future raw `new TraceId(...)` and explicitly delegates the mitigation to "Task 5's guard test." What shipped (`CorrelationIdParityTests.cs:245-266`) asserts four *currently existing* endpoints — it enumerates and scans nothing, so a new endpoint doing `TraceId t = new(context.TraceIdentifier)` adds no failing test. That is exactly the defect class this PR just fixed twice, and `HealthCheckEndpointModule` was invisible to `grep "new TraceId("` because it was target-typed. D-16's rationale asserts "a source-scanning architecture test is machinery this repository has no precedent for" — but `Backend.Cdc.Tests.Unit/CdcConnectorTemplateIntegrationBoundaryTests.cs:135,215-232` walks to the repo root, loads a `.csproj` and asserts an allowed-reference list. The precedent exists.
  *Fix:* add a ~30-line `CorrelationIdConstructionSiteGuardTests` using that same helper pattern, asserting the production construction-site set is exactly `{AspNetCoreFrontend.cs, Core/Model/No.cs}`.

- **[Test Coverage]** 30 of the 31 swapped Core/Backend log sites have no assertion, and the parity fixture cannot reach any of them.
  Only `RequestResponseLoggingMiddleware.cs:36` is covered. Existing fixtures for the other 30 (`ResolveMappingSetMiddlewareTests`, `ProfileWritePipelineMiddlewareTests`, …) all use allowlist-clean values like `new TraceId("123")`, so reverting any of those lines to the strict allowlist would be invisible to the suite. All four parity requests are answered before reaching the Core pipeline, so no Core-produced status (400/405/409/412/415/503) has an end-to-end parity assertion. `tasks/review-round-3.md:163` flagged this and it was not remedied.
  *Fix:* the guard test above can carry a regex assertion covering all 31 sites in the same fixture.

- **[Maintainability · Functionality]** Three coexisting conventions for logging a `TraceId`, with the deciding rule written down in one file only.
  Verified: `DescriptorReadHandler.cs` wraps the sanitizer at line 1063 but passes `request.TraceId.Value` raw into the identical `{TraceId}` template at 145, 505 and 1735. Repo-wide it is ~38 wrapped vs ~162 raw. Both are behaviourally identical (AD-1 makes raw correct), so the wrong choice is invisible — and the stated rationale, "keeps a CodeQL sanitizer marker at each site," holds at 38 of ~200. The reason to wrap is stated only in `CustomResourceValidationMiddleware.cs:28-36`, which itself concedes the distinction is "historical rather than semantic."
  *Fix:* state the rule once in `SanitizeCorrelationIdForLogging`'s `<remarks>`, then pick one direction. Also reword the PR's "Applied at every use site" claim.

- **[Maintainability · Test Coverage]** The parity fixture — the PR's central FR-LOG-6 evidence — has two flake vectors and a wall-clock dependency.
  (a) `RateLimitedArmLoggedEventCount = 2` / `MainFactoryArmLoggedEventCount = 3` / `HostileArmLoggedEventCount = 5` (`:75-89`, asserted `:281,:293`) encode *application-wide* `TraceId`-bearing log volume, because the recorder returns `IsEnabled => true` for every level and records any event with a `TraceId` property. Any future `{TraceId}` log line on those paths fails with `Expected 5 items, but found 6`, pointing at nothing. (b) The permitted `/health` request at `:153-158` is the one request whose body is never read, so its completion event may not be enqueued when the snapshot is taken at `:170` — TestServer returns when the response *starts*, and `LoggingMiddleware` logs after `await _next(context)` returns. (c) The 429 arm and its clean control arm share one 60-second rate-limit window across two factory boots plus five requests.
  *Fix:* filter the recorder to `RequestLoggingEventIds` and assert per-arm presence + `OnlyContain`; `await permitted.Content.ReadAsStringAsync()` after `:158`; give the control arm its own factory.

- **[Maintainability]** The allowlist and truncate-then-filter rationale is duplicated across six production/doc locations with no canonical home — and `tasks/declined-findings.md` D-9 records that this drift *already materialised once* during the review loop.
  `CorrelationIdNormalizer.cs:21,:49`; `LogSanitizer.cs:80,:94-101,:156-161`; `LoggingSanitizer.cs:39-41`; `docs/LOGGING.md:213-222`. `LogSanitizer.cs:156-161` restates `:94-101` almost verbatim two functions apart. `CorrelationIdNormalizer.cs` is 56 comment lines to 25 code lines for one 20-line method — the proportionality question plan §1 raises, answered in the other direction. Worse, the normalizer's own doc comment promises U+2028/U+2029 removal that it *delegates*, so a change in `LogSanitizer` silently falsifies it.
  *Fix:* make `SanitizeCorrelationIdForLog` canonical; reduce the rest to `<see cref=…>` pointers.

## 🔵 Low

- **[Security · Usability · Functionality]** *(three reviewers, independently)* A whitespace-only header defeats a guarantee the new code and new docs both state. `AspNetCoreFrontend.cs:431` tests `normalized.Length == 0`; space is not a control character, so `Normalize("   ")` returns `"   "` — pinned as intended by `CorrelationIdNormalizerTests.cs:252`. But `AspNetCoreFrontend.cs:405-412` and `docs/LOGGING.md:249-252` both promise the server identifier is substituted "so a client cannot blank the operational identifier." Kestrel strips ASCII OWS, but `U+00A0` survives Latin-1 header decoding. *Fix:* `string.IsNullOrWhiteSpace(normalized)` — one word, narrows no allowlist, so it does not touch D-7.
- **[Security]** Unicode **Cf** characters survive normalization: U+202A-U+202E (incl. RLO), U+2066-U+2069, U+200B-U+200D, U+00AD, U+FEFF. `char.IsControl` is true only for Cc. A bidi override renders the rest of a log line right-to-left in a viewer, and zero-width characters make two visually identical IDs distinct strings — defeating "search the logs for this ID." This is *not* D-7: it is the same reasoning the PR already applied to U+2028/U+2029, and it makes `LogSanitizer.cs:160`'s claim that "the effective allowlist is the printable range minus those two" inaccurate. *Fix:* `&& CharUnicodeInfo.GetUnicodeCategory(c) != UnicodeCategory.Format`, or correct the two claims.
- **[Usability · Functionality]** `docs/LOGGING.md:240-245` claims normalization reaches "every error response body, whatever the status code." Verified false: the 413 path (`LoggingMiddleware.cs:112-134`) sets the status and logs a `TraceId` but writes **no body**; `ManagementEndpointModule.cs:99,128,173,202`, `MetadataEndpointModule.cs:374,403,463,484`, `XsdMetadataEndpointModule.cs:80,87` return 404s with no `correlationId`. Not an FR-LOG-6 violation (those bodies use no correlation ID), and all pre-existing — but the sentence will be believed. *Fix:* narrow to "every error response body that carries a correlation ID."
- **[Functionality]** FR-LOG-4's "host-configurable override" has no shipped deployment path. Verified: `CorrelationIdMaxLength` appears in **no** compose file, `.env` template, or `appsettings.json`, while `CORRELATION_ID_HEADER` is plumbed through 10+ of them. Met at the code level; unreachable through supported artifacts. Pre-existing (#1206), but FR-LOG-4 is one of the four requirements this PR closes. *Fix:* add `AppSettings__CorrelationIdMaxLength: ${CORRELATION_ID_MAX_LENGTH:-255}` to `local-dms.yml`/`published-dms.yml` and the `.env` templates.
- **[Security · Usability]** `CorrelationIdMaxLength` has no upper bound (`AppSettings.cs:80` validates only `> 0`). A host setting `100000` puts up to Kestrel's ~32 KB of client-controlled string into every log event and error body. The adjacent `DefaultPartitionCount` row documents "between 1 and 200," so the repo's own convention argues for a ceiling. Host-config only, so robustness not vulnerability.
- **[Functionality · Usability · Maintainability]** `docs/CONFIGURATION.md:22` — "the service refuses to start otherwise" is inaccurate: `Program.cs:255` still calls `RunAsync()`, and `ReportInvalidConfigurationMiddleware` answers **every** request (including `/health`) with a bodiless 500. Violates Task 7's accuracy criterion. It also contradicts `CorrelationIdNormalizer.cs:43-46`'s rationale, giving a maintainer two answers to the same question.
- **[Maintainability]** The non-positive `maxLength` clamp (`CorrelationIdNormalizer.cs:66`) is documented as protecting an unreachable scenario, but is silently load-bearing: delete it and `maxLength == 0` makes `retained == 0`, so `char.IsHighSurrogate(value[retained - 1])` indexes `value[-1]` and throws. A future simplifier who trusts the comment introduces a crash. *Fix:* hoist the bound into the probe — `if (retained > 0 && char.IsHighSurrogate(...))` — so the guard is local to the hazard.
- **[Test Coverage]** "Net +6 tests" is wrong by an order of magnitude: the diff adds 58 `[Test]` and 12 `[TestCase]` attributes and removes zero — ~68 new cases. "+6" looks like a single-round delta from `review-round-5.md:189` presented as the whole-PR figure. Matters as a verification signal: if CI reports ~+6, ~60 tests are not being discovered.
- **[Test Coverage]** FR-LOG-5's "still *succeeds*" half is unasserted — no test shows a request that would return 2xx still returns 2xx with a hostile ID. The one 2xx carrying the hostile ID (permitted `/health`, `:153-158`) has its status asserted nowhere; it could return 429 and the fixture would still pass.
- **[Test Coverage]** Three documented behaviours untested: a lone surrogate already in the *input* (which `CorrelationIdNormalizer.cs:52-58` explicitly promises is preserved, not repaired), a single astral character at `maxLength: 1` → empty, and the exact `value.Length == effectiveMaxLength` no-truncation boundary. *(The two surrogate cases the PR does claim are genuinely correct — the arithmetic was checked, and widening `IsHighSurrogate` to `IsSurrogate` would fail the boundary test.)*
- **[Maintainability · Usability]** `AspNetCoreFrontend.cs:426` adopts `StringValues.ToString()` 35 lines below a comment in the same file warning against exactly that ("ToString() drops empty entries"). A repeated correlation header now comma-joins into one ID — deterministic and harmless, but undocumented, with two contradictory precedents in one file.
- **[Usability]** The catch-all 404 is served as `application/json`, not `problem+json`. Verified: `Program.cs:216` sets the content type, then `:228`'s `WriteAsJsonAsync` overwrites it — the explicit line is dead. It is the one DMS error response whose media type differs from every other. Pre-existing, but this is the block the PR edits and the fix is one argument.
- **[Maintainability · Security]** Merging `tasks/` depends on a manual `git rm` that nothing enforces — no `.gitignore` entry, no CI check. 2,252 lines across 8 files. Verified upside: no shipped code comment references a `tasks/` path, so nothing dangles. But honouring the deletion also removes the only durable record of decisions like D-11 (the `/health/document-cache` change is intended). *Fix:* delete in this PR, first folding the two or three durable decisions into `docs/LOGGING.md`; add `tasks/` to `.gitignore`.
- **[Maintainability]** `CorrelationIdRecordingLoggerProvider` is a second, lossier test-logging abstraction in a project that already has `TestLogger<T>` (`LoggingMiddlewareTests.cs:32`) capturing level, event id, exception and scopes. The pre-filtering is also what makes the exact counts above opaque.
- **[Security]** Two pre-existing sinks in files this PR touches, both worth folding into the follow-up ticket the PR already recommends: `ValidateResourceKeySeedMiddleware.cs:161` and `ValidateStartupInstancesTask.cs:217` pass `SanitizeForConsole` output into `ILogger`, and that function deliberately preserves `\r`/`\n`; `OAuthManager.cs:76` echoes the upstream IdP's raw response body into the 502 `detail`, so an IdP misconfiguration can leak internal detail to an unauthenticated caller. Neither is in the ledger (D-14 covers only the *log* side of the latter).

## ⚪ Nit

`HealthCheckEndpointModule.cs:14` inverts the `AppSettings` alias convention used at `WebApplicationBuilderExtensions.cs:27` · two `LoggingMiddlewareTests` names still attribute truncation to a class that no longer implements it · `docs/LOGGING.md`'s "normalized once" overstates it (the function is applied identically at six sites; the guarantee rests on idempotence, not single evaluation) · several new fixtures do arrange+act inside `[Test]` rather than `Setup`, against `AGENTS.md:121` · the 4-assertion `It_` methods at `CorrelationIdParityTests.cs:220-242` lack an `AssertionScope`, so the first failure masks the rest · `LoggingSanitizerTests`' differential assertion uses `NotBe` where an exact expected value would also catch a partial widening · `CONFIGURATION.md:21` is a ~130-word cell with two nested em-dash asides, and `:22` pulls UTF-16 surrogate mechanics into a config reference table · `CONFIGURATION.md:22` says "characters" where the cut is on a UTF-16 code unit · `MapFallback` service-locates a singleton `IOptions<AppSettings>` per unmatched request.

---

## Specialist Verdicts

| Reviewer | Verdict |
|----------|---------|
| Security | Net improvement — closes two genuinely uncapped client-controlled paths and a hardcoded header-name bypass; nothing above Low. |
| Functionality | **COMPLETE** — all four FRs implemented as specified with named tests; every finding Low or Nit, three documentation-only. |
| Maintainability | Well-shaped change (one type, one ingestion point, no layering violations); risk is in the surroundings, not the abstraction. |
| Usability | The core promise holds across every failure path reachable except the bodiless 413; docs unusually honest about rough edges. |
| Test Coverage | **ADEQUATE** — all ten plan §6 edge cases covered by correct assertions, but the two structural risks are protected by nothing. |

---

## Summary

**No merge blockers.** All five reviewers independently confirmed the central claims: exactly one reachable production `TraceId` construction site remains, the allowlist is a bare `!char.IsControl(c)`, `Method`/`Path` still use the strict allowlist unchanged, `src/config` has zero changes, `grep -ci whitelist` is 0, and — checked line by line by two reviewers — **no test was weakened, deleted or skipped**; the "literals only" claim is accurate, and the order-pinning, idempotence and `OAuthManagerTests` exact-equality assertions all genuinely fail if the behaviour regresses.

Three cheap things worth doing before merge: the `IsNullOrWhiteSpace` one-word fix (found by three reviewers, and it makes a new doc claim true), the `await permitted.Content.ReadAsStringAsync()` line that closes a real flake in the PR's own key fixture, and the `git rm tasks/` the author already asked for.

The most substantive finding is the guard test: **D-16 declined a source-scanning architecture test on the stated premise that the repo has no such precedent, and that premise is wrong** — `CdcConnectorTemplateIntegrationBoundaryTests` is exactly that pattern. Since AD-2 deliberately leaves `TraceId` unvalidated and names the guard test as the *only* mitigation, that gap is worth reopening with the product owner rather than carrying forward.

Two caveats on this report itself. Nothing here was compiled or executed, so CI remains the gate on the build and test claims — including a clean restore of `EdFi.DataStandard*.ApiSchema` for the frontend project, which is where the new parity fixture boots the whole app. And this review is AI-generated analysis produced by five parallel agents against the `tasks/` and PRD documents; it should be read and verified before it informs the merge decision, and a human remains accountable for that call.

stephenfuqua:
@claude 

1.  LogSanitizer.cs  now has two near-duplicate two-pass sanitization loops ( SanitizeForLog  and the new  SanitizeCorrelationIdForLog ). Consider refactoring to a single shared helper parameterized by the allow-list predicate (similar to how #1232 structured  LogSanitizer.Sanitize(...) ), to avoid drift between the two implementations over time.
2. Consider whether a full  CorrelationIdNormalizer  class is necessary versus extending the existing  LoggingSanitizer / LogSanitizer  utilities — worth a quick architectural gut-check since this introduces a second normalization entry point.
3. Naming consistency:  SanitizeCorrelationIdForLogging  vs. other sanitizer method names in the codebase — please align naming conventions across  LogSanitizer / LoggingSanitizer .
4. Import https://raw.githubusercontent.com/Ed-Fi-Alliance-OSS/Data-Management-Service/3947aeffb68d9026fe8e2fd5b7de49229eb98272/src/dms/tests/EdFi.DataManagementService.Tests.Integration/Tests/Given_CorrelationIdNormalization_Parity.cs which has a useful test; this comes from a competing PR that I'm going to close.

PR #1242, opened 2026-09-10
[DMS-1343] Resolve the apiClient GET key-first on one route

samuellugo-ship-it:
## What

Adds numeric-id lookup to the single-item API-client GET while preserving the existing OAuth client-key lookup, using one route that resolves client keys first and falls back to the numeric primary key only when no key matches.

## Why

The published Management API 3.0.0 specification models `GET /v3/apiClients/{id}` with `id` as `integer/int32`, and `apiClientModel` carries `id` and `clientId` as separate fields. CMS previously bound the route segment as a string and looked up only by OAuth `clientId`, so a spec-conformant caller could not read an API client by numeric id.

The OAuth client-key lookup must keep working because it is used by the DMS application-context lookup, CMS E2E flows, and the operator guide.

## Implementation Decisions

DMS-1343's Jira description is the source of truth for these decisions. This PR implements the same contract.

### Option chosen

Support both identifier styles on the existing single-item GET route. CMS serves both the OAuth client key and the numeric ApiClient id through `GET /v3/apiClients/{id}`. No `byId` route, query parameter, or relocated client-key route is introduced.

### Runtime route and authorization

CMS registers one GET route: `MapLimitedAccess("/v3/apiClients/{id}", GetByIdentifier)`. The limited-access read policy is unchanged and matches the collection and existing service-account read behavior.

### Identifier resolution contract

| Client-key lookup | Segment parses as int32 | Numeric lookup | Response |
|---|---|---|---|
| Success | not attempted | not called | 200, key-matched row |
| FailureUnknown | not attempted | not called | 500, sanitized |
| FailureNotFound | no | not called | 404, `ApiClient not found` |
| FailureNotFound | yes | Success | 200, id-matched row |
| FailureNotFound | yes | FailureNotFound | 404, `ApiClient with ID {parsedId} not found.` |
| FailureNotFound | yes | FailureUnknown | 500, sanitized |

Parsing uses `NumberStyles.Integer` with `CultureInfo.InvariantCulture`. The id-bearing 404 interpolates the parsed integer rather than the raw path segment.

### OpenAPI and metadata

The served OpenAPI document contains one single-item GET path item for apiClients: `/v3/apiClients/{id}`. The GET parameter is documented as string, because the runtime accepts the raw path segment and resolves it key-first with numeric fallback. `PUT`, `DELETE`, `reset-credential`, and ownership routes continue to use the numeric `int32` id and are unchanged.

### Scope exclusions

- No DMS change and no file under `src/dms/` is touched.
- No repository, DDL, response DTO, identity provider, ownership module, or application module change.
- No change to `PUT`, `DELETE`, `reset-credential`, or ownership behavior.

## Changes

- Replaced the API-client single-item GET handling with one key-first route and handler in `ApiClientModule.cs`.
- Added unit coverage for the six runtime outcomes, including the collision case where a path segment is both a stored `ClientId` and another row's numeric primary key.
- Updated OpenAPI contract tests and metadata expectations for one GET path item with a string parameter.
- Added CMS E2E coverage for numeric-id lookup, key lookup, numeric 404, non-numeric 404, and tenant scoping.
- Updated the operator guide and added the DMS-1343 design record to the configuration-service design index.

## Verification

| Gate | Result |
|---|---|
| `dotnet csharpier check src/config` | 435 files checked, clean |
| `build-config.ps1 Build -Configuration Release` | Build succeeded, 0 warnings, 0 errors |
| `build-config.ps1 UnitTest -Configuration Release` | 1780 passed, 0 failed, 0 skipped |
| Frontend unit project filtered to the affected test classes | 225 passed, 0 failed, 0 skipped |

The Release unit lane predates the key-first patch. The filtered unit run was executed after the key-first patch and covers the affected fixtures. CI is the full current-head gate.

E2E, with teardown before setup and again after each run:

| Feature | Environment | Passed | Failed | Skipped |
|---|---|---|---|---|
| ApiClients | PostgreSQL, self-contained identity, single tenant | 25 | 0 | 1 |
| Tenants | PostgreSQL with multi-tenancy enabled | 4 | 0 | 0 |

The E2E runs predate the key-first handler patch. After that patch, the feature-file edits were comments only; no executable E2E steps changed. The current-head CI run provides the authoritative E2E gate.

## Known Residual Behavior

A token whose `client_id` is numeric, matches no stored `ClientId`, and collides with an ApiClient primary key can still produce a CMS 200 for the numeric row; DMS then rejects the mismatch as `Unavailable`. CMS-issued keys are GUID strings, so this requires external provisioning.

## Follow-Up

Database-level uniqueness for `dmscs.ApiClient.ClientId` and `ClientUuid` is tracked separately as DMS-1529 under DMS-1072. A uniqueness constraint does not prevent a client key from looking numeric; a future format constraint on `ClientId` would remove that remaining ambiguity.

### DIFF
--- a/docs/API-CLIENT-AND-INSTANCE-CONFIGURATION.md
+++ b/docs/API-CLIENT-AND-INSTANCE-CONFIGURATION.md
@@ -104,10 +104,13 @@ today. Two different identifiers are involved:
 - the **`clientId`** (returned as `key` when credentials are issued) is the
   OAuth client identifier.
 
-The single-item `GET` accepts either one, and the path segment itself decides
-which lookup runs: a segment that is a valid 32-bit integer is read as the
-numeric `id`, and any other segment is read as the OAuth `clientId` key. The
-two values remain distinct, because a client's `id` and its `key` are different
+The single-item `GET` accepts either one and resolves them key-first. The
+segment is matched against the OAuth `clientId` key, and that match wins
+whenever it exists. Only when no client key matches, and the segment is a valid
+32-bit integer, is it read as the numeric `id`. Key-first is deliberate: a
+`clientId` has no enforced format, so a key made only of digits is possible, and
+reading the numeric `id` first could hand back an unrelated client. The two
+values remain distinct, because a client's `id` and its `key` are different
 values resolved through different columns. You just do not have to convert one
 into the other to read a client.
 
--- a/docs/API-CLIENT-AND-INSTANCE-CONFIGURATION.md
+++ b/docs/API-CLIENT-AND-INSTANCE-CONFIGURATION.md
@@ -97,12 +97,19 @@ an absent one, depending on the deployment's identity provider.
 #### Provisioning a Client With No Data Store Assignment
 
 The sequence below is the whole lifecycle, using only endpoints that exist
-today. Two different identifiers are involved and they are not interchangeable:
+today. Two different identifiers are involved:
 
 - the **numeric `id`** returned in the API-client response body is the
   management identifier used in `PUT`, `DELETE` and `reset-credential` routes;
 - the **`clientId`** (returned as `key` when credentials are issued) is the
-  OAuth client identifier, and it is what `GET /v3/apiClients/{clientId}` takes.
+  OAuth client identifier.
+
+The single-item `GET` accepts either one, and the path segment itself decides
+which lookup runs: a segment that is a valid 32-bit integer is read as the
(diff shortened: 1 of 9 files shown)


samuellugo-ship-it:
Design replaced after review: the `{id:int}` two-route approach is superseded by a single `/v3/apiClients/{id}` route with key-first resolution. PR body rewritten accordingly.

simpat-adam (approved):

PR #1257, opened 2026-09-16
[DMS-1369] Verify the served snapshot OpenAPI contract and its effective-schema-hash neutrality

samuellugo-ship-it:
## What

Verifies that the snapshot OpenAPI contract published in the pinned ApiSchema packages (version 1.0.335) is served correctly by DMS on every supported surface and intake path, is preserved through profile filtering, and leaves the effective schema hash and goldens untouched.

## Why

DMS-1371 added the snapshot contract upstream: a reusable `Use-Snapshot` header parameter and the `SnapshotNotFound` and `SnapshotMethodNotAllowed` response components, referenced from the operations that serve them. DMS serves the packaged OpenAPI documents as supplied, so the contract reaches clients only through the package pin already on `main`. DMS-1369 requires proof that the served resources, descriptors, Change Queries, and profile documents advertise the contract for every supported package family under both the bundled and file-based intake paths, that every reference resolves within its own document, and that the bump is OpenAPI-only with no effective schema hash movement.

One production gap surfaced along the way. Profile filtering pruned base schemas that a retained component response still pointed at, which left the snapshot `404` and `405` responses unresolvable in a profile document. That is fixed here.

## Changes

Production

- `ProfileOpenApiSpecificationFilter`: keep schemas reachable from retained `components.responses` entries when replacing base schemas with profile-suffixed versions, and seed the final orphan sweep from those responses as well. Request bodies and parameters are deliberately not seeded, so write-side and off-profile content is still dropped.

Package payload verification

- `SnapshotOpenApiPackagePayloadTests` (ApiSchemaDownloader integration): downloads all seven published ApiSchema packages at the pinned version and asserts each carries the contract in its OpenAPI payloads, with a pre-snapshot version used to prove the checks detect its absence.

Served document verification

- `OpenApiSnapshotContractAssertions` and its tests: shared primitives that resolve a served document against itself and assert snapshot operation coverage, so every fixture makes the same claim the same way.
- `PackagedSnapshotOpenApiContractTests`: asserts the contract on the documents DMS assembles from the packaged Data Standard 5.2, 5.2 with TPDM, and 6.1 schemas, plus a profile document derived from the real 5.2 package.
- `SnapshotOpenApiRuntimeParityTests`: pins the served snapshot response metadata to the runtime responses DMS actually returns.
- `SnapshotOpenApiMetadataTests` (frontend): asserts the contract over the metadata endpoints as served.
- `ProfileOpenApiSpecificationFilterSnapshotTests`: pins the contract through profile filtering, including the retained-response schema preservation above.

Hash neutrality

- `SnapshotOpenApiHashNeutralityTests`: for the full Data Standard 5.2 set (core, TPDM, Sample, Homograph) and the full 6.1 set (core, Sample, Homograph), removes every snapshot artifact from the OpenAPI payload regions of every package and proves the normalized schema and effective schema hash are byte-for-byte unchanged. Removal is asserted per package against its resource count so no package can be silently skipped.
- `PackagedApiSchemaContract`: adds an extension package loader that derives the restored root beside the core package's recorded root.
- Core unit test csproj and lock file: reference the 5.2 and 6.1 Sample and Homograph packages with all assets excluded and no generated path property, since the build targets treat a 5.2 extension path property as a request to bundle it into the staged workspace other fixtures load. The lock file gains four direct entries at the already-pinned version; no version changes.

End to end

- `SnapshotOpenApiSurface.feature` and `SnapshotOpenApiStepDefinitions`: eight scenarios covering the resources, descriptors, and Change Queries documents plus extension resources on the Data Standard 5.2 lane (shard 4, one MSSQL representative) and the 6.1 lane, including tracked-change and mutation coverage for Sample and Homograph.
- `ProfileOpenApiSpecification.feature`: one scenario proving a served profile document preserves the contract (shard 3).
- `CursorPagingStepDefinitions`: resolve the API response through the scenario context so its served-paths helper works from features whose request step lives in another binding class.

Documentation

- `docs/API-SCHEMA-DOCUMENTATION.md`: documents the snapshot contract the base documents carry since ApiSchema 1.0.334 and how profile documents inherit it.


samuellugo-ship-it:
Pushed 0e166afba: the hash-neutrality fixtures now read the extension ApiSchema documents from the staged build output instead of the runner's NuGet cache, which the unit-test job never restores.

samuellugo-ship-it:
Review findings addressed in 168f92489. One choice that is not obvious from the diff: the new package-bump hash proof compares 1.0.333 -> 1.0.334 (the snapshot-introducing bump, b0b68972c), not the current 1.0.335 pin — d45b508d2 (DMS-1409) moved the effective-schema hash and its goldens for an unrelated reason, so a comparison against today's pin would fail for something other than the snapshot contract.

simpat-adam (approved):

samuellugo-ship-it:
Rebased onto main and recorded the `EdFi.Api.Identity` transitive project reference in the `ApiSchemaDownloader.Tests.Integration` lock file — DMS-1514 (#1268) added that project under Core but only regenerated lock files for projects that referenced Core at the time, so locked-mode restore failed NU1004 only once both changes were present in the merge queue.

simpat-adam (approved):

PR #1258, opened 2026-09-16
[DMS-1547] Restore NuGet packages in the CI unit test job before it runs against the build output

samuellugo-ship-it:
## What
Adds a locked-mode `dotnet restore` of the DMS solution to the `Run Unit Tests` job so the runner's NuGet packages folder holds every package the lock files pin, instead of only what the NuGet cache happened to contain.

Tracked as DMS-1547, related to DMS-1369.

## Why
The job runs `dotnet test --no-build --no-restore` against the downloaded build output, so its `~/.nuget/packages` comes solely from `actions/cache`. That cache is keyed on a hash of `Directory.Packages.props` alone, and `actions/cache` never rewrites an existing key. A PR that adds a `PackageReference` at an already pinned version, or changes a lock file, therefore runs unit tests against a packages folder that predates the change, on the PR and on main after merge, until `Directory.Packages.props` next changes.

Several Core unit tests read the packaged `ApiSchema.json` of the pinned ApiSchema packages straight from that folder at test time, through the package roots the build records as assembly metadata. They pass today only because main already referenced those packages when the cache was last saved. DMS-1369 (#1257) hit the failure: its new fixtures referenced the Sample and Homograph packages, the build job restored them, and the unit test job failed in `OneTimeSetUp` with the packages named as missing. Rerunning could not fix it. #1257 worked around it by staging the documents into the test output; this PR fixes the job itself so that workaround, and the latent fragility of the existing reads, is no longer needed.

## Changes
- `.github/workflows/on-dms-pullrequest.yml`: new `Restore with locked mode` step in `run-unit-tests`, placed after `Setup dotnet` and before the build output is downloaded and extracted, so the build job's own restore markers laid down by the tar are what `--no-build` evaluates. The step compiles nothing; with a warm cache it costs about a minute. Locked mode fails by name if a pinned package cannot be fetched.
- Comment on the step records the cache-key behavior that makes the restore necessary.

No other jobs change. Once this merges, the `StageSnapshotContractExtensionApiSchemas` target added in #1257 can be removed in a follow-up.


samuellugo-ship-it:
Superseded by #1259 after the branch was renamed to DMS-1547; GitHub closed this PR on the rename.

PR #1265, opened 2026-09-17
[DMS-1442] Refuse ApiSchemas declaring query fields that collide with reserved query parameter names

samuellugo-ship-it:
## What

Refuses an ApiSchema that declares a resource query field spelled like one of the eight query parameter names DMS consumes as control parameters, and declares that set once instead of in five independent lists.

## Why

MetaEd rejects a property name colliding with `offset`, `limit`, or `totalcount`. DMS reserves five more — `pageToken`, `pageSize`, `minChangeVersion`, `maxChangeVersion`, and the `/partitions` count key `number` — and MetaEd protects none of them. An extension author could declare a colliding property, MetaEd accepted the model, and the property was then unfilterable on the affected endpoints with no warning at model build, at ApiSchema load, or at request time.

The failure was not even uniform. On a collection GET-many the filter was silently never applied; on `/deletes` and `/keyChanges` the same name was answered as an invalid query field, so one resource got an HTTP 400 for a filter its collection GET accepted. `number` was worse in a different way: reserved on `/partitions` alone, it filtered on the collection GET and was read as the partition count on that resource's sibling, so a client supplying it there got a different result set than it asked for, or a range error for a non-numeric value.

This is a pre-existing class that the cursor and partition work widened, not a partition defect. `minChangeVersion` and `maxChangeVersion` have been reserved since change queries shipped. Per the ticket, the fix is triage option 2, detect at ApiSchema load, plus the closing requirement that the reserved set be defined in one place so the next reserved name cannot be added without the collision question being asked.

Targets Ed-Fi API v8.1+.

## Changes

**One declaration of the reserved set**

- Adds `ReservedQueryParameters`, naming each reserved parameter with the operations it is reserved on, how a request compares it, and what it is consumed as. Spellings are read from the validator that owns each one rather than re-typed, so the name a request is validated against cannot drift from the name a schema is refused for.
- Derives every previous list from it: `ValidateQueryMiddleware`, `ValidatePartitionQueryMiddleware`, `CursorPagingOpenApiAugmenter`, `PartitionRequestValidator`, and the canonicalization in `AspNetCoreFrontend`. Behavior-neutral; the frontend still canonicalizes the count on `/partitions` alone, now by asking the catalog per operation rather than by a boolean threaded through the canonicalizer.

**Refusal at ApiSchema load**

- Adds `ReservedQueryParameterCollisionDetector` and `ReservedQueryParameterCollisionResult`. The result builds the operator-facing description once, naming the schema, project, resource, field, what the name is taken for, and the rename-and-rebuild remedy. Comparison is case-insensitive, so `PageSize` collides exactly as `pageSize` does.
- `ApiSchemaInputNormalizer` returns that result; `EffectiveSchemaBootstrapper` fails DMS startup with it, replacing a refusal that said only "Unknown normalization result".
- `LoadResultErrorHandler` reports it from the SchemaTools `hash` and `ddl` commands.
- The SchemaTools CDC commands carry that one classification out of loading as a typed exception holding the result, and describe it on the writer the host supplies as stderr. Every other load failure keeps its generic refusal, and nothing crosses the typed output boundary: `CdcDeploymentDiagnostic` is untouched, no `CdcDeploymentFailure` value is added, and the exit code and stdout JSON are unchanged. The description is written where the refusal is discovered rather than where the command returns, so it also covers the eight operations whose controllers classify the exception into a typed diagnostic.

**Retired parity difference**

- Deletes the approved ODS difference recording the `number` asymmetry, its scenario test, and its parity case. No loadable schema can reach that collision any more, and the guardrails require every approved difference to be executable. The served `/partitions` contract is unchanged: the count key is still `number`, still consumed as the count, still removed from filter matching on that operation alone.
- Renames the colliding query field in the cursor-partition fixture.

**Docs**

- `CURSOR-PAGING.md` gains the reserved-name table with what each name is consumed as and where, the load-time refusal, and a note for extension authors that this is a breaking change in v8.1 for any extension already using one of the names.
- `DATA-STRICTNESS.md` case-sensitivity row named only the original three; now names all eight, with one sentence for the `/partitions`-only scope of `number`.
- `API-SCHEMA-DOCUMENTATION.md`, the partitioned cursor-paging design doc, and the epic are corrected to match.

## Testing

Unit coverage is new for the catalog, the detector, the result text, normalization, startup, the SchemaTools handler, and the CDC path. Each behavior was confirmed falsifiable by reverting it alone and checking that the intended cases fail.

Two things reviewers should know:

- The SQL Server integration twins were not run locally and will first execute in CI.
- The SchemaTools unit project has 28 pre-existing failures in its CDC wrapper-bridge fixtures, unrelated to this branch. Confirmed by building pristine `main` in the same environment: 881 passed / 28 failed there, 897 passed / 28 failed here.


bradbanister:
Do this in MetaEd, not DMS. It's a terrible user experience. You don't find out until DMS startup time there is something wrong that can only be fixed by someone (who?) changing MetaEd project files.

PR #1269, opened 2026-09-18
[DMS-1356] Make vendor namespace-prefix updates consistent and lock-participating

samuellugo-ship-it:
## What

Makes `PUT /v3/vendors/{id}` resolve, lock, and persist correctly, so a namespace-prefix update can no longer return `204 No Content` while stranding the identity-provider client identifier of every ApiClient the vendor owns.

## Why

Under Keycloak, `UpdateClientNamespaceClaimAsync` deleted and recreated each client and returned the replacement's identifier, which `VendorModule.Update` discarded. Every affected `ApiClient.ClientUuid` was left pointing at a deleted client, so the next ApiClient update, delete or credential reset, and the next Application update or delete, addressed something that no longer existed. This was a successful-request consistency defect; no concurrent request or provider failure was required.

A failure part way through the client loop could also leave the vendor's clients partially migrated with no compensation, and the workflow mutated the same provider clients that the DMS-1218 aggregate locks serialize without acquiring any of them.

DMS-1218 excluded this deliberately, because a lock-only patch could not make the multi-client workflow correct. Details and the full decision record are in `reference/design/configuration-service/DMS-1356-vendor-namespace-update-consistency.md`.

## Changes

Identity provider

- Keycloak's `UpdateClientNamespaceClaimAsync` now updates the stored client in place under its existing identifier, using the mechanism DMS-1218 established for `UpdateClientAsync`. It upserts the namespace claim, preserves every unrelated mapper, collapses duplicate namespace claims, never writes the secret back, and classifies failures by phase. The client's secret, service account, realm role mappings and claim-set scope all survive, which the previous delete-and-recreate lost along with the identifier.
- OpenIddict is unchanged and now has regression cover for its in-place behavior.

Repository

- Adds `IVendorRepository.GetVendorUpdateState`, which reads the vendor and every ApiClient it owns inside one transaction that row-locks the vendor row, returning each client's row id, client id, current identifier and owning application id, ordered by application then row.
- Removes `VendorUpdateResult.Success.AffectedClientUuids` and the unlocked post-commit query behind it, which is the read the original defect grew from.

Workflow

- `VendorModule.Update` resolves every affected client before mutating anything, acquires the aggregate lock of each owning application in ascending id order, and re-reads under those locks so a workflow that committed while this one waited is honored rather than overwritten.
- Each provider call's reported identifier is persisted through the guarded `SyncApiClientUuid` before the next client is touched, and the vendor row is committed last.
- A failure restores the clients already changed, all under the held locks. The current client is treated as an ambiguous outcome rather than a proven no-op, and an unprovable restoration keeps the original classification while a known failed restoration downgrades to a sanitized 500.
- No partial provider or database failure returns success with mixed database/provider state. The tested outcomes are consistent rollback, retriable conflict, recovered 204 when the vendor commit is proven, or sanitized server failure with the inconsistency logged when consistency cannot be proven.
- An unknown or thrown `UpdateVendor` is resolved by the authoritative re-read: a state matching the command recovers the 204, a state matching the snapshot compensates, and anything else concludes nothing and is logged.
- Lock timeout and persistent drift return the same retriable 409 contract the Application and ApiClient workflows use.

Tests

- Unit cover for the provider repositories and for every branch of the workflow, including lock ordering, drift retry, stable and provider-rotated UUIDs, guarded UUID persistence, guarded persistence failure, release on every path, partial provider failure and rollback, ambiguous vendor persistence, inconsistency logging, and each compensation and resolution outcome.
- New `VendorConsistencyOperationTests` and vendor workflow-concurrency fixtures on PostgreSQL and SQL Server, the latter running against the real lock managers alongside the Application and ApiClient workflows.
- A real-Keycloak E2E scenario creates three clients across two applications, updates the vendor's prefixes, then verifies the stored identifiers remain live by running ApiClient update and credential reset on every affected client, ApiClient delete on one affected client, and Application update and delete on both affected applications. It fails against the pre-fix code at the first ApiClient update after the vendor change.
- The same E2E scenario also runs against the self-contained provider for the OpenIddict regression path, and the representative SQL Server E2E lane covers the database variant.

Documentation

- Adds the DMS-1356 design record with the decision log, the as-implemented commit table, the verification evidence, and the limitations found while implementing.
- Corrects the two DMS-1218 statements this work invalidated, in that document's dated-correction style: vendor update is no longer a non-participating writer, and the vendor namespace-claim update no longer delete-and-recreates.

Generated with [Claude Code](https://claude.com/claude-code)


samuellugo-ship-it:
Round 2 review response in `a35a31ce4`. Fixed: the above-cap fan-out refusal is now its own deterministic 409 (same `urn:ed-fi:api:conflict` type, detail no longer says "Retry the request"); the discarded `TryDeleteClientAsync` result now feeds `FinishCompensationAsync` without short-circuiting prior-client rollback. Intentionally unchanged: the zero-client insert race (D-2 accepted exclusion) and the provider-returned UUID replacement guard. Response-body table (§5.4) and the §13 risk row updated to match.

samuellugo-ship-it:
Round 3 pushed as aece7f055. The 25-application fan-out cap from rounds 1–2 is removed: it never addressed the actual hazard (lock connections drawn from the repository pool, exhaustible by concurrent workflows below any cap). Replacement: `IApplicationLockManager.AcquireAllAsync` takes a vendor's whole application set on one DB session (dedup, ascending, same lock keys), and both lock managers now draw sessions from a dedicated bounded pool (`ApplicationLockConnectionPool`, Max Pool Size 20). Inert `currentClientClean` removed with no classification change. Each of the third review's findings is recorded as fixed / rejected with evidence / deferred in design doc §20.

samuellugo-ship-it:
**Fourth review round — both findings fixed (`12e6d92d7`).**

1. **ApiClient parent move took two lock sessions.** The loop over source/target ids held the first session while waiting for a second from the 20-session lock pool this ticket introduced — connection hold-and-wait, broken by the driver's connect timeout into failed requests, not a DB deadlock; ascending order can't prevent it because the contended resource is the pool. Now one `AcquireAllAsync` per attempt.

2. **Each id of a lock set got its own `AcquireTimeout` window.** N locks could spend ~N windows under staggered contention. Now one monotonic budget per acquisition attempt, started after the session opens and shared by every id, on both backends. Connection opening and release stay outside it — it bounds contention, not the request. Worth noting this was a chosen correction, not an unmet AC: §17 had explicitly documented per-lock windows and Jira requires no set-wide timeout.

R4's architecture is unchanged — no cap restored, no pool limit raised, no new configuration. Design record §21 carries both dispositions plus dated corrections to §17 and §19.

The move fixtures previously asserted *flattened* lock ids, so they passed against the defective loop; they now assert one set holding both ids. Reverting each behavior fails its assertions (ApiClient loop: 10 of 12 move assertions; per-lock windows: 3.689 s PG / 3.508 s MSSQL against a 2.8 s bound).

samuellugo-ship-it:
## Review round 5 — `121e91bc8`

One finding, in `VendorModule.ResolveAmbiguousOutcomeAsync`: an ambiguous no-op `PUT` can satisfy `MatchesCommand` **and** `MatchesOriginal` at once, and command-first ordering returned an unproven `204`.

The overlap proves nothing. Both predicates compare business values, while `UpdateVendor` also writes audit fields they never see, so a state still matching the original values is not evidence that the failed write committed — however it came to match. Only a state that moved *off* the original values can only have come from a commit. This supersedes the earlier recommendation to preserve `204` for the overlapping case.

**Resolution table, applied only after an ambiguous repository result:**

| Matches command | Matches original | Outcome |
|---|---|---|
| Yes | No | `204`, no compensation |
| No | Yes | compensation, sanitized `500` |
| Yes | Yes | commit unproven → compensation, sanitized `500` |
| No | No | existing unresolved handling: logged, sanitized `500`, no compensation |

**The change is one conjunct** on the existing recovered-success arm:

```csharp
case VendorUpdateStateResult.Success resolved
    when MatchesCommand(resolved.State) && !MatchesOriginal(resolved.State):
```

The overlap now falls through to the unchanged original-state arm and takes the compensation and sanitized failure already defined there. No new branch, helper or result type. A `VendorUpdateResult.Success` never reaches this resolution, so an **ordinary no-op `PUT` still returns `204`**. Locks stay held through resolution and compensation; sanitized failures, inconsistency logging, the restore-every-client loop and the guarded persistence of returned UUIDs are untouched.

**Tests** — 15 in 4 fixtures on a new `NoOpAmbiguousVendorUpdateTestBase`: the ambiguous no-op on a returned `FailureUnknown` and on a thrown repository call, the same case against the rotating-provider fixture (compensation targets the UUIDs the forward pass left live and persists the returned ones through the existing guards), and an ordinary successful no-op holding `204`.

In a no-op, compensation re-applies prefixes equal to the requested ones, so the existing `RollbackCalls`/`UpdateCalls` prefix filters cannot separate a rollback from a forward call. The new fixtures identify them by position instead and assert on targeted UUIDs, sync-guard arguments and final stored UUIDs — observable responses and calls, not predicate ordering.

**Regression proof:** with the arm reverted to command-first and the tests kept, 7 of the 15 fail — the three ambiguous no-op fixtures on both the status code and the compensation assertions, while the successful-no-op fixture stays green and pins the `204`. Both directions of drift are covered.

`VendorModuleTests` 195/195; the full `Frontend.AspNetCore.Tests.Unit` project 1368/1368; CSharpier clean on the two touched files. Module-only change, so no new integration or E2E runs.

**Design record:** `D-4` carries the table and the reason; the §5.3 P5 matrix row restates it. Deliberately not touched: `ApplicationModule.Update` — same `ResolveAmbiguousOutcomeAsync` shape and therefore the same latent overlap, flagged rather than changed — along with the document revision header and the historical review sections.

simpat-adam (approved):

PR #1272, opened 2026-09-18
[DMS-1478] Authenticate and ownership-check POST /connect/revoke

stephenfuqua:
Implements the planned authentication and ownership work for `POST /connect/revoke`.

## Behavior changes

**1. `POST /connect/revoke` now requires an authenticated caller.** The route gains a bare
`.RequireAuthorization()` — no named policy. This is deliberate: `SecurityConstants.ServicePolicy`
requires a service-role claim that an ordinary client-credentials token minted by `/connect/token`
does not carry, so applying it would lock normal clients out of revoking their own tokens.
`connect/register`, `connect/token`, and `connect/introspect` are untouched. An unauthenticated or
invalid-bearer request now returns `401`.

**2. A caller may only revoke a token it owns.** `RevokeTokenAsync` takes the caller's `client_id`
and compares it to the `client_id` claim on the target token. On mismatch — or if the token has no
`client_id` claim, or fails verification — the method returns `false`, makes no repository call, and
throws nothing.

**Mismatch behavior: `200 OK`, silent no-op.** Per RFC 7009 conventions already used by this
endpoint, a mismatch is indistinguishable from "token not found", so nothing is leaked about whether
a token exists or who owns it. The only non-200 outcomes are the pre-existing `400` for a missing
`token` form field and the new `401` for an unauthenticated caller.

Recorded for integrators in `docs/changelog/8.1.0.md`.

## Security fix folded in

The target token's claims were previously read with an unvalidated
`new JwtSecurityTokenHandler().ReadJwtToken(...)`. Trusting an unverified `client_id` would have let
a caller forge a token asserting their own `client_id` while embedding a victim's real `jti`,
defeating the entire ownership check. That call is replaced with the signature-verifying
`JwtTokenValidator.ValidateToken`, using the same signing keys / issuer / audience as
`ValidateTokenAsync`.

`ValidateTokenAsync` itself is deliberately **not** reused: it gates on DB status `== "valid"`, which
would make re-revoking an already-revoked token a failure rather than the harmless idempotent no-op
RFC 7009 requires.

## client_id claim name across provider modes

The `client_id` claim name was confirmed identical across self-contained and Keycloak-issued tokens
(verified against a real Keycloak JWT), so **the ownership check needs no per-mode branching** to read
the claim.

One clarification worth carrying into review: identical claim names do not mean the check runs in both
modes. In Keycloak mode no `ITokenRevocationManager` is registered — only the Postgres/MSSQL OpenIddict
extensions register one — so the revocation branch is never entered, the request returns a bare `200 OK`,
and nothing is revoked at all. Ownership-checked revocation therefore executes in self-contained mode
only. The one real Keycloak-mode change here is the new `401` for unauthenticated callers, replacing a
`200` no-op. The docs in this PR say exactly this rather than claiming a "uniform" check.

## Also changed

- **Expired tokens are no longer revocable.** `JwtTokenValidator.ValidateToken` sets
  `ValidateLifetime = true`, so a target token past `exp` + skew is now a no-op where the old
  `ReadJwtToken` path revoked it by `jti` regardless. Harmless in practice and arguably more
  RFC-correct, but it is a real behavior change; documented.
- **Token-verification failures are now logged by category.** `ValidateToken` reports
  `Expired` vs `Untrusted`; expiry logs at Debug and signature/issuer/audience at Warning, so a
  forged-token attempt is no longer buried in routine expiry noise at the same severity. A
  not-yet-valid (`nbf`) token counts as `Untrusted`. This applies to the introspection path too,
  since both share the verifier.
- **E2E step definition.** `StepDefinitions.cs` — the "the current token is revoked" step posted with no
  `Authorization` header and would now `401`. Added the header; the step revokes the token it presents,
  so the caller is the owner. This is an existing test whose expected request shape this change altered.
  It could not be executed here (no container registry access).
- Ownership-mismatch logging is at Debug so an authenticated caller cannot drive unbounded
  Warning-level log growth.
- `TestAuthHandler` gains two opt-in headers (`X-Test-ClientId`, `X-Test-OmitRoleClaim`); absent-header
  behavior is unchanged, so no existing test was affected.

## Tests

Backend `680/680`, frontend `1216/1216`, 0 skipped, build clean, `csharpier check` clean. No test was
deleted, skipped, `[Ignore]`d, or weakened — the two renamed fixtures are behavior updates made in
place (`Given_a_revocation_request_with_a_token` → `..._from_an_unauthenticated_caller`, `200` → `401`).

Coverage added: unauthenticated → 401; invalid bearer → 401; owner → 200 + revocation happened;
non-owner → 200 + repository `MustNotHaveHappened`; forged token naming the caller → 200 + not revoked;
malformed token → 200 + not revoked; missing `token` field → 400; case-differing `client_id` → no-op;
wrong issuer and wrong audience (signed with the real registered key) → no-op; expired owned token →
no-op; empty `client_id`; sibling `connect/*` routes still anonymous; log severity by failure category;
plus manager-level equivalents with `MustNotHaveHappened` on every negative case.
`OWASP-AUTH-COVERAGE.md` carries the full enumeration.

One fixture is a deliberate canary: it authenticates with a principal carrying no role claim, so it
fails with `403` if anyone later attaches a named policy to this route. Verified by temporarily
applying `ServicePolicy` — the canary failed while all other revocation tests stayed green. A
companion fixture arms it by driving the same header at a genuinely `ServicePolicy`-gated route and
observing the `403`.

**Not executed here:** CMS E2E (including the scenario whose step definition changed) and the
Postgres/MSSQL integration suites — Docker's daemon runs in this environment but image pulls are
blocked, so the stack cannot start. The E2E project compiles. **Running `OwaspCriticalPaths.feature`
Scenario 18 against self-contained mode remains a pre-merge gate** — it is the only check that
exercises the step-definition change above.

## Follow-up ticket needed: `client_id` casing

One adjacent, pre-existing defect was found while implementing this and deliberately **not** fixed
here: a token is minted with the `client_id` casing the client supplied at `/connect/token` rather
than the stored canonical casing, while client lookup is case-insensitive under SQL Server's default
collation. One registered client can therefore hold two tokens whose `client_id` claims differ only
by case, and the case-sensitive ownership comparison will refuse to revoke one from the other —
silently, with a `200 OK`.

The `Ordinal` comparison is the correct, security-conservative choice and should stay; loosening it
would make the ownership boundary depend on the deployed engine's collation, and a test now pins it.
The real fix is to mint the claim from the stored canonical value, which carries its own
compatibility question for tokens already in circulation. **This warrants its own ticket.** The
consequence is documented in `CS-AUTH.md`, alongside the introspect-after-revoke pattern for
confirming a revocation actually took effect.

## Known gaps for the reviewer

- The *expired* (well-formed, past-`exp`) bearer-token half of the 401 criterion is not covered at the
  HTTP layer. Probing it would require stubbing the OIDC authority or accepting a flaky test.
  Expired-token rejection is framework behavior (`ValidateLifetime`) in both branches.
- `SecurityConstants.ClientIdClaimType` does not reach the `client_id` OAuth *request parameter* read
  from the form body in `GetAccessTokenAsync` — that is fixed by RFC 6749 and is kept a separate
  literal on purpose, so renaming the claim cannot rename the wire parameter. Nor does it reach the
  claim name in Keycloak-issued tokens, which Keycloak controls. Both limits are stated on the
  constant's doc comment.

## AI assistance disclosure

This change — code, tests, and documentation — was implemented and reviewed with AI assistance, then
checked by separate AI review passes. It is input to human review, not a substitute for it. Two points
deserve human eyes in particular: the security-relevant signature-verification path, and the Keycloak
`client_id` claim-name equivalence, which rests on a prior manual verification against a real Keycloak
JWT rather than anything confirmed in this branch.

https://claude.ai/code/session_01HYCAHs3QzgmuaywXJ7c67a


stephenfuqua:
# PR #1272 Review — [DMS-1478] Authenticate and ownership-check POST /connect/revoke
> Requirements source: `tasks/plan.md` / `tasks/todo.md` (no linked GitHub issue)
> Reviewed by: Security · Functionality · Maintainability · Usability · Test Coverage
> Reviewed independently of PR #1271 — no findings from that review were shared with these specialists.

---

## 🔴 Critical
None found.

## 🟠 High

- **[Usability]** No documented way for an integrator to confirm a revoke actually happened, and no recovery path when it silently didn't.
  Every outcome — owned-token success, wrong-`client_id`, forged token, malformed token, even the Keycloak-mode no-op — returns an identical bodyless `200 OK`. That's a legitimate RFC 7009 tradeoff, but neither `CS-AUTH.md` nor `OWASP-AUTH-COVERAGE.md` tells integrators how to *verify* the outcome (e.g., re-check via `/connect/introspect`). This is sharpened by a real, disclosed defect in `docs/parking-lot.md`: a client-id casing mismatch can cause exactly this silent no-op for a legitimate client revoking its own token — someone responding to a credential leak could believe an incident is contained when it isn't.
  Fix: document the introspect-after-revoke confirmation pattern in both doc files.

- **[Usability]** Breaking change (anonymous → `401`) shipped with no changelog, release note, or versioning discussion anywhere in the branch.
  Fix: add a changelog/release-note entry and consider a version bump.

- **[Test Coverage]** No test pins the `Ordinal` (case-sensitive) `client_id` comparison, despite `parking-lot.md` itself calling this "the correct, security-conservative choice" that "must stay."
  A future "fix" switching to `OrdinalIgnoreCase` would pass every existing test while reopening the exact cross-client revocation risk the parking-lot doc warns about.
  Fix: add a fixture with token/caller `client_id` values differing only by case, asserting no-op.

- **[Test Coverage]** No test proves rejection of a correctly-signed token with a mismatched issuer/audience — only "wrong signing key" is exercised.
  This PR is what newly makes issuer/audience verification load-bearing for the ownership boundary; a future loosening of `ValidateIssuer`/`ValidateAudience` would go undetected.
  Fix: add a fixture signing with the real registered key but a different issuer/audience.

- **[Test Coverage]** The canary test's own precondition (`X-Test-OmitRoleClaim`) is never independently verified against a real `ServicePolicy`-gated route — an admitted gap in the PR description itself.
  The canary proves "this caller isn't rejected," not "this caller would be rejected by `ServicePolicy` if it were attached" — its entire inferential weight rests on trusting a comment in `TestAuthHandler`, not an observed contrast.
  Fix: add a companion test hitting a real `ServicePolicy`-gated route with the same header, asserting `403`.

## 🟡 Medium

- **[Security] [Functionality]** `docs/parking-lot.md` ships in this PR's diff and, per its own banner, must be deleted before merging to `main` — that hasn't happened yet.
  Fix: merge gate — delete the file and file the suggested follow-up ticket first.

- **[Maintainability]** `RevokeTokenAsync` duplicates `ValidateTokenAsync`'s signing-key/issuer/audience construction verbatim, in the same class, on a security-critical path.
  Fix: extract a shared `VerifyTokenAsync` helper.

- **[Maintainability] [Functionality]** The new ownership check re-implements `client_id` extraction inline instead of using the existing (currently unused) `ClaimsPrincipal.GetClientId()` extension — a missed consolidation opportunity in a security-sensitive spot.
  Fix: use `httpContext.User.GetClientId() ?? string.Empty`.

- **[Usability]** `CS-AUTH.md`'s summary table lists `/connect/revoke` as simply "Required" auth with no footnote noting Keycloak-mode revocation is functionally a no-op regardless of authentication — a skim-reader would miss this.
  Fix: add a one-line footnote on the table row itself.

- **[Usability]** `parking-lot.md`'s disclosed casing defect isn't cross-referenced to its actual usability consequence (silent failure creating false confidence during incident response) — a reviewer skimming it in isolation might under-prioritize the follow-up ticket.
  Fix: add one sentence connecting the defect to the incident-response risk.

- **[Test Coverage]** No positive test proves `connect/register`/`connect/token`/`connect/introspect` remain anonymous under the same `AddTestAuthentication()` harness now used for revoke — existing tests for those routes never wire up an auth scheme at all, so they don't actually exercise the same infrastructure.
  Fix: add a fixture using `AddTestAuthentication()` that posts to each sibling route with no `Authorization` header and asserts non-401.

- **[Test Coverage]** No `RevokeTokenAsync` test for an expired-but-owned token (a plausible legitimate action — "clean up my old token").
  Fix: add `Given_RevokeTokenAsync_WithAnExpiredOwnedToken`, pinning the no-op as intentional.

- **[Functionality]** E2E/integration coverage for this exact change was never executed by the author (no Docker in their environment) — independently traced and confirmed logically coherent, but genuinely unverified end-to-end before merge.
  Fix: run the CMS E2E suite (at minimum `OwaspCriticalPaths.feature` Scenario 18) against self-contained mode in CI before merging.

## 🔵 Low

- **[Security] [Maintainability]** Bare `catch` in `IdentityModule.RevokeToken` swallows all exceptions with no logging at the call site (pre-existing pattern, extended rather than introduced by this PR).
- **[Security]** Keycloak `client_id` claim-name equivalence rests on a prior manual check outside this branch, not an automated assertion here; currently no exploitable risk since Keycloak mode never reaches the ownership branch.
- **[Security]** Minor timing side-channel between ownership-match (extra DB write) and mismatch outcomes; not practically exploitable given 128-bit `jti` GUIDs.
- **[Maintainability]** The `"client_id"` claim-type literal has no shared constant, despite this PR adding two more security-sensitive usages of it.
- **[Maintainability] [Test Coverage]** `RevocationOwnershipTests` is a separate top-level class/test-double strategy from the rest of `IdentityModuleTests`, with duplicated crypto/JWT helper code between it and `OpenIddictTokenManagerTests`; the split is defensible (real signature verification needs a real manager) but undocumented as to why.
- **[Maintainability]** New test helper `GivenTheActivePublicKeyIs` breaks the file's `Create*` helper-naming convention and reads like a Gherkin-style fixture name.
- **[Maintainability]** `tasks/plan.md`/`tasks/todo.md` introduce a `[~]` ("partially verified") checkbox marker with no legend explaining it.
- **[Usability]** The "dual-token" mental model (caller's own bearer token vs. the target token being revoked) is explained only in prose, with no concrete example.
- **[Test Coverage]** No fixture distinguishes an empty-string `client_id` claim from a missing one (same practical outcome today, different code path).
- **[Functionality]** Expired target tokens are no longer marked `revoked` in the DB (silent no-op instead) — already documented as a deliberate, low-risk change; worth a one-line note if any admin/status UI ever surfaces raw DB status.

## ⚪ Nit

- **[Maintainability]** Debug-level logging for ownership mismatches is a defensible, explicitly-reasoned tradeoff (avoids Warning-level log flooding from a caller-triggerable no-op), though it does mean an attacker probing many client_ids produces no signal above Debug.
- **[Test Coverage]** The E2E scenario only exercises self-revocation (bearer token and revocation target are the same token), adding no E2E coverage of the cross-client denial path — already covered at the unit/module level, so this is a nice-to-have, not a gap.

---

## Specialist Verdicts
| Reviewer | Verdict |
|----------|---------|
| Security | No Critical/High vulnerability; ownership check and signature verification are implemented correctly. Main action item is deleting `docs/parking-lot.md` before merge and hardening the bare `catch` for observability. |
| Functionality | COMPLETE — all seven planned tasks implemented correctly; build clean, both unit-test suites match the author's claimed pass counts exactly (independently re-run: 667/667 backend, 1211/1211 frontend). Residual risk is the E2E scenario being logically coherent but never actually executed. |
| Maintainability | Well-organized and thoroughly documented for its size; only minor duplication (signing-key setup, `client_id` lookup) and small test-organization inconsistencies, nothing blocking. |
| Usability | The ownership-check design and rationale are sound and clearly documented, but the PR ships a breaking change with no changelog and gives integrators no way to confirm a revoke succeeded or recover when it silently didn't — sharpened by a real, disclosed casing defect that can trigger exactly that silent failure for legitimate clients. |
| Test Coverage | ADEQUATE — happy path, mismatch, forgery, malformed-token, and auth-boundary cases are all well covered with good test quality, but two of the PR's own stated security guarantees (ordinal casing comparison, issuer/audience verification independent of signing-key correctness) are unpinned, and the canary test's own precondition is unverified. |

---

## Summary
This PR is materially more thorough than a minimal implementation — real attack fixtures (forged tokens, signature mismatches), an explicit canary against future policy creep, and a self-disclosed, well-scoped `parking-lot.md` entry for a genuine adjacent defect. Nothing here is Critical, but three items should be resolved before merge: delete `docs/parking-lot.md` (its own instruction), add the missing case-sensitivity and issuer/audience test coverage that the PR's own documented security guarantees call for, and either run the changed E2E scenario in CI or explicitly accept the risk of merging it unverified. The usability gap — no documented way to confirm a revoke succeeded, worsened by a real silent-failure defect — is worth a doc fix even though it's not code-blocking.

---


[DMS-1478]: https://edfi.atlassian.net/browse/DMS-1478?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

stephenfuqua:
Thanks — this was a genuinely useful review. Addressed in `9aeb2b6`. Backend `677/677`, frontend `1216/1216`, 0 skipped, build and `csharpier` clean.

## High — all five addressed

**Ordinal comparison now pinned.** `Given_RevokeTokenAsync_WithATokenWhoseClientIdDiffersOnlyByCase` (token `Owner-Client`, caller `owner-client`). Verified it bites by temporarily flipping to `OrdinalIgnoreCase` — and the result is worth reporting: **only the `MustNotHaveHappened()` assertion caught it.** The companion `It_returns_false` still passed, because the unstubbed fake returns `false` regardless. A return-value-only test would have been a false negative on exactly the regression this guards.

**Issuer/audience now pinned independently of signing-key correctness.** `Given_RevokeTokenAsync_WithATokenFromAnotherIssuer` and `..._WithATokenForAnotherAudience` sign with the service's *own registered key*, so only the claim is wrong — neither can pass on signature checking alone.

**Canary is now armed.** You were right that it proved "this caller isn't rejected," not "this caller would be rejected." `Given_the_role_claim_is_omitted_at_a_service_policy_route` drives `X-Test-OmitRoleClaim` at a real `MapSecuredGet` route and observes **403**, with the same request minus the header asserting not-403. The precondition is observed now, not asserted in a comment.

**Revoke-confirmation gap documented** in both files — the introspect-after-revoke pattern (`{"active": false}`), stated plainly that incident response must not treat `200` as proof, and tied to the casing defect as a reachable cause. This was the sharpest finding in the review; thank you.

**Changelog — no action taken, and I'd like your call.** There is no changelog file anywhere in this repo (full-tree search for `*changelog*`, `*release-note*`, `*whatsnew*`, `*CHANGES*` — all hits were unrelated source files). Release notes appear to live in GitHub Releases: `.github/workflows/on-release.yml` triggers on `release: released` and manipulates releases via the API. Creating a first-ever changelog file felt like inventing a repo convention on your behalf, so I didn't. **Suggest the anonymous→401 breaking change goes in the GitHub Release body — happy to draft it, or to add a file if you'd rather have one.**

## Medium — all addressed

`VerifyTokenAsync` extracted and shared between `ValidateTokenAsync` and `RevokeTokenAsync`, with a doc comment stating the omitted DB-status lookup is load-bearing and must not migrate in (backend suite held at 667/667 across the refactor before new tests were added). `GetClientId()` now used in both handler sites — note the frontend has no *direct* project reference to `Backend.OpenIddict` but already consumes its types transitively, so this added no new coupling and no reference. Keycloak no-op footnote added to the `CS-AUTH.md` table row. Incident-response consequence added to `parking-lot.md`. Sibling-route anonymity fixture (`TokenEndpointAnonymityTests`) and `Given_RevokeTokenAsync_WithAnExpiredOwnedToken` both added.

## Low — all ten addressed

Bare `catch` now logs (sanitized, 200 contract unchanged); `client_id` constant added; test-double split documented; helper renamed; `[~]` legend added to both task files; worked dual-token example added; empty-vs-missing `client_id` fixture added; expired-tokens-not-marked-revoked note added.

One deliberate exception: the `"client_id"` literal at `JwtTokenGenerator.cs:63` was **left alone**. That's the minting path the parking-lot entry flags, and consolidating it there belongs with the minting fix rather than this PR.

## Two items I could not close

**`docs/parking-lot.md` is still present, intentionally.** This PR targets `DMS-1478-token-revocation`, not `main` — the delete-before-merge gate applies at the `main` merge, and the instruction I'm working under is that a human performs that deletion rather than me. Flagging so it isn't mistaken for an oversight: **it still must not reach `main`.**

**E2E Scenario 18 still hasn't run, and I'm not claiming it would pass.** I got further this round — `dockerd` is present and I started it successfully (Docker 29.3.1) — but the image registry is blocked by the egress proxy: even `docker pull hello-world` fails with `Forbidden` on the CDN blob fetch, and there's no local cache. The CMS stack needs Postgres, Keycloak, and the CMS image. So this remains a real pre-merge gate for someone with registry access, and it is **the only check that exercises my E2E step-definition edit** (adding the `Authorization` header to the "the current token is revoked" step), which no unit test covers.

## Acknowledged, no change

Timing side-channel (not practically exploitable given 128-bit `jti` GUIDs), Keycloak claim-name equivalence not automated (no exploitable risk — Keycloak mode never reaches the ownership branch), Debug-level mismatch logging, and E2E covering only self-revocation.

Two things still rest outside anything verified in this branch and need your eyes: the Keycloak `client_id` claim-name equivalence, which rests on a prior manual check against a real Keycloak JWT, and E2E Scenario 18 above.

---
_Generated by [Claude Code](https://claude.ai/code)_

stephenfuqua:
# PR #1272 Re-Review — [DMS-1478] Authenticate and ownership-check POST /connect/revoke
> Requirements source: `tasks/plan.md` / `tasks/todo.md` (no linked GitHub issue)
> Reviewed by: Security · Functionality · Maintainability · Usability · Test Coverage
> This is a **re-review** of the branch as of commit `9aeb2b673` ("PR review corrections for /connect/revoke auth"), which follows two correction rounds ("QA round 1" and "PR review corrections") addressing an earlier review pass.

---

## 🔴 Critical
None found.

## 🟠 High
None found.

## 🟡 Medium

- **[Usability]** The breaking change (anonymous → `401`) still has no changelog or release-note entry anywhere a consumer would see it before upgrading.
  `CS-AUTH.md` and `OWASP-AUTH-COVERAGE.md` are internal design/security references, not integrator-facing docs, and the repo has no `CHANGELOG.md`. The PR description's own "Behavior changes" section is well-written but isn't published anywhere after merge.
  Fix: create a new `docs/changelog/8.1.0.md` file. This file should have a preface describing its purpose (brief descriptions of breaking changes and new features), followed by a bullet-point list containing this first entry about the change to the revocation endpoint. Keep the bullet point brief, and point back to design docs and/or ADR files for more information.

- **[Security]** Warning-level logging conflates a routine, expected no-op (revoking an already-expired token) with genuine forged-signature/wrong-issuer/wrong-audience attack attempts — both log the identical message at `Warning` in `VerifyTokenAsync`/`ValidateTokenAsync`. The one real attack signal this PR adds is buried in routine noise at the same severity.
  Fix: have the verifier report a failure category; log expiry-only failures at Debug and reserve Warning for signature/issuer/audience failures.

## 🔵 Low

- **[Maintainability]** `SecurityConstants.ClientIdClaimType`'s doc comment overstates its own guarantee ("treat any change here as a breaking change to that check") — the literal `"client_id"` is still hardcoded at the actual mint site (`JwtTokenGenerator.cs:63`), in `OpenIddictTokenManager.GetAccessTokenAsync:188`, and in `AuditContext.cs:47`. Renaming the claim today would compile cleanly everywhere except the one place the comment claims to protect.
  Fix: migrate the mint site to the constant, or scope the comment's claim accurately.

- **[Maintainability]** New test helper `CreateActivePublicKey` breaks the file's `Create*` naming convention — it returns `void` and mutates the shared fake as a side effect, unlike every sibling `Create*` helper which builds and returns a value. This reintroduces a naming inconsistency a prior review round flagged and appeared to have fixed elsewhere (`CreateConfiguredTokenManager` does follow the convention correctly).
  Fix: rename to `StubActivePublicKey` or `RegisterActivePublicKey`.

- **[Usability]** The "confirm via introspect" guidance is duplicated near-verbatim — including the identical framing sentence — between `CS-AUTH.md` and `OWASP-AUTH-COVERAGE.md`, with no cross-link. A future change to the no-op conditions has to be applied in both places.
  Fix: make one canonical (`CS-AUTH.md`) and have the other cross-reference it.

- **[Test Coverage]** `OWASP-AUTH-COVERAGE.md`'s test-coverage enumeration understates its own disclosed gap: the PR says the two newest fixtures aren't named, but six are actually missing (case-sensitivity, wrong-issuer, wrong-audience, the canary, the canary-arming fixture, sibling-anonymity, expired-owned).
  Fix: update the enumeration.

## ⚪ Nit

- **[Usability]** `CS-AUTH.md`'s dual-token HTTP example uses invalid trailing `<- comment` annotations inside a fenced ` ```http ` block — copy-pasting it verbatim into curl would fail until stripped.
  Fix: use a `#` character to turn that into a real comment.

## 🚫 Won't Do

- **[Security]** No rate limiting on `/connect/revoke` — every call, including garbage `token` values, now pays for full RSA signature verification, and the route requires only bare authentication (no scope/role policy).
  - Fix: add per-caller rate limiting or alert on abnormal call volume from a single `client_id`.
  - Justification: per-caller / per-client rate limiting would be a brand new feature for this application, and should not be introduced as a side effect of this PR.

- **[Usability]** `CS-AUTH.md`'s prose never states that an authenticated caller whose own bearer token carries no `client_id` claim also silently no-ops — only the target-token-side "no `client_id`" / "different client" cases are documented, even though the code and a dedicated test guard this as a distinct case.
  - Fix: add one clause covering the caller-side empty-`client_id` case.
  - Justification: an authenticated caller's JWT cannot be lacking a `client_id`.

- **[Test Coverage]** No test exercises a caller principal with the `client_id` claim entirely absent (only empty-string is testable via the harness) — fully mitigated at the manager level by `Given_RevokeTokenAsync_WithoutACallerClientId`; unit-test nit only.
  - Justification: See finding above

- **[Maintainability]** JWT-signing test scaffolding (`CreateSigningKey`/`CreateSignedToken`) is duplicated, not shared, between `OpenIddictTokenManagerTests.cs` and `RevocationOwnershipTests` — across a project boundary, so likely acceptable as-is, but a third copy would drift silently.
  - Justification: this is deliberate due to the project boundary.

- **[Security]** `IdentityModule.RevokeToken`'s new outer `catch (Exception ex)` is unreachable for the shipped self-contained manager, since `OpenIddictTokenManager.RevokeTokenAsync` already swallows everything internally; only matters for a hypothetical future implementation that throws instead of returning `false`.
  - Justification: harmless future-proofing

---

## Specialist Verdicts
| Reviewer | Verdict |
|----------|---------|
| Security | No Critical/High vulnerability. The forged-token defense (signature/issuer/audience/lifetime verification before trusting `client_id`) is correctly implemented and well-tested; remaining concerns are operational (log-signal dilution, no rate limiting) and doc hygiene. |
| Functionality | COMPLETE — independently rebuilt and re-ran the full suite (677/677 backend, 1216/1216 frontend, both green); every `[~]` "partially met" substitute claim in `tasks/plan.md`/`tasks/todo.md` was verified true, not fabricated. |
| Maintainability | Well-organized, no duplication regressions survived in the core `VerifyTokenAsync` extraction; only minor naming/constant-migration nits remain. |
| Usability | The technical explanation and incident-response guidance are excellent and code-accurate, but reach only internal-doc readers — no changelog exists anywhere for the breaking `401`. |
| Test Coverage | STRONG — all five gaps flagged in the prior review round are genuinely closed, independently re-verified by reading and running the actual fixtures; remaining issues are PR-description accuracy, not missing coverage. |

---

## Summary
This round closes every substantive gap the previous review found: case-sensitivity, issuer/audience, canary-precondition, sibling-anonymity, and expired-owned-token coverage are all genuinely tested and passing, and the `RevokeTokenAsync`/`ValidateTokenAsync` duplication was cleanly extracted into a shared, well-documented `VerifyTokenAsync` helper. Nothing here is Critical or High. The PR description's "Known gaps" list needs a quick pass to drop one now-stale bullet. Everything else is non-blocking polish (log-level tuning, a changelog entry, minor naming/doc nits).

---


[DMS-1478]: https://edfi.atlassian.net/browse/DMS-1478?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

stephenfuqua:
All seven items resolved in `ca420f0`. Backend `680/680` (+3), frontend `1216/1216`, 0 skipped, build and `csharpier` clean.

Note the branch moved since the commit you reviewed: `61fea58` trimmed the doc comments for concision, and `d043d35` **deleted `docs/parking-lot.md` and the `tasks/` files** at the author's instruction, relocating the casing-defect description inline into `CS-AUTH.md` and `OWASP-AUTH-COVERAGE.md`. The casing defect itself is deliberately still unfixed and now needs its own ticket.

## Medium

**Changelog** — `docs/changelog/8.1.0.md` created to your spec: preface stating the file's purpose, `## Breaking changes` with the `/connect/revoke` entry first, kept brief and linking to the two design docs. An empty `## New features` section gives the next contributor the slot. No version constant touched.

**Log severity split** — this was the best finding in the round. `JwtTokenValidator.ValidateToken` now reports a `TokenVerificationFailure` category (`None`/`Expired`/`Untrusted`); expiry logs at Debug, signature/issuer/audience at Warning.

Two sub-decisions worth surfacing: a **not-yet-valid (`nbf`) token is treated as `Untrusted`, not routine** — it stays at Warning. And the routing applies to `ValidateTokenAsync` too, not just the revoke path, since both share the verifier and an expired bearer token is equally routine there. Say the word if you'd rather that were confined to revoke.

Demonstrated rather than reasoned about: two new fixtures capture log levels from a faked `ILogger` and assert Warning-count `0` / Debug `> 0` for expiry and Warning `> 0` for an untrusted token. Temporarily collapsing both Debug branches back to Warning fails them with `Expected ... Warning to be 0, but found 2` — and that "2" is your finding restated, since one expired token previously emitted *two* Warning entries (validator + manager). Experiment reverted.

## Low / Nit

`StubActivePublicKey` rename done. `CS-AUTH.md` is now the canonical home for the introspect-after-revoke guidance, with `OWASP-AUTH-COVERAGE.md` keeping a standalone summary plus an anchored cross-link. The `http` example annotations are now real `#` comments.

**Coverage enumeration** was understated by **eight** fixtures, not six — your six plus empty-`client_id` and the two new log-severity ones. Re-derived by grepping the test files rather than trusting either list.

## One item I did not do as specified

**`OpenIddictTokenManager.cs:188` was not migrated to `ClientIdClaimType`**, and this is a semantic objection rather than the project-reference problem you might expect. That literal is `kvp.Key.Equals("client_id", ...)` iterating the **credentials collection from the form body** — it is the OAuth *request parameter* name fixed by RFC 6749 §2.3.1, not a JWT claim type. The two merely share a spelling. Binding them to one constant would mean renaming the claim silently renames the wire parameter and breaks every client.

The other two sites — `JwtTokenGenerator.cs:64` (mint) and `AuditContext.cs:47` (read) — did migrate, with no new project references needed.

Instead I rescoped the doc comment to state exactly what the constant does and does not reach. That surfaced a second gap the original wording papered over: it also does not reach the claim name in **Keycloak-issued tokens**, which Keycloak controls — so a change here would break the ownership comparison in Keycloak mode. Happy to migrate line 188 anyway if you disagree, but I'd want to rename the constant first to make the request-parameter/claim-type distinction explicit.

## Unchanged

The five Won't Do items are untouched.

---
_Generated by [Claude Code](https://claude.ai/code)_

stephenfuqua:
**The `client_id` casing defect is now fixed in this PR rather than deferred** (`790dc7e`, `0750636`). It materially affects this change, so deferring it risked confusion. No changelog entry — from a client perspective this is effectively non-breaking.

Backend `685/685` (+5), frontend `1216/1216`, 0 skipped, build clean, `csharpier` clean.

## The fix

**Canonical minting.** The claim is now minted from `applicationInfo.ClientId` (the stored value) rather than the request-supplied string, so every token for a client carries identical casing regardless of how the client cased its credentials.

That call site turned out to feed **four** consumers, not just `client_id`: `sub`, `client_id`, `azp`, and the persisted token row's subject column. All four were carrying request casing; fixing at the call site corrects them together. A fallback to the request value guards the case where the stored value is empty — minting an empty `sub` would be a worse failure than a non-canonical one.

**Case-insensitive lookup on Postgres**, matching SQL Server's existing default-collation behavior. This needed care: Postgres has a *case-sensitive* unique constraint, so a deployed database can already hold `acme-client` and `Acme-Client` as separate clients, and a naive `LOWER() = LOWER()` rewrite against `QuerySingleOrDefaultAsync` would throw — turning a working login into a 500. So: exact match first (the only path a correctly-cased credential takes, so nothing currently working changes), case-insensitive fallback only on a miss, and genuine ambiguity returns null — a normal `invalid_client` failure plus a sanitized Warning naming the match count — never an exception.

`LOWER()` can't use the unique index, so the fallback is a sequential scan. Judged acceptable and stated in the code, the ADR, and here: one row per registered client, and the query only runs after an exact match already missed.

**`Ordinal` in `RevokeTokenAsync` is unchanged**, and its pinning test still passes. Relaxing it would make the ownership boundary depend on the deployed engine's collation.

## ADR

`reference/adr-client-id-casing.md`, cross-linked from `CS-AUTH.md`. The passages in `CS-AUTH.md` and `OWASP-AUTH-COVERAGE.md` that described this as a live hazard are rewritten — they'd otherwise contradict the fix.

## Three things needing your judgement

**1. No unique-index migration.** A case-insensitive unique index is what would actually prevent new case-variant pairs on Postgres, but it would fail to apply on any existing database that already contains one — a code change becoming a failed deployment. Left as an open question in the ADR; it needs a migration strategy (detect-and-report, or require cleanup first), not a quiet addition here.

**2. Pre-existing tokens keep their requested casing** until they expire. Documented rather than migrated, on the assumption that short token lifetimes drain it. Flagging in case that assumption doesn't hold for your deployments.

**3. The new Postgres tests did not run here.** `OpenIddictDataRepositoryTests.cs` gains case-insensitive resolution, exact-match-wins, and non-throwing ambiguity coverage. They compile — the integration project builds as part of the solution — but no Postgres is available in this environment, so **they will first execute in CI**. MSSQL needed no change. Also note `OpenIddictDataRepository`'s constructor gained an `ILogger` for the ambiguity warning; three integration-test call sites were updated to pass `NullLogger` and compile but were not executed.

## On verification integrity

Worth stating plainly: the first attempt at proving these tests fail without the fix reported all five *passing* with the defect present. They weren't worthless — the build had failed on an analyzer error, output filtering hid it, and `--no-build` then ran a stale assembly. Redone properly, all five fail as they should, including `It_revokes_the_token → found False`, which is the silent no-op itself reproduced.

I re-ran the full build and both suites unfiltered before pushing and confirmed the numbers above independently.

---
_Generated by [Claude Code](https://claude.ai/code)_

stephenfuqua:
**Reverted the case-insensitive PostgreSQL lookup** (`d4a9393`, `cf59cc8`). RFC 6749 — "Unless otherwise noted, all the protocol parameter names and values are case sensitive" — makes case-sensitive the conformant behaviour, so PostgreSQL's default collation was already correct and the Robustness-Principle argument is withdrawn.

Backend `685/685`, frontend `1216/1216`, 0 skipped, build clean. Counts unchanged, as expected: everything reverted and everything added lives in the integration project.

**Canonical minting is kept.** That was the actual fix and it is untouched.

## Scope of the revert, verified

- `Backend.Postgresql/.../OpenIddictDataRepository.cs` — **diff against `ca420f0` is empty.** Byte-identical to its pre-fix state: plain `WHERE a."ClientId" = @ClientId` with `QuerySingleOrDefaultAsync`. The exact-then-fallback logic, ambiguity handling, and the `ApplicationLookupSql` split are gone.
- The `ILogger` constructor parameter is removed — its only consumer was the ambiguity warning. The three integration-test call sites are restored; `ApplicationTests.cs` also diffs to zero.
- **MSSQL: zero lines changed**, as instructed.

The three tests asserting case-insensitive resolution are removed — they pinned behaviour that is now deliberately absent. In their place, `Given_A_Client_Looked_Up_With_Different_Casing` pins the correct behaviour: exact match resolves, case variant returns null. Its comment quotes the RFC and records that an earlier revision loosened this and it was reversed on purpose, so nobody "helpfully" re-loosens it later.

## The reasoning the docs now carry

Worth stating explicitly, because it is what makes deferring the MSSQL work safe: **canonical minting alone closes the revocation silent-no-op on both engines.** Even where SQL Server's collation lets a client authenticate with non-canonical casing, the token it receives now carries the canonical `client_id`, so the `Ordinal` ownership comparison matches and revocation succeeds.

The SQL Server divergence is therefore an **RFC-conformance and cross-engine-consistency issue, not a revocation hazard** — recorded as a known issue in the ADR, `CS-AUTH.md`, and `OWASP-AUTH-COVERAGE.md`, each noting it is tracked separately and out of scope here. I swept all three for `casing|case-insensitive|collation|robustness`: nothing still implies a casing-driven silent no-op is reachable, and the unique-index open question is gone with the change that raised it.

The `Ordinal` comparison now has two recorded justifications — the collation-dependence argument, and the same RFC 6749 stance.

**RFC citation:** quoted verbatim with no section number, and the ADR says outright that the section was not verified rather than guessing one. Worth a reviewer's confirmation.

## Unchanged caveats

- `Given_A_Client_Looked_Up_With_Different_Casing` needs a live PostgreSQL and **did not run here** — it compiles and will first execute in CI, same status as the other integration tests and E2E Scenario 18.
- Tokens minted **before** this change still carry their requested casing and may resist revocation by a canonically-cased caller until they expire. Documented rather than migrated, on the assumption that short lifetimes drain the population.

---
_Generated by [Claude Code](https://claude.ai/code)_

stephenfuqua:
## Review of latest commits (9aeb2b673..cf59cc84dd)

The core fix looks sound: `client_id` is now minted from `applicationInfo.ClientId` (the canonical
stored value) instead of the request-supplied string, the `Ordinal` comparison in `RevokeTokenAsync`
is unchanged, and the brief case-insensitive Postgres lookup was correctly reverted back to
case-sensitive. `reference/adr-client-id-casing.md` accurately documents the decision.

Four findings from a detailed pass, most severe first:

1. **Stale comment contradicts the fix** — `OpenIddictTokenManager.cs:284`
   A comment added by 790dc7e60 states flatly "Client lookup is case-insensitive," but d4a9393d8
   reverted Postgres back to case-sensitive without updating this comment. It now contradicts the
   ADR, `CS-AUTH.md`, and the new test that pins Postgres as case-sensitive. Risk: a future
   maintainer trusts the comment and reintroduces the case-insensitive-lookup bug that was just fixed.

2. **Unsanitized `ex.Message` in new logging** — `JwtTokenValidator.cs:118`
   The diff sanitizes the `kid` header a few lines up (`LoggingUtility.SanitizeForLog`) but logs the
   `SecurityTokenExpiredException`/general exception message raw. Issuer/audience validation
   exceptions can embed attacker-supplied claim values, enabling log-line forgery in code added
   specifically to improve security visibility.

3. **`ClientIdClaimType` doc comment is incomplete** — `SecurityConstants.cs:14`
   Claims to enumerate every consumer of the constant but misses two DMS-core files
   (`JwtValidationService.cs`, `ApiClientDetailsProvider.cs`) that read the same claim via a
   hardcoded `"client_id"` string literal. A future rename — which the comment itself calls a
   breaking change — would silently break those readers with no compiler error.

4. **`Untrusted` failure bucket conflates forgery with config drift** — `JwtTokenValidator.cs:29`
   Bad signature, wrong issuer, and wrong audience all log as "potential forgery" at Warning under
   one bucket. A deploy-time `Authority`/`Audience` config typo would trigger the same alarm as an
   actual forgery attempt, risking either a phantom incident or alert fatigue.

stephenfuqua:
All four addressed in `799d9c3`. Backend `689/689` (+4), frontend `1216/1216`, 0 skipped, build clean — re-verified independently, not taken from the working notes.

**1. Stale comment.** Rewritten so the justification no longer rests on the false premise. It now leads with the engine-independent reason — every token issued to a client should carry one identity — and states the actual situation: where an engine resolves a mis-cased client id, which SQL Server does through its collation and Postgres does not, minting from the request would stamp two identities onto one client's tokens.

Your instinct that this was a trap for a future maintainer was right, and there was a second copy: the identical false claim sat in a test comment in `OpenIddictTokenManagerTests.cs`, from the same commit. Fixed both — leaving one behind would have preserved exactly the hazard you flagged. A repo-wide grep now returns no remaining assertion that lookup is case-insensitive.

**2. Unsanitized `ex.Message`.** All three catch sites now sanitize. The success-path log got a differentiated answer rather than a blanket one: `Issuer` and `Audience` are left raw because `ValidateToken` has just succeeded with `ValidateIssuer`/`ValidateAudience` on, so both are provably equal to configured values — sanitizing would imply a taint that isn't there. `Subject` **is** sanitized: nothing validates `sub` against configuration, it originates from a registration-time client id, and a control character there would forge log lines. The reasoning is recorded in a comment so it doesn't have to be re-derived.

**3. `ClientIdClaimType` doc comment.** No longer reads as an enumeration — the CMS consumers are labelled examples, and three categories of external reader are listed: DMS core, Keycloak-issued tokens, and the RFC 6749 request parameter. The DMS bullet states the consequence plainly: renaming the claim is a cross-product breaking change that compiles cleanly and fails at runtime.

Both files you named verified as genuine claim readers before citing — `JwtValidationService.cs:255` and `ApiClientDetailsProvider.cs:30`. No shared constant was attempted: `src/dms` has no reference to this assembly. Two other hits were deliberately *not* cited because they aren't claim readers — `ConfigurationServiceTokenHandler.cs` (outbound OAuth request parameter) and `TokenInfo.cs` (a `JsonPropertyName` on the introspection contract).

**4. Category split.** New `UntrustedIssuerOrAudience`, caught via `when (ex is SecurityTokenInvalidIssuerException or SecurityTokenInvalidAudienceException)` ahead of the general catch, with a message pointing at configuration: *"Check the configured Authority and Audience if this affects every token."* Signature, unparseable, and unresolved `kid` stay in the forgery-suspicious bucket — an unknown key id isn't a config typo in the same sense.

Both stay at **Warning**. Each warrants attention; they warrant *different* attention, and the message carries that rather than the level. Routing in `LogVerificationFailure` is now an explicit switch.

Because both categories log at Warning, level-counting can't tell them apart, so the tests assert on rendered message content. The wrong-audience fixture signs with the service's own key so only the audience is wrong, and asserts it does **not** carry the signature message. Verified the new assertion bites: collapsing the category back into `Untrusted` fails it. Log tests went 3 → 7.

`OWASP-AUTH-COVERAGE.md` updated to three categories. Also corrected the `<remarks>` on `ValidateToken`, which still said everything non-expired "indicates a token that was not issued by this service" — untrue of an issuer/audience mismatch, where the token *was* issued by this service. That was the same conflation your finding identified, one layer up in the docs.

Unchanged: the Postgres integration tests and E2E Scenario 18 remain unexecuted here (no database; Docker daemon runs but image pulls are proxy-blocked) and will first run in CI.

---
_Generated by [Claude Code](https://claude.ai/code)_

reference/adr-client-id-casing.md
```diff
@@ -0,0 +1,128 @@
...
+> Unless otherwise noted, all the protocol parameter names and values are case sensitive.
+
+`client_id` is such a value, so rejecting a mis-cased id is the conformant behaviour, and
+PostgreSQL's case-sensitive default collation already does the right thing. Accepting any casing
+would have been a deliberate departure from the specification in order to be lenient, which is not
+a trade this service needs to make: the revocation defect that motivated the leniency is already
+closed by canonical minting.
+
+(The quoted sentence is from RFC 6749; the section number was not verified when writing this ADR
+and is deliberately omitted rather than guessed.)
```
stephenfuqua:
```suggestion
```

Unnecessary detail.

reference/adr-client-id-casing.md
```diff
@@ -0,0 +1,128 @@
...
+
+If a stored row somehow carries an empty `ClientId`, the request value is used instead. Minting an
+empty subject would be worse than minting a non-canonical one.
+
+**This change alone closes the revocation defect, on both database engines.** Even where an engine
+lets a client authenticate with non-canonical casing, the token it receives now carries the
+canonical `client_id`, so the ownership comparison matches the client's other tokens and revocation
+succeeds. No change to lookup behaviour is required to fix it.
+
+### Client lookup stays case-sensitive
```
stephenfuqua:
```suggestion
### Client lookup stays case-sensitive for PostgreSQL
```

reference/adr-client-id-casing.md
```diff
@@ -0,0 +1,128 @@
...
+SQL Server's default collation is case-insensitive, so `GetApplicationByClientIdAsync` there
+resolves a client from a mis-cased `client_id` where PostgreSQL rejects it. The two engines
+therefore behave differently for the same request, and the SQL Server behaviour departs from the
+RFC 6749 case-sensitivity rule quoted above.
+
+**This is deliberately out of scope for this change and will be addressed under a separate
+ticket.** It is an RFC-conformance and cross-engine-consistency issue, **not** a revocation
+hazard: canonical minting means a client that authenticates on SQL Server with non-canonical
+casing still receives a token bearing the canonical `client_id`, so ownership-checked revocation
+works correctly for it. Nothing about this divergence reopens the silent no-op.
```
stephenfuqua: This ADR should be universal, not ticket-specific.
```suggestion
This is an RFC-conformance and cross-engine-consistency issue, **not** a revocation
hazard: canonical minting means a client that authenticates on SQL Server with non-canonical
casing still receives a token bearing the canonical `client_id`, so ownership-checked revocation
works correctly for it. Nothing about this divergence reopens the silent no-op.

The RFC-conformance issue may be reconsidered at a later date, at which time this ADR should be amended.
```

reference/adr-client-id-casing.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Consequences
+
+- Tokens minted after this change carry the canonical casing. Tokens minted **before** it still
+  carry whatever casing was requested, so a pre-existing token whose casing is non-canonical remains
+  unrevocable by a canonical caller until it expires. No migration is attempted; token lifetimes are
+  short and the population drains on its own.
+- PostgreSQL lookup behaviour is unchanged by this work, and a test now pins its case-sensitivity
+  so the property is deliberate rather than incidental.
```
stephenfuqua: Excessive detail for this ADR.
```suggestion
```

PR #1276, opened 2026-09-21
fix(config): include PathBase in the /metadata/specifications self-request

edg-jf:
## Problem

`GET /metadata/specifications` returns a 500 on any Configuration Service deployment that sets `AppSettings:PathBase`:

```json
{"detail":"","type":"urn:ed-fi:api:internal-server-error",
 "title":"Internal Server Error","status":500,"validationErrors":{},"errors":[]}
```

`MetadataModule` assembles its response by requesting its own OpenAPI document over HTTP, building the URL from the request scheme and host only:

```csharp
.GetStringAsync($"{context.Request.Scheme}://{context.Request.Host}/openapi/v1.json");
```

The document is served relative to `PathBase`, so the request goes to a path that does not exist. The 404 fails `EnsureSuccessStatusCode` inside `GetStringAsync`, and the unhandled `HttpRequestException` reaches the caller as a 500 with an empty `detail` — the cause is visible only in the service log:

```
System.Net.Http.HttpRequestException: Response status code does not indicate success: 404 (Not Found).
   at System.Net.Http.HttpResponseMessage.EnsureSuccessStatusCode()
   at System.Net.Http.HttpClient.GetStringAsyncCore(...)
   at EdFi.DmsConfigurationService.Frontend.AspNetCore.Modules.MetadataModule.<MapEndpoints>b__2_0(HttpContext context)
      in .../Modules/MetadataModule.cs:line 25
```

## Reproduction

With `AppSettings__PathBase=/edfi/config/saas/v8.0/ds5.2/core` behind a reverse proxy:

| URL | Result |
| --- | --- |
| `https://<host>/openapi/v1.json` — what the code requests | 404 |
| `https://<host>/edfi/config/saas/v8.0/ds5.2/core/openapi/v1.json` — where it is served | 200 |

This does not reproduce under docker-compose, which sets no path base: the omitted segment is empty there, which makes the URL correct by coincidence.

## Fix

Include `context.Request.PathBase` when composing the self-request.

Every other self-referential URL in both services already does this — `InformationModule`, `DataStoreContextModule`, and the DMS's `HttpRequestResponseExtensions.RootUrl` — so this appears to be an oversight rather than an intended difference.

## Test

`Metadata_Specifications_Requests_The_OpenApi_Document_Under_The_PathBase` asserts the URL that is requested rather than the status that is returned.

Status cannot detect this defect under `TestServer`: `UsePathBase` only strips a prefix it matches, so a request for a bare `/openapi/v1.json` is passed through unchanged and still routes. Both the correct and the incorrect URL therefore answer 200 in process. A real deployment differs because the proxy in front routes only the path-base prefix to this service, so the bare path never arrives. (I first wrote this test against the status code, and it passed against the unfixed code — hence the change in approach.)

Verified against the unfixed code:

```
Failed Metadata_Specifications_Requests_The_OpenApi_Document_Under_The_PathBase
  Expected requestedUris[0]!.AbsolutePath to be "/config/v8.0/ds5.2/openapi/v1.json",
  but "/openapi/v1.json" ... differs near "ope" (index 1).
```

and with the fix applied: `Passed: 2` (the pre-existing `Metadata_Specifications_Endpoint_Is_Registered` is unaffected).

`CreateFactory` gained an optional `pathBase` and an optional request-URL recorder, both defaulting to the previous behavior, so every existing case is unchanged. `dotnet csharpier check` reports both files conform.


stephenfuqua:
@edg-jf Thank you for providing this fix!

Please also modify the `CONTRIBUTORS.md` file to give EdGraph credit for this work.

simpat-adam (approved):

